PR #18, opened 2010-09-14
Import Debian packaging files

frandieguez:
I added some files to allow the git-extras scripts to be packaged into deb and be installed in Debian based systems.

And I added a manpage for git-changelog and a dummie manpages for the rest of scripts.


nicolaiskogheim:
@hemanth It looks like we should close this one. The merge branch is gone, even.

hemanth:
Hmm, ok.

PR #177, opened 2013-06-14
Pass release version to git-release

twolfson:
I wanted to be able to update my `package.json` with the same version that I am about to release. As a result, I introduced passing of the release version to the `pre-release` and `post-release` hooks.


breneser:
I was going to open a new issue for this. Maybe this should be re-openned ? @twolfson's library is node specific, it would be nice to get the tag name that is being release.

spacewander:
@breneser 
I suggest to reuse #192.
That one is also related and still opened.

PR #192, opened 2013-09-27
pass release arguments to pre/post-release scripts

xdbr:
the _git-release_ command should pass the tagname to the pre-post/release scripts for them to be able to use this version string to automatically update some files in the repo.

when already at it, one can also pass all arguments to the hooks: the avid user is free to simply attach a third, fourth,... argument to _git-release_ without affecting that command's spec

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -266,7 +266,7 @@ Matt Colyer (1)
 total 1844
 ```
 
-## git-release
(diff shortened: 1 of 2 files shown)


spacewander:
@xdbr 
Your pull request has been merged as #651(with some modification)
Thanks for your contribution!

PR #202, opened 2013-12-05
Locking files in git repository

julionc:
The way to ignore certain files that are already committed is:

```
git update-index --skip-worktree filename
```

I thought that this shortcut might be useful to exclude/protect files from version control.

Usage:

```
git proctect-file config/database.yml
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -66,6 +66,8 @@ $ brew install git-extras
  - `git local-commits`
  - `git archive-file`
  - `git missing`
+ - `git lock`
+ - `git unlock`
 
 ## git-extras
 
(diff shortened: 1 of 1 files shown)


Bodacious:
I love git extras so far—hugely useful!

Can someone please explain to me the benefit of *locking* files in git?

qw3rtman:
From what I understand, if you `git lock <file>`, it will hide `<file>` in `git status`.

This makes it useful for config files that may be changed often (or left completely untracked, leading to a bunch of files being shown as untracked in `git status`), with no contribution to the actual code.

[See this StackOverflow answer for more info.](http://stackoverflow.com/a/14462290)

Bodacious:
So this is a feature I'd use for a file already committed to the repo, rather than one I want to ignore (like `config/database.yml`?

PR #208, opened 2014-02-04
add git-swap command to swap the two last commits

stephanebachelier:


megawac:
:+1: though I'd prefer `git swap sha1 sha2` with `HEAD~1` and `HEAD~2` being the default shas

kaustubh77:
OPs query must have been solved till now but just for the record putting my answer here. I had encountered a similar situation and this is how I proceeded. I have explained how this can be done with the help of available commands logically. Revert back if you have any doubt.
https://stackoverflow.com/a/60671512/8418240

stephanebachelier:
6 years later, I close the PR, as I'm doing it manually using interactive rebase. I often need to squash or edit some commits in a development branch, so polishing history is much more work than just swapping commits.

@kaustubh77 your link is a great explanation of the process to swap the last two commits and a good lesson for people not familiar with git.

PR #214, opened 2014-02-20
Delete branches in all listed remote repositories

hickey:
This pull will look for branches in all of the remote repositories and delete the remote branches, not just origin.

### DIFF
--- a/bin/git-delete-branch
+++ b/bin/git-delete-branch
@@ -3,5 +3,12 @@
 branch=$1
 test -z $branch && echo "branch required." 1>&2 && exit 1
 git branch -D $branch
-git branch -d -r origin/$branch
-git push origin :refs/heads/$branch
+
+# check other repos for the branch
+for other in `git branch -r | grep $branch`; do
+  repo=${other%/$branch}
+  
+  git branch -d -r $repo/$branch
+  git push $repo :refs/heads/$branch
+done
+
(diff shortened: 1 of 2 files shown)


bturrubiates:
Maybe this should be an option like `--all` instead of the default behavior.

hickey:
Sorry, I have left this alone for far too long......

I have at times found the default nature of deleting the remote branch a hindrance as well as a feature. So the more I think about it, @bturrubiates has the better idea to trigger deleting the remote branch off of a switch. By now this may be in the git-extras code base. So I will close this PR and do badly needed update of my local repo :-)

qw3rtman:
If this were to get implemented, I agree with @bturrubiates that this should not be the default behavior. In my opinion, it's pretty unexpected and may cause problems for the vast majority who is not expecting this kind of behavior.

bturrubiates:
> I have at times found the default nature of deleting the remote branch a hindrance as well as a feature. So the more I think about it, @bturrubiates has the better idea to trigger deleting the remote branch off of a switch. By now this may be in the git-extras code base. So I will close this PR and do badly needed update of my local repo :-)

I don't think it's in the `git-extras` code base yet. I haven't had enough time work up the other PR I was going to do yet, but I could add that to it whenever I get to it. 🎉 



> If this were to get implemented, I agree with @bturrubiates that this should not be the default behavior. In my opinion, it's pretty unexpected and may cause problems for the vast majority who is not expecting this kind of behavior.

It's potentially very dangerous. If it is behind a `--all` switch, then at least `git-extras` is less responsible for causing damage.

qw3rtman:
Agreed, @bturrubiates. Would love to see a PR whenever you get the chance. :smile:

PR #308, opened 2015-02-05
extra mktemp into git_extra_mktemp

spacewander:
now we can put all utility functions to git-extra-utility


hemanth:
Sorry, this went unnoticed.

PR #310, opened 2015-02-10
Add git-delta

imsky:
This lists files that differ from a branch (by default all new changes vs. master). One useful application is getting a list of files to check with a lint tool.


hemanth:
:+1: Cool, could you please update the wiki with the same?

imsky:
@hemanth Will do.

PR #312, opened 2015-02-11
Added in option to concat extra feature names

RichardLitt:
This takes any extra arguments supplied to `$git feature` and sets
them as the new feature name. For instances, instead of throwing an
error if `$git feature new thing` is called, it will create branch
`feature/new-thing`.


hemanth:
:+1:

PR #314, opened 2015-02-12
[feature] add custom commit message options supports

snowyu:
+ new optional argument "-m" to custom commit message instead of the default commit message.

```bash
git release v1.0.0 -m "+ cusom commit message supports"
```

### DIFF
--- a/man/git-release.md
+++ b/man/git-release.md
@@ -3,11 +3,11 @@ git-release(1) -- Commit, tag and push changes to the repository
 
 ## SYNOPSIS
 
-`git-release` &lt;tagname&gt; [-r &lt;remote&gt;] [-c]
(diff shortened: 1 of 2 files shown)


man/git-release.md
```diff
@@ -30,6 +34,10 @@ git-release(1) -- Commit, tag and push changes to the repository
 
     $ git release 0.1.0
 
+  * Release commit with the given &lt;tagname&gt; and custom commit message.
+
+    $ git release 0.1.0 -m "+ powerful feature added."
```
hemanth: With `"+ powerful feature added."` the commit message would look like?

snowyu:
It's just an example. The defaults commit message is "Release ${version}", but now you can custom your commit message.

hemanth:
SG.

PR #316, opened 2015-02-16
changelog config

pirxpilot:
Added support for git config:
- `changelog.format` - let's you change the `--pretty` format
- `changelog.opts` - is passed directly to git-log when generating

The idea behind those is that by manipulating global (per machine) and local (per repo) git config you can make `git-changelog` work seamlessly with all the legacy formats people use in various projects.

The defaults remain unchanged - if you don't configure any options, the changelog behaves as it did before those changes.

Also: some  superfluous white-space removal - if it turns our I am seeing things just don't take that commit.

### DIFF
--- a/bin/git-changelog
+++ b/bin/git-changelog
@@ -3,7 +3,9 @@
 FILE=""
 LIST=false
 TAG="n.n.n"
-GIT_LOG_OPTS=""
+GIT_LOG_OPTS=$(git config changelog.opts)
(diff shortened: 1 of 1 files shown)


hemanth:
@pirxpilot Nice, could you please squash the commits?

pirxpilot:
sure: everything as one commit one

hemanth:
Thanks.

Issue #302, opened 2014-12-29
summary prints folder instead of project name


benbacardi:
If you're inside a subfolder inside the git repo, `git summary` prints the name of the folder you're in under `project:` instead of the actual name of the project:

```bash
ben@mb18154:~/github/git-extras/bin$ git summary

 project  : bin
```

PR #317, opened 2015-02-19
git-summary correctly displays project name

jykntr:
git-summary correctly displays project name when not executed in the top
level directory of a repository.  Fixes tj/git-extras#302.

Uses the git-sh-setup script to find top level directory and then uses
that directory name for the project name.


hemanth:
`git exec-path` nice, thanks.

Issue #321, opened 2015-03-05
Add merge-to


adius:
How about adding a `merge-to` command like described here: http://stackoverflow.com/questions/3672073/git-merge-to-another-branch ?

hemanth:
SGTM. Would like to hear it from @tj and @spacewander.

pfctgeorge:
:+1:

spacewander:
That will be useful. I will try to implement it next week.

hemanth:
Cool, @adius might also be interested.

adius:
Interested in implementing? I would if I have to, but if someone is more eager to implement it … sure, go ahead. =P
I was just wondering if merge-to is the right term. Maybe rather merge-into?

hemanth:
heh heh! `merge-into` is a better name.

PR #323, opened 2015-03-10
add git-merge-into

spacewander:
For the proposal #321

### DIFF
--- a/bin/git-merge-into
+++ b/bin/git-merge-into
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+current_branch() {
+    git rev-parse --abbrev-ref HEAD
+}
+
+usage() {
+    echo "Usage: git merge-into [src] dest [--ff]"
+}
+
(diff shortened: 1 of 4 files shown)


spacewander:
This command works as http://stackoverflow.com/questions/3672073/git-merge-to-another-branch describes.

At first, I tried to implement `merge-into` with `git push`, as the highest voted answer does. Soon I found out it only works with fast-forward mode.

Then, I implement `merge-into` based on `git checkout` and `git merge`. But here is a problem, while user is running `git-merge-into`, there will be temporary change in working tree due to `git checkout`. If the user has some hooks with files change, this command will trigger them accidentally.

Hence, I add `--ff` to the command. If the user make sure the branch can be merged fast-forward, he/she'd like to use this option to avoid potential risk.

hemanth:
:+1: Awesome. 

Please update the wiki as well.

PR #324, opened 2015-03-11
Support ~ in git-ignore

chernjie:


hemanth:
Neat, could you please squash the commits?

hemanth:
Separate commits makes sense in this case. Thanks.

PR #325, opened 2015-03-11
Complete rewrite to support commit ranges for pretty and list output.

markeissler:
This is a complete rewrite of git-changelog to support more functionality. All previous behavior has been retained for backwards compatibility. 

The primary new functionality is centered around support for tag ranges: users can now specify a --start-tag and/or a --final-tag (working backwards, --start-tag is furthest back in history; the --final-tag acts as a stop point). Users can also just specify the --all option to return all commits from now back to the very first commit whether or not tags exist. The --all option can be specified with the --list option to return list format instead of "pretty" format (commits separated by titles--tags/date).

Finally, the --stdout option has been introduced for additional flexibility. This option does exactly what it does everywhere else: sends output to stdout instead of writing it to a file.

I wrote this because it seems like a natural upgrade to this awesome collection of git utils. And I got tired of looking for other git-changelog alternatives when I realized that the exist git-extras/git-changelog was lacking more functionality. This contribution has been rebased on top of 06a8f74 and includes latest mktemp change that appeared in git-changelog a few days ago (14ef2a8).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,6 +13,7 @@ Maintainer
 Patches and Suggestions
 ```````````````````````
 
+- Mark Eissler
 - Kenneth Reitz
 - Aggelos Orfanakos
 - Jonathan "Duke" Leto
--- a/bin/git-changelog
+++ b/bin/git-changelog
@@ -1,59 +1,307 @@
 #!/usr/bin/env bash
 
-FILE=""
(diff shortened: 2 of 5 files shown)


hemanth:
I just love these detailed PRs (:

Please do update the wiki.

PR #327, opened 2015-03-14
Updated bash completions for better-changelog.

markeissler:
Missed updating bash completions on the first go around.


hemanth:
Thanks.

PR #331, opened 2015-03-16
git-changelog fixes

markeissler:
Fix for when tag and HEAD point to the same commit. Fix for when a branch name appears as a tag (happens when gitflow is used).

Hopefully that's it for now. :)


hemanth:
:+1:

PR #334, opened 2015-03-17
when deleting merged branches, preserve "svn" branch for git-svn

### LINKED ISSUE #328: git delete-merged-branches: don't delete "svn"
When I run `git delete-merged-branches`, I don't want it to delete my `svn` branch, which is needed by svn.

Can you make this the default behaviour? If not, can you add a configurable whitelist of branches to keep, either globally or per-project?

paulschreiber:
Fixed #328


hemanth:
Thank you.

PR #344, opened 2015-04-08
Prevent strange input and environment variables from causing issues

Somasis:
GNU `grep` always looks for GREP_OPTIONS in the environment, and since
we don't unset it on run, this can result in some issues concerning the
usage of `grep` in scripts. So, we should unset it so nothing strange
can occur.

In addition, strange things can happen if you give `grep` input, and
it interprets it as an argument. So, we should use `--` whenever using
it for patterns that have a chance to be influenced by the user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ BINS = $(wildcard bin/git-*)
 MANS = $(wildcard man/git-*.md)
 MAN_HTML = $(MANS:.md=.html)
 MAN_PAGES = $(MANS:.md=.1)
-LIB = "helper/git-extra-utility"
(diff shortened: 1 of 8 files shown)


hemanth:
Thank you!

nicolaiskogheim:
This breaks `git alias` for me. Haven't tested the other changes this introduces.

hemanth:
@nicolaiskogheim Oh, shall we revert?

nicolaiskogheim:
I don't know if the changes besides the one on `git alias` does what it is supposed to do, so I wouldn't rush it before having tested it.
But we need `git alias` working before release anyway. I'll send a PR.

hemanth:
Awesome!

PR #345, opened 2015-04-09
Hope to fix #341

### LINKED ISSUE #341: [feature] git guilt
> Calculates the change in blame between two revisions.

> git-guilt is a little tool I wrote that lets you see the transfer of blame in a repository caused by a range of commits. For example, if you wanted to see how blame shifted from one author to another during your last commit

Maybe this should be added to the git-extras? It's very useful to find reviewers for pull requests.

Detailed explanation:
https://www.atlassian.com/git/articles/git-guilt-blame-and-code-review#introducing..-git
https://www.npmjs.com/package/git-guilt

spacewander:
1. Change the regex `[:digit:]` to `0-9`, make it compatible with mawk.
2. Filter the output of `git blame` and only store the result into tmp file - to reduce the size of tmp file.
3. use `@` as the default value of `until`

### DIFF
--- a/bin/git-guilt
+++ b/bin/git-guilt
@@ -32,16 +32,6 @@ do
 done
 
 cd "$(git-root)" # cd for git blame
-SINCE_LOG=$(git_extra_mktemp)
-UNTIL_LOG=$(git_extra_mktemp)
-for file in $(git diff --name-only "$@")
-do
-    test -n "$DEBUG" && echo "git blame $file"
-    # $1 - since $2 - until
-    git blame $NOT_WHITESPACE --line-porcelain "$1" -- "$file" 2> /dev/null >> $SINCE_LOG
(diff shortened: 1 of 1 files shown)


bin/git-guilt
```diff
@@ -49,19 +39,29 @@ then
...
     PATTERN='s/^author //p'
 fi
-sed -n "$PATTERN" $SINCE_LOG | sed 's/^\(.\)/- \1/' >> $MERGED_LOG
-sed -n "$PATTERN" $UNTIL_LOG | sed 's/^\(.\)/+ \1/' >> $MERGED_LOG
+
+for file in $(git diff --name-only "$@")
+do
+    test -n "$DEBUG" && echo "git blame $file"
+    # $1 - since $2 - until
+
```
robertwhitebit: Either you are missing a space character or the line should be empty overall. :wink:

spacewander: @robertwhitebit 
I removed it right now. Now it looks more compact.

Issue #338, opened 2015-03-20
`change-log -a` must not expect tags.


hemanth:
```
± |master ✓| → git-changelog -l -a

ERROR:  Specified start-tag does not exist!

```
```
  -a, --all                 Retrieve all commits (ignores --start-tag, --final-tag)
  -l, --list                Display commits as a list, with no titles
```

markeissler:
Sorry, I didn't see this. Will take a look.

markeissler:
This is a duplicate of #337.

Issue #337, opened 2015-03-20
change-log still needs some fixes.


hemanth:
```js
/usr/local/bin/git-changelog: line 230: local: -A: invalid option
local: usage: local name[=value] ...

ERROR:  Specified start-tag does not exist!
```
@spacewander @chernjie  Noticing this?

`local -A` is not working on:

```
Darwin Apples-Macintosh.local 14.0.0 Darwin Kernel Version 14.0.0: Fri Sep 19 00:26:44 PDT 2014; root:xnu-2782.1.97~2/RELEASE_X86_64 x86_64
```

```
GNU bash, version 3.2.53(1)-release (x86_64-apple-darwin14)
Copyright (C) 2007 Free Software Foundation, Inc.
```

yogin:
Same issue here running on a mac.

From the bash man pages, `local` uses the same options as `declare`:

Here's the list of supported options under Ubuntu:
```
   declare [-aAfFgilnrtux] [-p] [name[=value] ...]
       typeset [-aAfFgilnrtux] [-p] [name[=value] ...]
              Declare variables and/or give them attributes.  If no names are given then display the values of variables.  The -p option will display the attributes and values of each name.  When -p is used with name arguments, additional options, other than -f and
              -F,  are  ignored.   When  -p is supplied without name arguments, it will display the attributes and values of all variables having the attributes specified by the additional options.  If no other options are supplied with -p, declare will display the
              attributes and values of all shell variables.  The -f option will restrict the display to shell functions.  The -F option inhibits the display of function definitions; only the function name and attributes are printed.  If the extdebug shell option is
              enabled using shopt, the source file name and line number where the function is defined are displayed as well.  The -F option implies -f.  The -g option forces variables to be created or modified at the global scope, even when declare is executed in a
              shell function.  It is ignored in all other cases.  The following options can be used to restrict output to variables with the specified attribute or to give variables attributes:
              -a     Each name is an indexed array variable (see Arrays above).
              -A     Each name is an associative array variable (see Arrays above).
              -f     Use function names only.
              -i     The variable is treated as an integer; arithmetic evaluation (see ARITHMETIC EVALUATION above) is performed when the variable is assigned a value.
              -l     When the variable is assigned a value, all upper-case characters are converted to lower-case.  The upper-case attribute is disabled.
              -n     Give each name the nameref attribute, making it a name reference to another variable.  That other variable is defined by the value of name.  All references and assignments to name, except for changing the -n attribute itself, are  performed  on
[...]

Issue #336, opened 2015-03-19
It's time for a new version


chernjie:
@hemanth :)

hemanth:
Yup, will be doing in this weekend :)

chernjie:
Awesome

chernjie:
#343

hemanth:
After #337 and #338 are fixed.

chernjie:
@hemanth I suggest we revert #325 until the above issues are resolved. What do you think?

hemanth:
@chernjie How long would it take to resolve those? Looks like it's shaping up well, right?

chernjie:
@hemanth wait and see?

Say, we should look into some testing framework, e.g. travis, various versions of Git etc... It would help protect you from issues such as backward compatibility.

markeissler:
@chernjie But what would that look like? If we consider the comment for git 1.9.2 not supporting "%D" for  a pretty format placeholder... (https://github.com/tj/git-extras/pull/355#issuecomment-94010021), then some of this testing could get down in the weeds pretty quickly. Testing would require knowing which functionality is required by each git-extra, and then knowing which installed versions of system utilities are required to support that functionality.

I think a start would be to outline, project-wide, which minimal version of git, grep, sed, etc., must be supported by each git-extra. That's easier to work with. Just looking at grep and sed, for instance, there's GNU and BSD versions of each.

spacewander:
I agree with @markeissler 
We should have a project-wide requirements. 
Therefore we can have a limited range to test for, and stop the support if the version of git/bash/other tools is too old.

chernjie:
@markeissler I agreed with you and @spacewander.

One step at a time :)

chernjie:
@hemanth so what's the verdict? Are we going to tag a new release?

hemanth:
Revert 325 and release?

chernjie:
I'm fine with either. To revert you can merge #352

markeissler:
@hemanth @chernjie Just rebased and repushed better-changelog branch #355.

hemanth:
Thanks @markeissler merged it. Was it checked on GNU/Linux?

markeissler:
Just  tested it on CentOS 7. Looks good.

hemanth:
Cool, will be doing the new release by tomorrow.

spacewander:
@hemanth 
Well, there is a pr to fix a broken regex: https://github.com/tj/git-extras/pull/360
[...]

PR #352, opened 2015-04-16
Revert #325 better changelog

chernjie:
#325 introduces some pretty sweet improvements, but unfortunately there are a couple of backward-incompatible bugs that needs to be addressed. (i.e. #337 and #338)

In order to not hold back a new version release (#336), we should revert this feature and submit a new pull request with it's fixes.

cc: @markeissler, @hemanth, @spacewander


chernjie:
Closing this in favor of #355

PR #354, opened 2015-04-16
Don't treat ignored patterns as regexps

rkennedy:
See issue 353 for details of the bug. This adds the `grep` arguments `-F` (to turn of regex processing) and `-x` (to turn off substring matching) when checking for duplicate entries in *.gitignore*. Both are POSIX arguments, so should be portable.


rkennedy:
This was not the set of commits I intended to include. I meant to include just c68f0d9. Please accept my apologies and ignore this request.

PR #356, opened 2015-04-17
add git-versions

dmil:
Gets all the versions of a particular file from the repository
Places them in a directory called `filename_version` right next the file

Usage 

```
git versions <filepath>
```

Please feel free to message me if there is anything else I need to do to contribute this change back (generate man page etc...)

### DIFF
--- a/bin/git-versions
+++ b/bin/git-versions
@@ -1,0 +1,36 @@
+#!/bin/sh
+# Gets all the versions of a particular file from the repository
+# Places them in a directory called filename_version right next the file
+
+repo_is_clean() {
+  grep_result=`git status | grep "nothing to commit, working directory clean"`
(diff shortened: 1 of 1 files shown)


hemanth:
@dmil Any use case for this?

dmil:
@hemanth I was storing some data, which was updating daily, in a git repo. This let me get all the versions of the file and observe how the data had changed.

hemanth:
@spacewander what say?

spacewander:
IMHO, it doesn't have wide range of use. So I will wait for other people who want to vote for it.

BTW, `git status` doesn't always display in English. It depends on the language setting of the machine.

bl-ue:
I think we should close this PR — there have not been any votes for it.

PR #361, opened 2015-04-25
Fix git tag substring extraction for bash 3.2.

markeissler:
Found it and fixed it. I tested this on:

- CentOS 7
- OSX Bash 4.3.33
- OSX Bash 3.2.57

Output appears to be consistent.


hemanth:
:+1: Thank you!

PR #365, opened 2015-04-30
Don't treat ignored patterns as regexps

### LINKED ISSUE #353: Git-ignore treats arguments as regular expressions instead of file names
Suppose we want to ignore the file *preview.c*:

```bash
git ignore preview.c
```

Now suppose we want to ignore the file *view.c*.

```bash
git ignore view.c
```

But *view.c* isn't actually ignored. The *.gitignore* file contains just one entry:

```
preview.c
```

This is because `git-ignore` uses `grep` to check for duplicate entries before adding new ones. The pattern `view.c` matches the existing `preview.c` line, so the new entry is treated as a duplicate of the old one.

Note that it's not just a substring match. The command `git ignore e.*c` also has no effect because `e.*c` is a regular expression that matches the value `preview.c`.

The command should use the `grep` settings that disable regular expressions and substring matches. After all, once the pattern is added to *.gitignore*, it's not treated as either of those; it's treated as a shell glob.

rkennedy:
Fixes #353.

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -32,7 +32,7 @@ function add_patterns {
   local file="${1/#~/$HOME}"
   for pattern in "${@:2}"; do
     echo "... adding '$pattern'"
-    (test -f "$file" && test "$pattern" && grep -q -- "$pattern" "$file") || echo "$pattern" >> "$file"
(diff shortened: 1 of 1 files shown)


hemanth:
:+1:

PR #368, opened 2015-05-09
Fix some fatal for git-line-summary

zlx:
1. print format when exist special characters
2. special characters in git blame output
3. exist not committed new files will raise `fatal: no such path 'newfilename' in HEAD`

### DIFF
--- a/bin/git-line-summary
+++ b/bin/git-line-summary
@@ -3,13 +3,13 @@
 project=${PWD##*/}
 
 #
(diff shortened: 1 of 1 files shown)


hemanth:
@zix Thanks! Could you please squash the commits?

zlx:
@hemanth of cause!

hemanth:
:+1:

zlx:
Nice. Thank you

PR #370, opened 2015-05-11
git-effort: replace "wc | cut" with "wc | awk"

phigoro:
It looks "cut" may have different behaviour, e.g. BSD one works the way
git-effort doesn't expect. "awk" looks to be one of possible fixes here.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -139,7 +139,7 @@ trap show_cursor_and_cleanup INT
 heading
 
 effort "${files[@]}" | tee $tmp
-test "$(wc -l $tmp | cut -d' ' -f1 )" -gt 1 && sort_effort
+test "$(wc -l $tmp | awk '{print $1}')" -gt 1 && sort_effort
 echo
 
 show_cursor_and_cleanup


phigoro:
A little example to give an idea:
```
$ wc -l bin/git-effort
     145 bin/git-effort
$ wc -l bin/git-effort | cut -d' ' -f1

$ wc -l bin/git-effort | cut -d' ' -f6
145
$ wc -l bin/git-effort | awk '{print $1}'
145
```

phigoro:
I've checked the same with GNU coreutils:
```
$ wc -l bin/git-effort
145 bin/git-effort
$ wc -l bin/git-effort | cut -d' ' -f1
145
$ wc -l bin/git-effort | awk '{print $1}'
145
```
So, "wc" looks to be "the one", i.e. behavior diff between GNU & BSD ones.

hemanth:
Finally, `awk` helps ?

chernjie:
Perhaps we could consider using stdin, it produce consistent result across GNU and BSD, and it does not depend on `cut` or `awk`
```shell
$ wc -l < bin/git-effort
145
```

phigoro:
@chernjie: unfortunately, it still prefixes it with a tab:
```
$ wc -l < bin/git-effort
     145
```

@hemanth: yes, "awk" still helps.

hemanth:
Let's merge this rather than fixing `wc`?

phigoro:
@hemanth: yep, there's nothing to do with "wc" itself, it differs from system to system.

hemanth:
Thanks @phigoro  :+1:

Issue #379, opened 2015-06-01
Silent setup.


hemanth:
As of now for the one line installer we get the below verbose output:

```sh
Cloning into 'git-extras'...
remote: Counting objects: 3124, done.
remote: Compressing objects: 100% (13/13), done.
remote: Total 3124 (delta 3), reused 0 (delta 0), pack-reused 3111
Receiving objects: 100% (3124/3124), 757.68 KiB | 549.00 KiB/s, done.
Resolving deltas: 100% (1699/1699), done.
Checking connectivity... done.
Note: checking out '3.0.0'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at 438de98... Version 3.0.0
Password:
```

We should probably bring it down to:

```sh
Setting up `git-exras`....

Password:
```

What say @spacewander ?

spacewander:
Sure, I will deal with it this weekend!

hemanth:
:+1:

hemanth:
@spacewander Even better idea would be to ask the user to `curl` an URL and pipe it to `bash`?

PR #381, opened 2015-06-07
installation with curl and bash

spacewander:
Does it work well for #379 ?

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -24,5 +24,5 @@ $ [sudo] make install
 One-liner:
 
 ```bash
-$ (cd /tmp && rm -rf git-extras && echo "Setting up 'git-extras'...." && git clone https://github.com/tj/git-extras.git &> /dev/null && cd git-extras && git checkout $(git describe --tags $(git rev-list --tags --max-count=1)) &> /dev/null && sudo make install)
+curl -s https://raw.githubusercontent.com/tj/git-extras/master/install.sh | [sudo] bash /dev/stdin
 ```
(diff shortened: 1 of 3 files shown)


Installation.md
```diff
@@ -24,5 +24,5 @@ $ [sudo] make install
 One-liner:
 
 ```bash
-$ (cd /tmp && rm -rf git-extras && echo "Setting up 'git-extras'...." && git clone https://github.com/tj/git-extras.git &> /dev/null && cd git-extras && git checkout $(git describe --tags $(git rev-list --tags --max-count=1)) &> /dev/null && sudo make install)
+curl -s https://raw.githubusercontent.com/tj/git-extras/master/install.sh | [sudo] bash /dev/stdin
 ```
```
hemanth: That URL is giving a 404, anyway I have shortened the same URL to `http://git.io/git-extras-setup` so if you fix that URL, we can use the shortened URL ;) :+1:

hemanth: Still 404!

hemanth: Well, it took some time to update.

Final version: `curl -sSL http://git.io/git-extras-setup | sudo bash /dev/stdin`  :)

apjanke:
Looks like this has an issue with the `Makefile`: in some cases, `Makefile` will prompt for user input, like if a git alias conflicts with one of the `git-extras` commands. It can't interact with the user in the `curl ... | bash` form, since `stdin` holds the script itself.

The `sh -c` form might work better, because then stdin is still available for user input.

```
bash -c "$(curl -sSL http://git.io/git-extras-setup)"
```

PR #383, opened 2015-06-12
Makefile: Use a more portable mktemp invocation

Somasis:
Some mktemp implementations do not support the `-t` argument. This makes it compatible with both GNU and non-GNU implementations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ install:
 	@mkdir -p $(DESTDIR)$(BINPREFIX)
 	@echo "... installing bins to $(DESTDIR)$(BINPREFIX)"
 	@echo "... installing man pages to $(DESTDIR)$(MANPREFIX)"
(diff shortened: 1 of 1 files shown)


hemanth:
Thanks.

apjanke:
This is breaking the build on OS X for me.

```
$(eval TEMPFILE := $(shell mktemp -q ${TMPDIR:-/tmp}/git-extras.XXXXXX 2>/dev/null || mktemp -q))
```

That `${TMPDIR:-/tmp}` needs to have its `$` escaped as `$$` so it's passed to the shell for evaluation instead of interpreted by `make`. If `make` does it, it evalutes to empty and tries to place the temp file at `/git-extras.XXXXX`.

I've put in a fix: #454.

hemanth:
:cactus: 

Thanks @apjanke :+1:

apjanke:
Thanks. That fixed the build for me.

PR #391, opened 2015-07-07
[RFC] effort: Allow sending options to log. Fix #326

### LINKED ISSUE #326: git-effort on a old codebase
Hi, 

git effort is an awesome tool! :D I am running it on the LibreOffice codebase and I started yesterday and it is still running. The purpose of running it is to find what are the most actively commited files so I can figure out what are the most important files that I need to understand. The problem is the fact that it takes TOOO long.

A proposed solution is to pass as an argument an interval for what commits to check since the latest commits are more relevant. Passing `--since` to `git log` seems like a simple sollution.
```bash
git effort --since='last month'
git effort --since='last year'
```
Do you think this is possible/useful?
Thank you!

nicolaiskogheim:
Putting this up for discussion.

This makes it possible for `git effort` to take other options other than `--above`.
These get sent to `git log` in effort().

So this is possible now:
`git effort --since="one year ago" --until="one week ago" --above 10 mydir/*`

Invalid options is handled by git log, which I think is nice.

Note though: The original issue wanted to have these options to speed up the process by limiting the commit history, but there is no remarkable performance gain by this as far as I were able to see.  

TODO if this is wanted:
 - [x] Man pages
 - [x] Completion

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -107,12 +107,32 @@ sort_effort() {
   < $tmp sort -rn -k 2
 }
 
-# --above <n-commits>
+#
+# parse arguments
+# we handle --above, and send the rest to git log
+#
+while [[ $# > 0 ]] ; do
+  key=$1
+  declare -a log_args
 
-if test "$1" = "--above"; then
-  shift; above=$1
+  case $key in
+    --above)
+            shift; above=$1
+            ;;
+        --*)
+            log_args+=$1
+            ;;
+         --)
+            shift
+            break # files from here on
+            ;;
+          *)
+            break # files from here on
+            ;;
+  esac
   shift
-fi
+done
+
 
 # [file ...]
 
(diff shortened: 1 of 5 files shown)


hemanth:
//cc @spacewander

spacewander:
This is a good idea.

hemanth:
@nicolaiskogheim Please go ahead and complete the PR :)

nicolaiskogheim:
This is finally working as it should now.
Feel free to try it out.

The man page and completion remains. I should have that done soon.

hemanth:
:+1:

nicolaiskogheim:
All done.

nicolaiskogheim:
Hold on a bit. I think I found a way to avoid using `eval`. It's not dangerous to have it, I just think I can make it look a bit cleaner.

hemanth:
Done. Cool :+1:

PR #394, opened 2015-07-15
Handle signals in git-changelog.

### LINKED ISSUE #393: git-changelog should exit with an error code if the editor does
We have a script that wraps around things like `bump`, `git-changelog`, `npm publish`, etc to craft new releases for things. Right now, if I make a mistake in the invocation of this script, it's hard to abort midway.

I discovered that `git-changelog` is optimistic and will ignore the exit code of the editor, even if it's a failure code. I was hoping to force my editor (vim in this case) to exit with a code, and thus abort all subsequent steps of the process.

The point is, if the editor exits with a non-zero, `git-changelog` should propagate that code.

markeissler:
Fixes #393


hemanth:
Thanks! :+1:

PR #400, opened 2015-07-25
effort: don't count untouched files

nicolaiskogheim:
This makes `git effort` able to differentiate between a file that was touched by 1 commit and one that was not touched by a commit.

So if you run  
`git effort --author="Nicolai Skogheim"`  
it will list only files that was touched, and not pretend that every single file has at least one commit touching it.


hemanth:
:+1:

PR #401, opened 2015-07-25
effort: error out on bad value to --above

nicolaiskogheim:
Just because the output is so massive and ugly if you pass anything other than an integer to `--above`, we should exit early with a message.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -140,6 +140,12 @@ do
   esac
 done
 
+# Exit if above-value is not an int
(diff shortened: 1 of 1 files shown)


bin/git-effort
```diff
@@ -140,6 +140,12 @@ do
   esac
 done
 
+# Exit if above-value is not an int
+if [ -z "${above##*[!0-9]*}" ] ; then
```
hemanth: Nice.

PR #408, opened 2015-07-29
contrib: refactor, and conform to git

nicolaiskogheim:
`git contrib` is much more complex than it needs to be.

Moreover, it will now behave like other git commands - output in a pager if that's the default setting for the user, and to STDOUT otherwise.
Also, you could't do `git --no-pager contrib`, but now you can.

I am certain there are even more benefits by doing it this way.  
Edit: Yes, now you can glob on names.
Ex: `git contrib "Nicolai*"`


hemanth:
:+1:

PR #409, opened 2015-07-29
Refactor alias

nicolaiskogheim:
Showing one alias stopped working after 5b1ac3ccc69f65f8670746f5e0b9c14653cf6685
This removes the double dash that was added.

Also some small alterations on the man page for `git alias` while I were at it.


hemanth:
:trophy:

qw3rtman:
What about a67b427? Is this an issue due to different versions of `grep`?

@nicolaiskogheim @go2null

Issue #407, opened 2015-07-29
Time for a new release.


hemanth:
@nicolaiskogheim @spacewander you guys want to do a quick round of testing?

nicolaiskogheim:
I have done some testing, almost done with looking through the files that have been changed since the previous release.  

Here is my progress (list generated with `git diff --name-only 3.0.0..master`
```shell
√ Commands.md
√ Installation.md
√ Makefile
√ Makefile.msys
√ bin/git-alias
√ bin/git-archive-file
√ bin/git-changelog
√ bin/git-contrib
√ bin/git-effort
√ bin/git-extras
√ bin/git-fork
√ bin/git-graft
√ bin/git-guilt
√ bin/git-ignore
√ bin/git-line-summary
bin/git-merge-into
etc/bash_completion.sh
install.cmd
install.sh
man/git-alias.md
man/git-contrib.md
man/git-effort.md
man/git-extras.md
man/git-fork.md
man/git-graft.md
man/git-guilt.md
man/git-line-summary.md
man/git-summary.md
man/index.txt
```

spacewander:
:+1: 
I will run them on Linux this weekend.

spacewander:
The test of Linux part has been done.
Maybe I can do a test with msysgit? I have got access to a Windows computer recently.

hemanth:
Nice! Go ahead please. We shall wait.

spacewander:
I am testing with msysgit, maybe it can be done in this week.

spacewander:
Tested on windows slightly.

| command | status | failure reason |
|----|----| ---- |
|scp|✗|problem with basename |
|fork|✗| problem of curl implementation |
|fresh-branch|✗| break not accepted in case block |
|reset-file|✗| further debug required |
|obliterate|✗| further debug required |
|authors|✗| further debug required |
|changelog|✗| further debug required |
|extras|√||
|squash|√||
|summary|√||
|effort|√||
|commits-since|√||
|count|√||
|create-branch|√||
|delete-branch|√||
|delete-submodule|√||
|delete-tag|√||
|delete-merged-branches|√||
|guilt|√||
|merge-int|√||
|graft|√||
|alias|√||
|ignore|√||
|info|√||
|release|√||
|contrib|√||
|repl|√||
|undo|√||
|gh-pages|√||
|setup|√||
|touch|√||
|feature|√||
|refactor|√||
|bug|√||
|chore|√||
|local-commits|√||
|archive-file|√||
[...]

PR #411, opened 2015-07-31
Add the git-psykorebase command.

Natim:
Git psykorebase let your do a rebase without changing the history and with only one conflict handling commit.

See https://github.com/benoitbryon/psykorebase

### DIFF
--- a/bin/git-psykorebase
+++ b/bin/git-psykorebase
@@ -11,7 +11,7 @@ function current_branch() {
 
 function usage()
 {
-    echo "USAGE: ${PROGRAM} PRIMARY_BRANCH [SECONDARY_BRANCH]"
+    echo "USAGE: ${PROGRAM} PRIMARY_BRANCH [SECONDARY_BRANCH] [--no-ff]"
     echo "USAGE: ${PROGRAM} --continue"
     echo ""
     echo "OPTIONS:"
--- a/bin/git-psykorebase
+++ b/bin/git-psykorebase
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+PROGRAM=$0
+PRIMARY_BRANCH=""
+SECONDARY_BRANCH=""
+FF="--ff"
+CONTINUE="no"
+
+function current_branch() {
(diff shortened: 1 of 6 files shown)


Natim:
Here is my first proposal. r? @hemanth @benoitbryon

hemanth:
LGTM, //cc @spacewander

spacewander:
What about adding documentation for it under `./man`?

bin/git-psykorebase
```diff
@@ -0,0 +1,86 @@
...
+    git rev-parse --abbrev-ref HEAD
+}
+
+function usage()
+{
+    echo "USAGE: ${PROGRAM} PRIMARY_BRANCH [SECONDARY_BRANCH]"
+    echo "USAGE: ${PROGRAM} --continue"
+    echo ""
+    echo "OPTIONS:"
+    echo "  --no-ff:			Force rebase commit."
```
nicolaiskogheim: The usage message doesn't say where the `--no-ff`-option goes :)

bin/git-psykorebase
```diff
@@ -0,0 +1,86 @@
...
+CONTINUE="no"
+
+function current_branch() {
+    git rev-parse --abbrev-ref HEAD
+}
+
+function usage()
+{
+    echo "USAGE: ${PROGRAM} PRIMARY_BRANCH [SECONDARY_BRANCH]"
+    echo "USAGE: ${PROGRAM} --continue"
```
nicolaiskogheim: I would argue that people understand what  ${0} is, so you can drop the variable. But you make the call.

bin/git-psykorebase
Natim: Actually you can put it where you want but I will add it.

Natim:
Updated with the documentation.

hemanth:
Cool :+1:

Natim:
This is great thanks. Can we release a new version of git-extras to get this public?

hemanth:
Yes, after #407 is resolved.

benoitbryon:
Great! Thanks a lot!

PR #413, opened 2015-08-03
changelog: wording in docs

nicolaiskogheim:
I think this is a slightly better way to put it.

Also, I think we could remove most of the examples as many of them are very trivial and well explained in the options section anyway. With me on that?

### DIFF
--- a/man/git-changelog.1
+++ b/man/git-changelog.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "GIT\-CHANGELOG" "1" "April 2015" "" ""
+.TH "GIT\-CHANGELOG" "1" "August 2015" "" "Git Extras"
 .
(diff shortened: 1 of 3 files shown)


markeissler:
I disagree. The wording was meant to explain the potentially confusing legacy behavior where an existing Changelog is appended automatically to the output. I decided to preserve that behavior (along with all other legacy behavior) since git-changelog is often triggered from much larger/comprehensive automated build/release scripts--that is, I didn't want to be the one to break someone's build script. The --prune flag was introduced to override that legacy behavior (if it wasn't for legacy support the --prune mode would have been the default).

As for __ignores__ vs __overrides__, I'm not sure one is better than the other.

Finally, I don't really see the point of removing examples if they promote clarity.

nicolaiskogheim:
Okey, I am with you on the first point.

I chose **overrides** because that's what I usually see in git's man pages, but it's obviously no biggie.

Alright, just wanted to air the subject. Seeing as you have this under control, I am closing this.
Thanks for the reply.

PR #415, opened 2015-08-04
contrib: update docs

nicolaiskogheim:
Add in the docs that `contrib` can match on emails as well as user names, and that the command will return contributions from every user that the search string matches.


hemanth:
:+1: 

We must find a neat way to automate these.

nicolaiskogheim:
Sorry, automate what?

hemanth:
My bad, should have been more specific; Updating Month and Year.

PR #416, opened 2015-08-04
fork: original->upstream in docs

nicolaiskogheim:
The man page for `git fork` said that `git fork` adds the forked repo as a remote called 'original'.
Change that to upstream, as that is what it actually is called.


hemanth:
:+1:

PR #422, opened 2015-08-08
effort: fix active days always 1

nicolaiskogheim:
Introduced in cdb773c06 (Fix error on FreeBSD with process substitution)
when trying to squash a bug.
This fix should be valid for that bug as well.

I have no idea how I missed this. But all is well now.

The problem was that for any input, `git effort` would report `active days` to be 1.


hemanth:
:+1:

PR #423, opened 2015-08-08
effort: document that it takes any path, not just filenames

nicolaiskogheim:
In addition to what the title says
 - Adjust some of the numbers in one of the examples to be realistic
 - Add a note in the docs that the examples omit showing the initial list output by `git effort`, which is the same list, but unsorted.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -54,13 +54,13 @@ color_for() {
 }
 
 #
-# compute the effort of the given <file ...>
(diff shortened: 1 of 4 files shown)


nicolaiskogheim:
hold, on. This looks wrong.

nicolaiskogheim:
Hold on, spotted a typo

hemanth:
:+1:

Issue #269, opened 2014-10-16
ronn is required


jmarca:
suggestion to modify README that ronn is required to make and install this, and that ronn itself installed by: 

```
gem install ronn
```

nicolaiskogheim:
Ronn is only a  dev-dependency. As long as people run `make install` and devs remember to run `make docs` before they commit new stuff, this should't be a problem.

As long as you don't delete the docs, you don't have to build them @jmarca.

jmarca:
Okay.  fine with me.  I just cloned the repo into a clean directory and followed the install docs and everything worked fine.  

So as you say, as long as the "devs" remember to run make docs...

However, isn't the idea of github that everybody can download the source and be a dev?  So again, a short note somewhere saying that ronn is a dependency to make the docs might be helpful.  So how about another link in the main readme, something like:

```

# Regenerating the documentation

The docs are already included in the repository and they should be installed automatically.  If you are a developer and want to expand or add additional documentation, see [Documentation](man/Readme.md)

```


My misstep was running git pull in a local clone repository, rather than deleting it and re-cloning.  After pull, it makes sense to me to run make clean to clear out any old cruft, but that also wipes the docs...

nicolaiskogheim:
The `make clean` only cleans docs. But that should be mentioned somewhere.
If you want to build yourself (assuming docs are up to date), it should suffice to pull new changes and run `[sudo] make install`.

I'm going to do something about this. I understand the confusion. Thank you for reporting this.

ansell:
I am still confused about why ronn (and the rest of the full build instructions) cannot appear in the README.md file. No novice users should ever be required to look at the README.md file, there should be an HTML website for that, which GitHub provides for free if you don't have hosting available. 

Those who do find the README.md file at GitHub can be expected to distinguish between the build dependencies compared to the runtime dependencies. There is no difference on GitHub between "devs" and other people who are interested in the source code, and they may want to contribute and have found that this lack of information is an artificial barrier to that.

qw3rtman:
Now `make` defaults to `make install` (see #425), so unless someone runs `make docs`, this should not be an issue.

@ansell, novice users most likely wouldn't be installing this way anyway (cloning from GitHub and building from source), they would most likely be installing via Homebrew (`brew install git-extras`).

PR #425, opened 2015-08-08
Set default make task to be install

nicolaiskogheim:
The default task for `make` now is `docs`, just because that task is defined earliest in the file.
This commit sets the default task explicitly to `install`. This is to avoid confusing people who want to build git-extras themselves.

This partially adresses #269

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,4 +74,4 @@ docclean:
 	rm -f man/*.1
 	rm -f man/*.html
 
-.PHONY: docs clean docclean install uninstall
+.PHONY: default docs clean docclean install uninstall
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ COMMANDS_USED_WITHOUT_GIT_REPO = git-alias git-extras git-fork git-setup
 COMMANDS_USED_WITH_GIT_REPO = $(filter-out $(COMMANDS_USED_WITHOUT_GIT_REPO), \
 							  $(subst bin/, , $(BINS)))
 
+default: install
+
 docs: $(MAN_HTML) $(MAN_PAGES)
 
 install:


hemanth:
Is there anyway to fix that fully?

nicolaiskogheim:
Sorry, fix what?

17.15 Lør 8. Aug 2015 skrev hemanth.hm <notifications@github.com>:

hemanth:
`ronn` dependency.

nicolaiskogheim:
Do you want to get rid of it, or just make it so that users doesn't have to
know about it?

17.36 Lør 8. Aug 2015 skrev hemanth.hm <notifications@github.com>:

hemanth:
If we can manage to gendocs without it, yes.

spacewander:
So, will we write some scripts to replace `ronn`?

hemanth:
> So, will we write some scripts to replace ronn?

Would be great, but not sure of the complexity.

Makefile
```diff
@@ -11,6 +11,8 @@ COMMANDS_USED_WITHOUT_GIT_REPO = git-alias git-extras git-fork git-setup
 COMMANDS_USED_WITH_GIT_REPO = $(filter-out $(COMMANDS_USED_WITHOUT_GIT_REPO), \
 							  $(subst bin/, , $(BINS)))
 
+default: install
```
stephenmathieson: should be marked as `.PHONY`

nicolaiskogheim:
Sorry for the delay here, guys.
`default` is now marked as `.PHONY` and this is ready to be merged as far as I am concerned.

I don't think we can or should remove ronn. We have to have something to generate our docs, but to create and maintain our own script in here would be wrong.
We just have to do what we can to make sure that regular users doesn't need to be bothered by it.

hemanth:
:+1: Let merge this for now.

PR #426, opened 2015-08-08
abort git-alias when too many arguments given.

spacewander:
Avoid incorrect usage probably because of the missing quotes, like "git
alias cm checkout master"

### DIFF
--- a/bin/git-alias
+++ b/bin/git-alias
@@ -1,7 +1,16 @@
 #!/usr/bin/env bash
 
+usage() {
+cat <<HERE
+usage: git alias                         # list all aliases
+   or: git alias <search-pattern>        # show aliases matching pattern
+   or: git alias <alias-name> <command>  # alias a command
+HERE
+}
+
 case $# in
   0) git config --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort ;;
   1) git alias | grep -e "$1" ;;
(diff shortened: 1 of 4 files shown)


hemanth:
:+1:

nicolaiskogheim:
This is a good idea.

Maybe we could just print a usage message instead?

spacewander:
Just add a "usage" banner

nicolaiskogheim:
I'm sorry I didn't just do this straight away. I was thinking we should do it like i.e `git rebase` does it.

```sh
error: unknown option `fsjljksldf'
usage: git rebase [-i] [options] [--exec <cmd>] [--onto <newbase>] [<upstream>] [<branch>]
   or: git rebase [-i] [options] [--exec <cmd>] [--onto <newbase>] --root [<branch>]
   or: git-rebase --continue | --abort | --skip | --edit-todo

etc..
```

as in

```sh
error: too many arguments
usage: git alias                         # list all aliases
  or: git alias <search-pattern>         # show aliases matching pattern
  or: git alias <alias-name> <command>   # alias a command
```

( I think the comments was nice so I left them in, although git doesn't usually do it that way)

And maybe we should call it `<search-pattern>`, because it is a regex. Then you can edit the man page to say `search-pattern` instead of `search-term` as well.

spacewander:
All right, updated it.

hemanth:
:+1:

PR #429, opened 2015-08-14
merge-into: change --ff option to --ff-only

nicolaiskogheim:
Before you merge this: Should we have a special error message for `--ff`? Currently, we print usage if you pass `--ff` to `git merge-into` **with** this patch.

The best would be if we could accept it silently, as it is the default for `git merge`. But that could confuse users who have adapted to use `--ff` with `git merge-into`.

What do you think?

### DIFF
--- a/bin/git-merge-into
+++ b/bin/git-merge-into
@@ -5,11 +5,11 @@ current_branch() {
 }
 
 usage() {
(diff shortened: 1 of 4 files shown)


hemanth:
:+1:

PR #430, opened 2015-08-16
make it possible to update via `git extras`

spacewander:
Users can't execute git commands in batch file probably, so I put those git commands into update function.


hemanth:
:+1:

PR #432, opened 2015-08-25
fix link of `git feature|refactor|bug|chore`

stevemao:


Commands.md
```diff
@@ -30,7 +30,7 @@
  - [`git setup`](#git-setup)
  - [`git touch`](#git-touch)
  - [`git obliterate`](#git-obliterate)
- - [`git feature|refactor|bug|chore`](#git-feature|refactor|bug|chore)
+ - [`git feature|refactor|bug|chore`](#git-featurerefactorbugchore)
```
hemanth: It would be better if it's hyphenated.

stevemao: Unless we change the title?

stevemao: In fact, it shouldn't matter because 
1. You don't wanna change your title for this.
2. The link is not visible.
3. GitHub doesn't change `|` to `-`.

hemanth: Makes sense.

PR #433, opened 2015-08-26
Git clear

grindhold:
git-clear is like git reset --hard  but it also removes all untracked files that reside inside the working directory. Clears the repo as if it was freshly cloned at the current HEAD-commit.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -6,6 +6,7 @@
  - [`git effort`](#git-effort)
  - [`git authors`](#git-authors)
  - [`git changelog`](#git-changelog)
+ - [`git clear`](#git-clear)
  - [`git commits-since`](#git-commits-since)
  - [`git count`](#git-count)
  - [`git create-branch`](#git-create-branch)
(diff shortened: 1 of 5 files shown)


spacewander:
IMHO, we can do the same thing with an alias:
```
alias gclear="git clean -d -f -x && git reset --hard"
```

hemanth:
Yes, that looks better, btw we add such one liners to http://git.io/git-tips ;)

grindhold:
ok

hemanth:
@grindhold Hold! We didn't mean to close the PR, we can have `git clean -d -f -x && git reset --hard` instead of [this](https://github.com/tj/git-extras/pull/433/files#diff-027e63237a1ce78ab40f11082c9fc6c3R4) what say @spacewander ?

grindhold:
ah okay, i presumed git-clear was kind of too simple and obvious to be included into git extras ;)

grindhold:
i use the command from above to implement git-clear now.

hemanth:
LGTM @spacewander what say?

spacewander:
Maybe a clearer prompt will be better.
"Sure?" is not easy to understand in a glance, we need to tell  the user what this command will do.

grindhold:
@spacewander now it's a more understandable question

spacewander:
Good job.

hemanth:
:+1:

PR #439, opened 2015-09-07
Add git-ignore-io

Lee-W:
Add git ignore-io to generate sample gitignore file from [gitignore.io](https://git.gitbook.com/lee-w/freebsd-and-system-administration.git)


bin/git-ignore-io
```diff
@@ -0,0 +1,124 @@
+#/bin/sh
```
spacewander: Should it be `#!/usr/bin/env bash`?

bin/git-ignore-io
```diff
@@ -0,0 +1,124 @@
...
+gi_list=`cat $default_path | tr "," "\n"`
+
+update_gi_list() {
+    curl -L -s  "${gitignore_io_url}/list" > ~/.gi_list
+}
+
+print_in_alphabetical_order() {
+    for i in {a..z};
+    do
+        echo "$gi_list" | grep "^$i" | tr "\n" " "
```
spacewander: As I see, the default order of "https://www.gitignore.io/api/list" is in alphabetical.
IMHO, even if we don't believe in the default order, sorting the items in `"$gi_list"` is better.

bin/git-ignore-io
```diff
@@ -0,0 +1,124 @@
+#/bin/sh
+
+gitignore_io_url="https://www.gitignore.io/api/"
+
+default_path="$HOME/.gi_list"
+gi_list=`cat $default_path | tr "," "\n"`
```
spacewander: Another way to get `gi_list`:
`gi_list=$(sed 's/,/\n/g' "$default_path")`

Or:
`gi_list=$(sed 's/,/ /g' "$default_path")`. It seems no item contains whitespace.

bin/git-ignore-io
```diff
@@ -0,0 +1,124 @@
...
+    coulumn_width=25
+
+    counter=0
+    for item in $gi_list; do
+        printf "%-${coulumn_width}s" $item
+        counter=$(($counter+1))
+        if [[ $(($counter%$table_width)) -eq 0 ]]; then
+            echo
+        fi
+    done
```
spacewander: Just found a better way to print it in table format:
```
echo "$gi_list" | xargs -n 5 | column -t
```

Commands.md
```diff
@@ -408,6 +409,54 @@ build
 *.log
 ```
 
+## git ignore-io
+
+Generate sample gitignore file from [gitignore.io](https://www.gitignore.io)  
+
+Without option, `git ignore` show the sample gitignore on screen.
```
spacewander: `git ignore` shows global gitignore if option not given.
`git ignore-io` prints usage if option not given.

.gitignore
```diff
@@ -0,0 +1,37 @@
+# Created by https://www.gitignore.io/api/vim,osx
```
spacewander: Please remove the test file.

man/git-ignore-io.md
```diff
@@ -0,0 +1,77 @@
+git-ignore-io(1) -- Get sample gitignore file
+================================
+
+## SYNOPSIS
+
+`git ignore-io` [OPTIONS]
+
+## DESCRIPTION
+Get sample gitignore file from [gitignore.io]("https://www.gitignore.io")
```
spacewander: Url surrounded by quote is invalid markdown syntax.

bin/git-ignore-io
Lee-W: The first method cannot run on my Mac. I found the solution on this [Replace comma with newline in sed](http://stackoverflow.com/questions/10748453/replace-comma-with-newline-in-sed).

Since I'm not the author of gitignore.io, I cannot promise there won't be one. Thus, I use new line as delimiter.

Commands.md
Lee-W: Thanks. I modified it.

bin/git-ignore-io
Lee-W: Thanks for your advice. But, I'm wondering why sorting would be better in this case. Since I use grep to get types from a to z, it seems whether it's ordered or not doesn't matter.

bin/git-ignore-io
Lee-W: It looks great!!!
Thanks!

.gitignore
Lee-W: Ok!

man/git-ignore-io.md
Lee-W: It's fixed. Thanks,

bin/git-ignore-io
Lee-W: Both works on my Mac.

bin/git-ignore-io
spacewander: #Today I Learn#  Maybe it is another difference between GNU coreutils and the BSD one.

bin/git-ignore-io
spacewander: All right.

bin/git-ignore-io
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env bash
+
+gitignore_io_url="https://www.gitignore.io/api/"
+
+default_path="$HOME/.gi_list"
+gi_list=`cat $default_path | tr "," "\n\n"`
```
spacewander: I catch a duplicate '\n'! `tr "," "\n"` is enough.

bin/git-ignore-io
```diff
@@ -0,0 +1,114 @@
...
+            print_last_modified_time
+            ;;
+        -u|--update-list)
+            update_gi_list
+            ;;
+         -s|--search)
+            search $2
+            ;;
+        -L)
+            print_in_alphabetical_order
```
spacewander: The doc says -L will print in table format.
But actually it prints in alpabetical order? :confused:

bin/git-ignore-io
Lee-W: Oh, I was trying to use sed and forgot to change it back. I'll update latter. Thanks~

bin/git-ignore-io
Lee-W: Sorry for that. It's been fixed.

bin/git-ignore-io
```diff
@@ -0,0 +1,114 @@
...
+    do
+        if [[ $type == *$1* ]]
+        then
+            echo $type
+        fi
+    done
+}
+
+print_last_modified_time() {
+    local gi_list_date=`stat -f "%t%Sm" $default_path`
```
spacewander: Just tested this command on my machine(Linux).
Unfortunately, the `-f` means file system on GNU coreutils: http://linux.die.net/man/1/stat .
We have to find another way to do this thing.
Please GNU and BSD at the same time, is one of the difficulty to write a portable shell script.
Keep it up!

Lee-W: I found a solution on this [Best Way to Get File Modified Time in Seconds](http://stackoverflow.com/questions/17878684/best-way-to-get-file-modified-time-in-seconds).
Thanks for reporting.

bin/git-ignore-io
```diff
@@ -0,0 +1,120 @@
...
+    gi $1 >> .gitignore
+}
+
+show_usage() {
+    echo "Usage:"
+    echo "  git ignore-io <types>...                    Show gitignore template"
+    echo "                [-a|--append] <types>...      Append new .gitignore content to .gitignore under the current directory"
+    echo "                [-e|--export] <types>...      Export new .gitignore to the current directory (The old one will be replaced)"
+    echo "                [-l]                          Print available types in alphabetical order    "
+    echo "                [-L]                          Print available types in table format"
```
spacewander: Forget to update here? Or just haven't finished the update?
If you get all things done, please @ some and let us know.

Lee-W: @spacewander 
Sorry, I didn't notice that.
It confused me as well.
Thus, I add full name for both listing options.

spacewander:
I think this pr is ok.

nicolaiskogheim:
```patch
From e760c190151a4539c2fb37503011cff790666f5c Mon Sep 17 00:00:00 2001
From: Nicolai Skogheim <nicolai.skogheim@gmail.com>
Date: Tue, 15 Sep 2015 00:14:33 +0200
Subject: [PATCH] ignore-io: minor fixes

* Redirect STDERR to /dev/null
* Remove call to `cat`
* Change `cmd` to $(cmd)
---
 bin/git-ignore-io | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/bin/git-ignore-io b/bin/git-ignore-io
index 5aa3fd6..42053ae 100755
--- a/bin/git-ignore-io
+++ b/bin/git-ignore-io
@@ -3,7 +3,7 @@
 gitignore_io_url="https://www.gitignore.io/api/"
 
 default_path="$HOME/.gi_list"
-gi_list=`cat $default_path | tr "," "\n"`
+gi_list=$(tr "," "\n" < "$default_path" 2>/dev/null)
 
 update_gi_list() {
     curl -L -s  "${gitignore_io_url}/list" > ~/.gi_list
-- 
2.5.2
```

I tested your command and got an error from `cat` because of a missing file, hence these little changes.
What do you think?

nicolaiskogheim:
Also, try and run [`shellcheck`](https://github.com/koalaman/shellcheck) on the file.

Lee-W:
@nicolaiskogheim

Lee-W:
@nicolaiskogheim Both method cannot run on my mac. So, I check the existence of ~/.gi_list first and then set this variable.
Thanks for sharing spellcheck. It's awesome!

PR #441, opened 2015-09-08
remove duplicate checkout

spacewander:
And also print checkout info to user.
I think it is no need to conceal checkout message. The message is not unexpected. And when `checkout` is failed(as uncommited change, existed branch, etc.), people can know what happened.


hemanth:
:+1:

PR #444, opened 2015-09-13
add alias conflict prompt

spacewander:
Since the alias conflict detect need to run `git config --get-regexp 'alias.*'`, and we can't access to git command in the Windows CMD, I just implement the prompt in Makefile.


hemanth:
:+1:

PR #445, opened 2015-09-13
Straighten up effort

nicolaiskogheim:
This commit does mainly two things:
1. Simplify logic, making it easier to understand the code and add to it
2. Change the synopsis of the command so that we are less likely to be needing to break it later.

Tested only on OSX with Bash 4.3


hemanth:
:+1:

PR #458, opened 2015-10-02
Makefile: refactor is-git-repo inclusion logic

apjanke:
Here's a suggestion for refactoring the Makefile to simplify it. This avoids the need for a big repeated block of code for the command-construction logic, where the two blocks differ by only one line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,10 @@ BINS = $(wildcard bin/git-*)
 MANS = $(wildcard man/git-*.md)
 MAN_HTML = $(MANS:.md=.html)
 MAN_PAGES = $(MANS:.md=.1)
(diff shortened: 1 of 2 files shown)


hemanth:
:+1:

apjanke:
Thanks!

nicolaiskogheim:
Sweet!

dwijnand:
This patch does the wrong thing. `COMMANDS_USED_WITHOUT_GIT_REPO` doesn't mean "commands that use is_git_repo, and should have it included in the "built" version of the command".

For fallout see #493.

apjanke:
Oh, jeez. Sorry about that. Here's a PR that fixes the sense of the test: #494

PR #459, opened 2015-10-02
git-effort: use portable terminal escape sequences

apjanke:
The current code is using hardcoded xterm style escape sequences for its coloring. This switches to using `tput` capabilities, so it'll work with other terminals. More readable, too, IMHO.

Also uses "default color" instead of "black" for the low-activity things, so it still shows up on colorschemes that have dark backgrounds. In my terminals using Solarized Dark, the output wasn't readable.


hemanth:
I wonder how this when unnoticed. 

Thanks!

PR #460, opened 2015-10-02
git-summary: protect against character encoding issues with LC_ALL=C

apjanke:
On OS X 10.9.5, I get a couple sed errors in `git-summary-line`.

These look like character encoding issues.

```
[~/local/opp/omz/oh-my-zsh on ⇄ git-extras-compatibility]
$ git-summary --line

 project  : oh-my-zsh
sed: RE error: illegal byte sequence
sed: RE error: illegal byte sequence
 lines    : 34168
 authors  :
 3938 Andrew Janke              11.5%
 3374 Felipe Contreras          9.9%
```

Running `sed` under `LC_ALL=C` makes the errors go away. This PR does that.

### DIFF
--- a/bin/git-line-summary
+++ b/bin/git-line-summary
@@ -9,7 +9,7 @@ single_file() {
   while read data
   do
     if [[ $(file "$data") = *text* ]]; then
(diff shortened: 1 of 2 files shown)


hemanth:
:+1:

PR #465, opened 2015-10-08
Fix color

spacewander:
1. use tput instead of ansi code sequences.
2. disable color if the output is not printed to tty.


hemanth:
:+1:

PR #468, opened 2015-10-13
use real purplish

spacewander:
Get the purplish color code from here:
https://en.wikipedia.org/wiki/ANSI_escape_code#Colors


hemanth:
:+1:

PR #469, opened 2015-10-17
Use the default installation path in install.cmd

JanSchulz:
That's the same path as cmdr checks when starting up: https://github.com/cmderdev/cmder/blob/master/vendor/init.bat#L31

With this patch you can run `install.cmd` without any arguments. Adding arguments still overwrites the path.


hemanth:
Thanks!

PR #471, opened 2015-10-18
git-changelog: option to only use merges

JanSchulz:
using `-m` or `--merges-only` will only show the merge
commit messages. It will also use the body of the merge
commit message in addition to the subject, as github merges
usually have a useless subject line for commit messages.

If used with `-n/--no-merges`, the last one specified will win.

Will need regenerated man pages (.html + .1)

Closes: https://github.com/tj/git-extras/issues/470

PR #472, opened 2015-10-18
Also display body for normal commits

jankatins:
Changes the default format for normal commits to include the body of a commit message.

### DIFF
--- a/bin/git-changelog
+++ b/bin/git-changelog
@@ -3,7 +3,7 @@
 DEF_TAG_RECENT="n.n.n"
 GIT_LOG_OPTS="$(git config changelog.opts)"
 GIT_LOG_FORMAT="$(git config changelog.format)"
-[[ -z "$GIT_LOG_FORMAT" ]] && GIT_LOG_FORMAT='  * %s'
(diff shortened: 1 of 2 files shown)


JanSchulz:
As requested in #471

qw3rtman:
Could you make this specific to merge commits instead of all commits?

For normal commits, the title should be enough; however, the title for a merge commit it "Merge ...", which isn't very descriptive.

JanSchulz:
You mean that if I start it up without any of `--no-merges|--merges-only` it will display the merges with body and the normal commits without body?

I don't think this will be possible (or at least not easy), as as far as I understand git log, it can't be done in one pass: at least I haven't seen a "for merges print body, for rest only subject" kind of formating string. You would need to either reorder the logs (aka two got log calls, first for the merges with body (`git log --merges`), then for the rest of the commits (`git log --no-merges`)) or one would need to compute the whole thing yourself (printout as machine readable thing, filter for "Merged...", remove the body from the rest...

Unfortunately my bash and git log foo isn't good enough for any of the solutions :-( -> probably just close this PR?

PR #474, opened 2015-10-19
Rewrite `install.cmd`

gisphm:
1. `install.cmd` uses multiple subroutines and labels.
2. `install.cmd` provides two options to install `Git Extras`. 
3. The first is the standard installer. This option requires administrator's privilege to look `GitForWindows's InstallPath` up in registry table. Once the `InstallPath` is found, the scripts of `Git Extras` will be copied into it.
4. The Second is the portable way. User needs to manually type where `GitForWindows` installs. Of course, user can use tab key to complete the path in cmd.
5. The copy section in `:JUSTCOPY` has integrated the scripts which need `is-git-repo` in the for loop.
6. I have tested it on my windows machine(Windows 10 pro 64bit) and it works well. But I am not sure whether it works on other windows machines or not. It needs test!

### DIFF
--- a/install.cmd
+++ b/install.cmd
@@ -1,32 +1,128 @@
 @ECHO OFF
-SET PREFIX=C:\SCM\PortableGit\mingw64
-:: default for Git for Windows 2.x
-@if exist "%ProgramFiles%\Git" (
-    set PREFIX="%ProgramFiles%\Git\mingw64"
-)
-IF NOT "%~1"=="" SET PREFIX="%~1"
-SET HTMLDIR=%PREFIX%\share\doc\git-doc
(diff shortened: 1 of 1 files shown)


hemanth:
:+1:  

Let take some help here to get it tested on other versions of windows.

gisphm:
Well, I find it still needs some modifications so that `git extras update` can execute successfully.

gisphm:
Well, I give up. Due to my poor knowledge of batch file and the weird syntax, I finally find it difficult for me to rewrite a new one. I will make another pr about install.cmd -- refactoring the for loop to reduce duplicate code.

Issue #478, opened 2015-10-21
git-ignore: add way to set the global gitignore file


JanSchulz:
first of all: thanks for this command, I didn't even know that there is a global gitignore file.

Unfortunately I had to read the source code to see how to enable it:

```
λ git ignore -g .idea/
Adding pattern(s) to: .idea/

There is no Global .gitignore yet
```

-> How do I enable a global gitignore file and where should I place it?

It would be nice if there was a "create" flag (or it would simple do that if the file is not there), which would put the file into e.g. `~/.gitignore` and set the global setting...

spacewander:
Maybe we should use `Use 'git config --global core.excludesfile' to create your global .gitignore first` instead of `There is no Global .gitignore yet`. So that users can know what to do.

I think it is better to let user define their global ignore file.

JanSchulz:
```
Can't find global .gitignore.

Use 'git config --global --add core.excludesfile ~/gitignore' to set the path to your global .gitignore file to '~/gitignore'.
```
?

I can do a PR for that. If so: `~/.gitignore` or `~/gitignore` (hidden or not)? Or somewhere different?

spacewander:
@JanSchulz 
`~/.gitignore_global` is a common choice.

JanSchulz:
PR in #479

raviranjananalist:
For Expert 5 Packers and Movers Hyderabad go to  
http://www.expert5th.in/packers-and-movers-hyderabad/
For Expert 5 Packers and Movers Bangalore go to 
http://www.expert5th.in/packers-and-movers-bangalore/
For Expert 5 Packers and Movers Mumbai go to 
http://www.expert5th.in/packers-and-movers-mumbai/

raviranjananalist:
For Expert 5 Packers and Movers Gurgaon go to 
http://www.expert5th.in/packers-and-movers-gurgaon/
For Expert 5 Packers and Movers Delhi go to 
http://www.expert5th.in/packers-and-movers-delhi/
For Expert 5 Packers and Movers Pune go to  
http://www.expert5th.in/packers-and-movers-pune/

raviranjananalist:
For Expert 5 Packers and Movers Chennai go to 
http://www.expert5th.in/packers-and-movers-chennai/
For Expert 5 Packers and Movers Noida go to 
http://www.expert5th.in/packers-and-movers-noida/
For Expert 5 Packers and Movers Thane go to 
[...]

PR #479, opened 2015-10-22
Add more info how to set global gitignore

JanSchulz:
Closes: #478

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -14,7 +14,15 @@ function show_global {
 }
 
 function add_global {
-  add_patterns `git config --global core.excludesfile` "$@"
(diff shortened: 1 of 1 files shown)


bin/git-ignore
```diff
@@ -14,7 +14,15 @@ function show_global {
 }
 
 function add_global {
-  add_patterns `git config --global core.excludesfile` "$@"
+  local global_gitignore=$(git config --global core.excludesfile)
+  if [ "x${global_gitignore}" == "x" ]; then
```
spacewander: we can use a modern style of empty string check:
`-z "$global_gitignore"` or `"$global_gitignore" = ""`

hemanth:
Thanks.

PR #480, opened 2015-10-23
Fixed the Debian installation commands in Installation.md

dylanaparker:
The command was incorrect


spacewander:
`$apt_pref` is a variable. It might refer to `apt-get` or `aptitude` or something else.

Issue #467, opened 2015-10-13
`git touch` with multiple filenames does not work as expected


al-the-x:
Little gotcha:

    $> git touch A B
    $> git status -s
    A  A B

WTF? Internally, the `git-touch` script uses `filename="$*"` and `touch "$filename"`, which turns the argument list `A B` into `"A B"`: a _single_ file named `"A B"` not _two_ files named `A` and `B`).

Patch / PR accepted?

al-the-x:
I ask because I've already fixed for _my_ usage, but I'd like to contribute that upstream.

qw3rtman:
From `man bash`, under the section "Special Parameters":

* `$*` is equivalent to `"$1c$2c..."`
* `$@` is equivalent to `"$1" "$2" ...`

So `A B` becoming `"A B"` is expected.

Regardless, I think `git touch` should follow the functionality you expect, so I would vouch if you sent in a PR. :+1:

al-the-x:
Happy to submit a PR... It's an easy fix.

al-the-x:
Do you mind leaving this issue open until I have a PR to replace it (and link it appropriately)? I don't want to lose track of this issue...

raviranjananalist:
For Expert 5 Packers and Movers Hyderabad go to  
http://www.expert5th.in/packers-and-movers-hyderabad/
For Expert 5 Packers and Movers Bangalore go to 
http://www.expert5th.in/packers-and-movers-bangalore/
For Expert 5 Packers and Movers Mumbai go to 
http://www.expert5th.in/packers-and-movers-mumbai/

raviranjananalist:
For Expert 5 Packers and Movers Gurgaon go to 
http://www.expert5th.in/packers-and-movers-gurgaon/
For Expert 5 Packers and Movers Delhi go to 
http://www.expert5th.in/packers-and-movers-delhi/
For Expert 5 Packers and Movers Pune go to  
http://www.expert5th.in/packers-and-movers-pune/

raviranjananalist:
For Expert 5 Packers and Movers Chennai go to 
http://www.expert5th.in/packers-and-movers-chennai/
For Expert 5 Packers and Movers Noida go to 
http://www.expert5th.in/packers-and-movers-noida/
For Expert 5 Packers and Movers Thane go to 
http://www.expert5th.in/packers-and-movers-thane/
For Expert 5 Packers and Movers Navi Mumbai go to 
[...]

PR #481, opened 2015-10-31
Make `git-touch` behave like `touch` when passed multiple files

al-the-x:
Re #467:

* [x] use `$@` instead of `$#`
* [x] test that arguments have been passed (`$#`)
* [ ] maybe print usage message?
* [ ] test revisions

### DIFF
--- a/bin/git-touch
+++ b/bin/git-touch
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
-test $# -lt 1 && echo "filename required" 1>&2 && exit 1
+USAGE(){
+  echo "git touch <filename> [ <filename> . . .]" 1>&2
+  exit 1
+}
 
-filenames="$@"
+test $# -lt 1 && USAGE
 
 for filename in $@; do
   touch "$filename" \
--- a/bin/git-touch
+++ b/bin/git-touch
@@ -4,5 +4,7 @@ test $# -lt 1 && echo "filename required" 1>&2 && exit 1
 
 filenames="$@"
 
-touch "$filenames" \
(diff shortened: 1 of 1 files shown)


al-the-x:
I haven't tested this yet, but I have metering a commit so that I could open a PR as a reminder.

hemanth:
Let us know once it's good to go.

qw3rtman:
@al-the-x, is this in progress?

al-the-x:
Ready for review @qw3rtman. Please provide feedback.

bin/git-touch
```diff
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
-filename="$*"
+USAGE(){
```
qw3rtman: Why capitalized?

qw3rtman: No need to write another function.

bin/git-touch
```diff
@@ -1,8 +1,13 @@
...
+  echo "git touch <filename> [ <filename> . . .]" 1>&2
+  exit 1
+}
 
-test -z "$filename" && echo "filename required" 1>&2 && exit 1
+test $# -lt 1 && USAGE
```
qw3rtman:
```shell
test $# -lt 1 && echo "usage: git touch <filename> [<filename> ...]" 1>&2 && exit 1
```

qw3rtman:
Everything looks good, except you don't need to write another function to display usage (it's only used once). See my line notes.

qw3rtman:
Looks great, thanks! :+1:

Issue #237, opened 2014-07-14
make install fails for different install locations of `bash_completion`


jbnicolai:
Running on OS X, having installed `bash_completion` through [`Homebrew`](http://brew.sh/) my completion directory is **not** `/etc/bash_completion.d/`. 

Instead it is `/usr/local/etc/bash_completion.d/`, or more general: `$(brew --prefix)/etc/bash_completion.d/`

Running `make install` I get the following error:

```
... installing git-summary
... installing git-touch
... installing git-undo
... installing git-unlock
cp -f man/git-*.1 "/usr/local/share/man/man1"
mkdir: /etc/bash_completion.d: Permission denied
make: *** [install] Error 1
```

This is because `/etc/` is owned by root, `/etc/bash_completion.d/` does not exist (nor should it!) and it is trying to mkdir as my current user.

I suggest adding in a check that _if_ the user has `brew` and _if_ `$(brew --prefix)/etc/bash_completion.d/' exists that then that directory is used, rather than the current hardcoded default.

Would a pull request for this be welcome?

fibo:
Same problem if I try to set a custom location

    $ PREFIX=/my/custom/path make install

Also PREFIX and DESTDIR looks like repeated information, if I try

    $ DESTDIR=/my/custom/path PREFIX=/my/custom/path make install

also does not work.

hemanth:
Strange. //cc @spacewander  @chernjie

revolter:
I get this:

```
... installing git-extras
... installing git-fork
... installing git-setup
cp -f man/git-*.1 /usr/local/share/man/man1
cp -f etc/bash_completion.sh /etc/bash_completion.d/git-extras
cp: /etc/bash_completion.d/git-extras: Permission denied
make: *** [install] Error 1
/usr/local/bin/git-extras: line 30: syntax error near unexpected token `&'
/usr/local/bin/git-extras: line 30: `& cd "$orig" \'
```

when I try to `git extras update`, which I guess is running `make install` under the hoods too.

If it is of any help, this is what I get when I `brew install git-extras`:

```
==> Downloading https://downloads.sf.net/project/machomebrew/Bottles/git-extras-2.2.0.yosemite.bottle.tar.gz
Already downloaded: /Library/Caches/Homebrew/git-extras-2.2.0.yosemite.bottle.tar.gz
==> Pouring git-extras-2.2.0.yosemite.bottle.tar.gz
==> Caveats
Bash completion has been installed to:
[...]

Issue #487, opened 2015-12-27
[RELEASE] v4.0.0


qw3rtman:
Hey all! It's time for a new release, `v4.0.0`!

Here's a list of all the changes made in this release; the full changelog can be found in [`History.md`](https://github.com/tj/git-extras/blob/master/History.md). Please let me know if there are any errors below.

* (#365) prevent `git ignore` from avoiding files it thinks are already in `.gitignore`, but aren't – @rkennedy
* (#368) fix printing special characters (from `git blame` output) in `git line-summary` – @zlx
* (#369) fix hashbangs for `git guilt` and `git merge-into` – @phigoro
* (#370) move from `wc` to `awk` in `git-effort` to avoid `wc`'s different behavior on diffent platforms – @phigoro
* (#372) add installation one-liner that installs latest release from GitHub – @spacewander
* (#373) add documentation for `git line-summary` – @zlx
* (#377) fix `mktemp` behavior on Linux in `Makefile` – @go2null
* (#377) fix `grep` syntax in `git contrib` – @go2null
* (#378) update installation in `git-extras update` – @spacewander
* (#380) keep installation silent – @spacewander
* (#381) modify installation to work with `curl` and `bash` – @spacewander
* (8a3f988) create shorter URL for one-liner installation – @hemanth
* (#383) use more portable `mktemp` implementation in `Makefile` – @somasis
* (#385) add `install.cmd` for installation on Windows – @gisphm
* (#386) use https for `git fork` – @rstacruz
* (#392) improve speed of `git effort` – @nicolaiskogheim
* (#390) fix `git effort` color formatting with `--above` – @nicolaiskogheim
* (#394) make `git changelog` respect status code of editor – @markeissler
* (#396) make sure `git changelog` always looks at `HEAD` – @tocker
* (#399) cleaner way to current branch in `git archive-file` – @nicolaiskogheim
* (#398) require destination branch in `git graft` – @nicolaiskogheim, @muhtasib
* (#388) update all man pages – @chernjie
* (#391) extend `git effort` to accept `git log` options, such as `--since` – @nicolaiskogheim
* (#400) prevent `git effort` from counting untouched files – @nicolaiskogheim
* (#401) make `git effort` exit early to prevent messy error – @nicolaiskogheim
* (#403) write documentation for `git effort --since` – @andreicristianpetcu
* (#406) fix `git extras`, `git changelog`, `git info`, and `git alias` on Windows (`msysgit`) – @southly, @nicolaiskogheim
[...]

Issue #463, opened 2015-10-03
Want an updated Makefile for BSD/Homebrew/non-root?


apjanke:
Hi folks,

I've just discovered git-extras through my work with Oh My Zsh and Homebrew and I'm digging it; I'd like to start using it regularly. I'm mainly an OS X/Homebrew user with a multi-user, multi-machine setup, and I'm running in to a few issues on this platform. I've got space for a weekend project. Would you be interested in a PR that refactored the `Makefile` and build process to do some or all of:

* Make sure the whole build, including doco generation, works under BSD/OS X
* Support relocating `etc` (#237)
* Support disabling selected commands and suppressing prompting for alias conflicts (#442, #228)
 * (will make fully unattended, repeatable installs possible)
 * Remove man & doco for disabled commands, just to be tidy
* Split the build and install into two steps, so you could build the commands (slurping in the library files) and run them inside the local source tree without having to install to the final destination (useful for iterating development)
* Install the HTML and other format docs under `/usr/share/doc` (right now it looks like they're going under `man`, which should really only be for man-formatted stuff)
* Unify `makefile.msys` with the main `Makefile` (#427)

Regardless, cheers for the good work.

spacewander:
Waiting for it :smile:

qw3rtman:
My only concern is that the build and install into two steps should be opt-in and not opt-out, so that the installation process doesn't become more complicated.

Otherwise, looking forward to seeing a PR from you. Thanks! :+1:

apjanke:
Yeah, it'll be opt-in and transparent to users: `make install` will do the same thing as it does now, and do the `build` step internally as a dependency. This will just give developers the option to request just a local build with a `make build` or similar command.

qw3rtman:
Sounds like a good idea to me! :smile:

raviranjananalist:
For Expert 5 Packers and Movers Hyderabad go to  
http://www.expert5th.in/packers-and-movers-hyderabad/
For Expert 5 Packers and Movers Bangalore go to 
http://www.expert5th.in/packers-and-movers-bangalore/
For Expert 5 Packers and Movers Mumbai go to 
http://www.expert5th.in/packers-and-movers-mumbai/

raviranjananalist:
For Expert 5 Packers and Movers Gurgaon go to 
http://www.expert5th.in/packers-and-movers-gurgaon/
For Expert 5 Packers and Movers Delhi go to 
[...]

PR #489, opened 2015-12-27
Makefile: allow customizing of SYSCONFDIR (/etc) location

apjanke:
The current Makefile hardcodes the `/etc` installation location. Users might want to install `etc` files to other places, like `/usr/local/etc` for non-root installs or `brew --prefix` for Homebrew. GNU software and Homebrew locate `SYSCONFDIR` (the conventional Makefile variable for the `/etc` dir location) under `PREFIX` by default, even.

This PR introduces a `SYSCONFDIR` variable to the Makefile to allow the user to customize the `/etc` location. It leaves the default as `/etc`, not changing it to `$(PREFIX)/etc`, so the current default behavior is preserved. I think it's worth discussing whether to change to locate `etc` under `$(PREFIX)` by default; that would allow non-root or fully-relocated installs by just specifying `PREFIX`, and would match the behavior of a lot of other software.

This will help users installing as non-root (as in #237), and remove the need for [one of the patches](https://github.com/Homebrew/homebrew/blob/41025fb70fcab7a8d379cc41558bc1133869c3d7/Library/Formula/git-extras.rb#L23) in the Homebrew `git-extras` installation formula.

I've tested this on OS X.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PREFIX ?= /usr/local
 BINPREFIX ?= "$(PREFIX)/bin"
 MANPREFIX ?= "$(PREFIX)/share/man/man1"
+SYSCONFDIR ?= /etc
 BINS = $(wildcard bin/git-*)
 MANS = $(wildcard man/git-*.md)
 MAN_HTML = $(MANS:.md=.html)
@@ -47,8 +48,8 @@ install:
 		cp -f man/git-*.1 $(DESTDIR)$(MANPREFIX); \
 		echo "cp -f man/git-*.1 $(DESTDIR)$(MANPREFIX)"; \
 	fi
-	@mkdir -p $(DESTDIR)/etc/bash_completion.d
-	cp -f etc/bash_completion.sh $(DESTDIR)/etc/bash_completion.d/git-extras
+	@mkdir -p $(DESTDIR)$(SYSCONFDIR)/bash_completion.d
+	cp -f etc/bash_completion.sh $(DESTDIR)$(SYSCONFDIR)/bash_completion.d/git-extras
 
 man/%.html: man/%.md
 	ronn \
@@ -72,7 +73,7 @@ uninstall:
 		echo "... uninstalling $(DESTDIR)$(MANPREFIX)/$(notdir $(MAN))"; \
 		rm -f $(DESTDIR)$(MANPREFIX)/$(notdir $(MAN)); \
 	)
-	rm -f $(DESTDIR)/etc/bash_completion.d/git-extras
+	rm -f $(DESTDIR)$(SYSCONFDIR)/bash_completion.d/git-extras
 
 clean: docclean
 


Makefile
```diff
@@ -1,6 +1,7 @@
 PREFIX ?= /usr/local
 BINPREFIX ?= "$(PREFIX)/bin"
 MANPREFIX ?= "$(PREFIX)/share/man/man1"
+SYSCONFDIR = /etc
```
qw3rtman: Should this be `?=` instead of `=`?

`SYSCONFDIR ?= /etc`

apjanke: Yeah, I'll switch it to `?=` to match the others.

qw3rtman:
Thanks!

Could you document this in the installation instructions?

apjanke:
Sure. I'm going to take a stab at another Makefile revision per #463; I'll include updated installation instructions.

What do you think about sticking `etc` under `$PREFIX`? It would be a change from current behavior, but would simplify the installation process for relocated or non-root installs.

qw3rtman:
@apjanke, do you mean change `$PREFIX` to be `/usr/local/etc`?

apjanke:
No, I mean change `$SYSCONFDIR` to be `$(PREFIX)/etc` by default instead of `/etc`. This way everything that is installed goes under `$PREFIX` by default.

```
PREFIX ?= /usr/local
SYSCONFDIR ?= $(PREFIX)/etc
```

qw3rtman:
Would this still play nicely with Homebrew? Also, is the simplification you mentioned syntax-related (meaning typing `$(SYSCONFDIR)` instead of `$(PREFIX)/$(SYSCONFDIR)`)?

Apologies if my question doesn't make much sense; I'm not all too versed with Makefiles.

apjanke:
> Would this still play nicely with Homebrew and follow GNU standards?

Yes. This would simplify the commands needed in the Homebrew formula. And it is in line with the Makefile Conventions in GNU's Coding Standards, which say that `sysconfdir` should default to `$(prefix)/etc`. (In [the "Directory Variables" section](http://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables); scroll down to "sysconfdir".) The difference between us and them would be that we use an upper-case `$PREFIX`, which isn't unusual, and I think makes sense for a variable that is inherited from the environment.

>  is the simplification you mentioned syntax-related?

The simplification is for users who are installing `git-extras`; it simplifies the installation command for when they want everything under a different location, including the `etc` dir. Right now, they have to specify both `PREFIX` and `SYSCONFDIR`. Changing it, they could supply just `PREFIX`.

Before:

```
$ make install PREFIX=/usr/local SYSCONFDIR=/usr/local/etc
```

After:

```
$ make install PREFIX=/usr/local
```

The flip side of this is that if they want to install the commands to a non-default prefix, but still want the bash completion definitions (and configuration files, if you add some in the future) to go to the system `/etc`.

```
$ make install PREFIX=/usr/local SYSCONFDIR=/etc
```

So it depends on which case you'd like to make easier. I think relocating sysconfdir along with everything else is the more common case: this is what you need to do for a non-root installation, or an installation local to a user's home directory. It supports full side-by-side installation of multiple versions of the package. It's the normal case for GNU software, or stuff packaged with the GNU autotools chain. And I think it's what a lot of users would expect, since it's the default for the GNU stuff.

qw3rtman:
Thanks for clarifying! I'm all for it!

I'd like to get this out in the next major release (#487) if possible. By when do you think you can have this implemented (and added to installation instructions)?

apjanke:
By tomorrow. It's an easy change.

apjanke:
Here you go: #490.

qw3rtman:
Merged #490! Could you also update whatever is needed in the Homebrew formula?

apjanke:
Thanks!

Not much point in updating the Homebrew formula until the new release is made: we still need the more complicated `make` call in the formula for the current release, so adding a special case to do the simpler `make` call for the `--HEAD` version would wind up complicating the formula overall.

Once the new release is out, I'll update the Homebrew formula.

qw3rtman:
Sounds good. Thanks, @apjanke!

PR #490, opened 2015-12-28
installation: put $SYSCONFDIR (etc) under $PREFIX by default

apjanke:
This changes the installation process to default to using the `etc` dir under `$PREFIX`, instead of the system `/etc` dir. This is consistent with GNU's installation practices, and simplifies the installation commands for non-root user installations and similar.

See discussion under #489 for details.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -1,17 +1,23 @@
-## Debian
+# Installing git-extras
+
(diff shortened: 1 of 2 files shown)


qw3rtman:
Great, thanks! :+1:

PR #491, opened 2015-12-28
Revamp makefile

### LINKED ISSUE #228: Install only some commands?
Is there a way to install only _some_ of the extra commands?

I would really like this feature, as I see absolutely no value in commands like `git create-branch` or `git feature`, but I see much value in others.

### LINKED ISSUE #442: Is possible to disable some commands?
I install git-extras from brew.
And I already map git-pr and git-pull-request to my own command. But this overrides mine.
Because I use Stash often instead of Github.
Is there a way to disable some of these, or make my command (alias in ~/.gitconfig) have a high priority?
Or just delete, rename those conflict ones from `/usr/local/bin`?
Below is my script in git global config file
```shell
[alias]
      pull-request = "!f() { git fetch $1 refs/pull-requests/$2/from:pull-requests/$2; }; f"
      pr = "!f() { git pull-request origin $1; git checkout pull-requests/$1; }; f"
```

### LINKED ISSUE #463: Want an updated Makefile for BSD/Homebrew/non-root?

apjanke:
Work-in-progress. Please don't merge yet.

This PR revamps the build process and Makefile as suggested in #463. It gives the user more configuration options for customizing the installation. This will also be useful for packagers like Homebrew, which currently needs to patch the distribution. And it sets things up for local execution and testing of the commands.

This is a good-sized change, so no reason to merge it right away. We can review/discuss/test first.

Closes #463
Closes #442 
Closes #228
- Adds an intermediate `make build` target that generates working commands in a local `build/` dir, so you can test it without installing it
  - `make install` continues to work as before, automatically calling `make build` if needed. (You will need to manually `make clean` if you change the source code between repeated installs from the same source directory.)
- Adds an option to disable the self-update feature, which is needed by external package managers like Homebrew
- Adds an option to disable selected commands (as requested in #442, #228)
- Adds an option to have it skip all or install all commands that conflict with aliases, without prompting. With this you can do a fully noninteractive installation
- Revises manpage generation code to be consistent between the Makefile and `man/manning-up.sh`.
- Generates HTML for the non-man-page documentation, too, and installs it to `$PREFIX/share/doc`
- Folds the windows-specific stuff into `install.sh` and removes the `updateForWindows()` special case

The separate `Makefile.msys` still exists; I don't have a setup for testing that.


hemanth:
:+1:

apjanke:
Thanks hemanth!

Okay, hold on: I figured out an easy way to do per-file builds for the commands and fix the updated-dependency problem, so they get properly rebuilt on subsequent `make install`s even without intervening `make clean|build`. Let me roll that in before you form firm opinions...

apjanke:
Okay, I've added per-file builds and fixed the dependency problem.

This new mechanism does away with the hardcoded logic for including libraries that was in the Makefile or main command building script and replaces it with `#@@ include: @@` directives, so you can see a command's dependencies when looking at the command's source code itself. It then adds a conditional-inclusion directive to support doing the "disable self-update" with this mechanism.

hemanth:
@spacewander Got some time for this?

spacewander:
I am reading the diff now. It may take me some time to make a feedback.

man/Readme.md
```diff
@@ -1,64 +1,63 @@
...
 html: ./git-effort.1.html                +man
-
-$ mv git-effort.1.html git-effort.html
+$ git add git-effort.*
 ```
 
 ## SHELL SCRIPT
 
-Alternatively you can run the `manning-up.sh` automated shell script included in the man folder. The script will recreate the git-extras index based on the list of .md files available before it runs `ronn` against each one to generate the documents as well as renaming the generated `.html` files to their desired form.
+The `manning-up.sh` shell script included in the `man/` folder contains the logic to . The script will recreate the git-extras index based on the list of `git-*.md` files and then run `ronn` against each one to generate the documents, as well as renaming the generated `.html` files to their desired form.
```
spacewander: Contains the logic to what?

man/manning-up.sh
```diff
@@ -1,7 +1,29 @@
...
+#  - Regenerates the ROFF and HTML files for all the `git-*.md` files
 
+if [[ $# > 1 ]]; then
+  echo "error: $0 takes 0 or 1 arguments"
+fi
+
+# 1-arg form
+if [ -n "$1" ]; then
+  ronn --manual "Git Extras" $1 && mv -f ${1/.md/}.1.html ${1/.md/}.html && exit 0
+  exit 1
```
spacewander: I think `exit 0` for success is more practical. Consider this case:
```
./manning-up.sh git-file.md && git add git-file.html
```

man/manning-up.sh
```diff
@@ -1,7 +1,29 @@
...
+#
+# With one argument, generates the .1 and .html files from the given .md file
+#
+# With no arguments, updates all generated document files:
+#  - Updates the index inside git-extras.md
+#  - Generates index.txt
+#  - Regenerates the ROFF and HTML files for all the `git-*.md` files
 
+if [[ $# > 1 ]]; then
+  echo "error: $0 takes 0 or 1 arguments"
```
spacewander: I'd like to exit the script after it said something error happened.

tools/stick_in_includes
```diff
@@ -0,0 +1,17 @@
...
+	destfile=$file
+	# Propagate the shebang line
+	head -3 "$srcfile" > "$destfile"
+	# Add common includes invocation
+	#echo '' >> $destfile
+	echo '#@@ include: helper/is-git-repo.sh @@' >> $destfile
+	# Rest of file
+	# And then the rest of the original source file
+	tail -n +4 "$srcfile" >> "$destfile"
+done
```
spacewander: Shall we safely throw it away now?

tools/render_gh_markdown
```diff
@@ -0,0 +1,38 @@
...
+# TODO: Figure out how to generate anchors. The index links in Commands.md 
+# aren't working when rendered locally with this, though they do work when
+# viewed on GitHub.
+
+require 'optparse'
+require 'github/markup'
+
+options = {}
+OptionParser.new do |opts|
+	opts.banner = "Usage: render_gh_markdown.rb <file> ..."
```
spacewander: Better use `$0`. For example:
`Usage:  #{$0} <file>`

tools/render_gh_markdown
```diff
@@ -0,0 +1,38 @@
...
+
+infiles = ARGV
+
+if infiles.empty?
+	$stderr.puts("Error: no files specified")
+	abort
+end
+
+for infile in infiles do
+	basefile = infile.gsub(/\.[^.]*$/, '')
```
spacewander: If my guess about the regex is correct, this line does the same as `basefile = File.basename(infile, File.extname(infile))` ? It will remove the last part of extension.

tools/build_commands
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
```
spacewander: Is the script still in used?

spacewander:
@apjanke  
I suggest rewriting `tools/build_command` from perl to bash. IMHO, we should make `git-extras` only depend on bash and other coreutils. Perl is out of scope.

man/manning-up.sh
apjanke: Oops... this is a merge error; the `exit 1` should go with the "error: $0 takes 0 or 1 arguments" case, not the normal 1-arg form.

apjanke: Oh, wait, no it isn't. The `exit 0` is at the end of the chain of operations in the previous line, line 21.

```
ronn --manual "Git Extras" $1 && mv -f ${1/.md/}.1.html ${1/.md/}.html && exit 0
```

This `exit 1` only gets hit if something in the pipeline failed.

tools/build_commands
apjanke: No. If people are okay with the switch to single-file command building, it can be removed. The rest of its functionality is replaced by `tools/install_commands`.

tools/render_gh_markdown
apjanke: Yes. Do you prefer the `basename`/`extname` form?

man/Readme.md
apjanke: Fixed.

apjanke:
I've rebased the branch on master and addressed most of the line notes.

man/manning-up.sh
spacewander: Oh sorry, I missed the trailing exit.

tools/render_gh_markdown
spacewander: I thought it again, the `basename/extname` is not surely better than regex. It may be just my personal preference.
Therefore, keep the regex form is ok.

apjanke:
> I suggest rewriting `tools/build_command` from perl to bash. IMHO, we should make git-extras only depend on bash and other coreutils. Perl is out of scope.

Hmm. Sorry; I wasn't really thinking about dependencies when I wrote this.

You could probably re-write `build_command` as a `sed` command with a `bash` wrapper around it, especially if you removed the conditional-inclusion feature. I'll see if that's doable.

qw3rtman:
Hey, just wanted to check in on the status of this.

We're planning on having a release tomorrow (it's been 1 month since the last release), but we can push this to the next monthly release if it's not ready.

apjanke:
Not ready, or close to it. Let's wait until next monthly release.

qw3rtman:
Sure, moved it to the next milestone. Thank you so much for this contribution!

PR #494, opened 2016-01-01
Makefile: fix inverted list of commands that use is_git_repo

### LINKED ISSUE #493: "Not a git repo!" when use git-setup
I installed git-extras by Homebrew.

apjanke:
PR #458 inverted the sense of the test for whether is-git-repo should be included in a file. This fixes it.

Fixes #493

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ install:
 			echo "... installing $(COMMAND)"; \
 			head -1 bin/$(COMMAND) > $(TEMPFILE); \
 			cat $(LIB) >> $(TEMPFILE); \
-			if grep "$(COMMAND)" need_git_repo >/dev/null; then \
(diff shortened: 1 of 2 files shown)


dwijnand:
Nice one :+1:

hemanth:
Thanks.

spacewander:
Have it fixed the bug in v4.0.0? Or users have to update their version to get this bugfix? 
I am not sure about the mechanism of Homebrew...

apjanke:
Homebrew only uses labeled releases. When a new v4.0.1 (or "v4.0.0a" or whatever) is made and released, Homebrew will be updated to pick it up. Until then Homebrew users will still have v4.0.0, which has the bug.

spacewander:
@apjanke
Thanks, I get it now.

PR #475, opened 2015-10-19
Update install.cmd

gisphm:
1. Reduce duplicate code in the previous two for loops and copy bash_completion.sh to where it should be installed as the makefile describes.
2. Those commands in file `need_git_repo` actually don't need `is-git-repo`.

PR #496, opened 2016-01-05
Make the win installer more robust

JanSchulz:
User visible:
* Also accept the git install path (and not the mingw64 subdir) as a user
  supplied install location. Make it the new default and warn the user if
  a mingw64 dir is supplied (=this dir is not checked, so it might be
  anywhere...)
* Find git.exe in path and use that as an install location
* Add error|status messages and output less noise (robocopy)
* do some check if column.exe is not found...
* adjust the installation text for windows...

not user visible:
* more.exe does not work when used in a cmd which has a unicode codepage
  -> save the current codepage and set a sane default during the install
* do some "magic" with quotes around path variables -> makes it work with
  pathes with spaces in it (both final git dir and the git-extra checkout)
* Do not leak variables to the sourounding interpreter

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -25,21 +25,21 @@ git clone https://github.com/tj/git-extras.git
 git checkout $(git describe --tags $(git rev-list --tags --max-count=1))
 ```
 
-After that, execute the `install.cmd` in the command prompt, with the path your
-`Git for Windows 2.x` is installed. If you installed git as admin, you need to 
-run this promt as admin, too. Per default, it finds a `Git for Windows 2.x` in
-`%ProgramFiles%\Git`, which is the default install location. The fallback is
-`C:\SCM\PortableGit`. If you didn't install git into one of these dirs, you have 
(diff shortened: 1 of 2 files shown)


JanSchulz:
This probably clashes with https://github.com/tj/git-extras/pull/475 

CC: @gisphm

JanSchulz:
BTW: would it make sense to ask the git for windows people to include column.exe in the installer?

gisphm:
@JanSchulz I think you have made a better one.:smile: 
What I am trying to do in #475 is to combined the two "for" loops together. You can use it or not.
It would be great if 'column.exe' is distributed with `Git for Windows`.

JanSchulz:
I added a bug in git for windows to include column.exe: https://github.com/git-for-windows/git/issues/586

gisphm:
:+1:

hemanth:
:+1:

JanSchulz:
column.exe seems to get included in the git for windows installer: https://github.com/git-for-windows/build-extra/pull/92

PR #497, opened 2016-01-06
Update windows install.cmd

JanSchulz:
* Fix a regression ("!" needs to be escaped when `enabledelayedexpansion` is used -> this destroyed the shebang line and resulted in not useable commands...)
* check for write rights to the install directory
* add information about column.exe, which will be included in the next Git for Windows release


JanSchulz:
If the rest of this PR is controversial, please at least cherry pick the first commit, as this fixes a brown paper bug in my last PR which currently makes all installed scripts unusable on windows :-(

qw3rtman:
Thanks, @JanSchulz!

PR #500, opened 2016-01-13
add git-sync

kumon:
This command syncs local branch with remote branch.

This runs:
```
git fetch <remote-name> <branch-name>
git reset --hard <remote-name>/<branch-name>
git clean -d -f -x
```

### DIFF
--- a/bin/git-sync
+++ b/bin/git-sync
@@ -44,16 +44,20 @@ function main()
     shift
   done
 
+  local remote_branch
   if [ "${remote}" = "" ]; then
-    local remote_branch="$(git rev-parse --abbrev-ref --symbolic-full-name @{u})"
-    branch="$(git rev-parse --abbrev-ref --symbolic-full-name @)"
-    remote="${remote_branch%%/${branch}}"
+    if ! remote_branch="$(git rev-parse --abbrev-ref --symbolic-full-name @{u} 2>/dev/null)"; then
+      echo "There is no upstream information of local branch."
+      exit 1
+    fi
+    local branch="$(git rev-parse --abbrev-ref --symbolic-full-name @)"
+    local remote=$(git config "branch.${branch}.remote")
   elif [ "${branch}" = "" ]; then
     echo -e "Error: too few arguments.\n"
     _usage
     exit 1
   else
-    local remote_branch="${remote}/${branch}"
+    remote_branch="${remote}/${branch}"
   fi
 
   echo "Are you sure you want to clean all changes & sync with '${remote_branch}'? [y/N]"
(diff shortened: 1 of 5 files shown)


bin/git-sync
```diff
@@ -0,0 +1,61 @@
...
+Example:
+  ${command} origin master
+EOS
+}
+
+function main()
+{
+  while [ "$1" != "" ]; do
+    case $1 in
+      -h | --help | help | "?" )
```
spacewander: `help` can be a legal name of a remote branch.

kumon: Thank you for reviewing it. I should fix that.

spacewander:
What about using a default remote/branch when `remote/branch` argument is missing?
Just follow the convention of `git fetch/pull`.

kumon:
I will follow the convention of `git fetch/pull`. It's more practical. Thank you so much.

bin/git-sync
nicolaiskogheim: As we don't use "?" for seeing usage with any other commands, we shouldn't do it here either.
`?` is a special character in some terminals with `extglob` (i think) enabled, and could confuse people as well.

kumon: I will remove '?'. Thank you!

kumon:
I changed to use upstream of local branch by default, when remote/branch is not specified.

bin/git-sync
```diff
@@ -0,0 +1,73 @@
...
+        fi
+        ;;
+    esac
+    shift
+  done
+
+  if [ "${remote}" = "" ]; then
+    local remote_branch="$(git rev-parse --abbrev-ref --symbolic-full-name @{u})"
+    branch="$(git rev-parse --abbrev-ref --symbolic-full-name @)"
+    remote="${remote_branch%%/${branch}}"
```
spacewander: A little quibble:
`git rev-parse --abbrev-ref --symbolic-full-name @{u}` will return nothing if there is not related upstream branch. 
Then users will see `... & sync with ''? [y/N]` and the command may simply reset to current branch, which is not really wanted.

kumon:
I added upstream check that is based on exit status of `git rev-parse --abbrev-ref --symbolic-full-name @{u}`.

spacewander:
@hemanth  @qw3rtman 
It looks good to me. What about your idea?

qw3rtman:
Yup, looks good to me! Merging.

In the future, I think it'd be a good idea to standardize how help/usage is displayed for each command. Each command should have the same name for the function that displays the help/usage. Additionally, the formatting of the usage should be the same.

kumon:
Thank you for reviewing it!

qw3rtman:
Thanks for contributing, @kumon! :smile:

PR #512, opened 2016-02-16
Fix ignore-io searching bug

Lee-W:
The original searching in "git ignore-io" command used wrong the variable and thus malfunctioned.
The bug is solved.


hemanth:
:+1:

PR #530, opened 2016-04-21
added graph

Keloran:
Added a single command for graph online, rather than having to remember it all


spacewander:
IMHO, it's better to keep as a git alias instead of a separate command. The command is too tiny.

hemanth:
Yup, sorry, need to close this.

Issue #531, opened 2016-04-26
Clear removes ignored files


Addvilz:
Pretty self-explanatory. Clear removes all untracked files, even the ones ignored. That probably shouldn't happen, or at least there should be a flag for that.

hemanth:
There is a prompt for the same reason with that [command](https://github.com/tj/git-extras/blob/master/bin/git-clear#L3)

Addvilz:
Yes, I get that, but the command would be infinitely more useful with option of not deleting whatever is ignored.

hemanth:
Feel free to do a PR 😇

Addvilz:
@hemanth  the point of opening the issue was to inform about that there is, in my opinion, an issue. If I will find time to implement a fix, I most certainly will do so, however, if someone is willing to address this before me, the info is here.

PR #533, opened 2016-05-01
Add git clear-soft, modify git clear to indicate explicit removal of files

Addvilz:
This could be one potential solution to issue #531 - a separate command that allows for "soft" cleanup of the repository honoring `.gitignore`.

This PR also includes modified description and notice on `git clear` about that it does not honor `.gitignore`.

I was swinging between this and modifying the existing command with a flag, but this somehow seemed a cleaner solution.


hemanth:
Thank you!

PR #544, opened 2016-06-27
Remove myself from AUTHORS

sorbits:
This removes my email from AUTHORS as I do not want it there.

This rewrites two commits that change AUTHORS and must be force-pushed to GitHub to erase the old history (that has my email).

Suggested instructions for “merging” this change:

    git remote add -f sorbits git@github.com:sorbits/git-extras.git
    # Verify there is only one change (to AUTHORS)
    git diff origin/master sorbits/master
    git push -f origin sorbits/master:master


sorbits:
When can I expect you to remove my email address?

nicolaiskogheim:
Hi, and sorry for taking so long. We shall try to be swift about this, but this demands a little planning from our side so I will need to discuss how to approach this. The discussion is mainly targeted at the contributors who, in the end, will have to clean up the mess we've created, but if you @sorbits, or anyone else have questions we should consider, or experience with doing this (force pushing to an active remote branch with multiple PRs pointed at it) it's of course welcome. 

---------

I'm not entirely sure about how to best do this, but it seems there are two options. Either way we will have to force push rewritten history which will break all pull requests.

So let's say I force push the rewritten history. Now all pull requests will be conflicting, and it will look like everyone is trying to submit about 50 commits. This will confuse those not familiar with the situation.

Our options:

1. We can ask pull requesters to fix their PR, by pulling down the new master, and re-applying their commits onto that, in the branch the PR is pointing at. This may be uncomfortable for certain users.


2. We as contributors can close all PRs, and label them for re-opening perhaps, and create new PRs with the history fixed. This may or may not be more work on our hand. We have to it properly, though; use the commits and not just make the changes manually. We wouldn't want to "steal" commits of people by accidentally having the author id being set to some of ours.

There is maybe a third option: Let's say we write a little text explaining what has happened, and post it in every open PR. It would go something like "fix this and update the PR or ask for help" and we could provide commands for doing it, but advise that they let us do it if they're afraid to mess things up.

We have about 20 open PRs, but most of them are old and should be closed. This is an opportunity get rid of dead PRs, and we could always cherry-pick stuff if we want it.

-----

Another issue is what to do with the 584 forks we invalidate. I think it warrants an explanation in the Readme at the least. People can either delete the fork and fork anew, or `git reset --hard upstream/master` and juggle their way from there if they have commits they want to preserve.

szpak:
@nicolaiskogheim In general pushing with force to master is a very bad idea for every project. You mentioned PRs, but there are also hundreds of forks with changes made by other people. In addition, a lot of people for sure have local clones of this repo and after the next "git pull" they would get very strange and confusing error message. Having local changes it will be really hard to merge them with the new upstream (after the history rewrite) and it definitely will not help project which has problems with external contributions.

@sorbits In the Internet things made publicly tend to stay in that or another form (e.g.  compromising photos). Even rewriting this repository lefts 500+ other forks. At least part of them have that ill-fated commit. I bet that you will not be able to convince all those people to do the same. Apart from that there is RSS feed for every GitHub repository and that information has been propagated to many external systems aggregating data from GitHub (here's a [story](https://it.slashdot.org/story/15/01/02/2342228/bots-scanning-github-to-steal-amazon-ec2-keys) about stealing EC2 token even after the author rewrite the history with "push force").

To sum it up. IMHO this is (technically) a very bad idea to do push with force to master, especially that the expected results (removing email address definitely) will not be achieved. I would propose to just remove an email in one "normal" commit to not have it available in the current version.

nicolaiskogheim:
@szpak I did mention the forks, but you pointed out something that slipped my mind which is very important here, and that is that whatever we do here, the commit will still exist in the forks.

sorbits:
I created PR #564 to remove my email from AUTHORS without rewriting history.

Though had I added someone’s non-public email address to a repository that I control I would not hesitate to rewrite history. For semi-experienced git users it should not be difficult to handle rewritten history.

PR #545, opened 2016-06-27
fix git-fork by removing '.git'

spacewander:


bin/git-fork
```diff
@@ -16,6 +16,7 @@ read user
 # extract owner + project from repo url
 project=${1##*/}
 owner=${1%/$project}
+[[ $project == *.git ]] && project=${project%.git}
```
nicolaiskogheim: You don't need to do the test here. Just `project=${project%.git}` is fine.
If the suffix is there, it will be removed; if not, nothing happens.

hemanth:
Let us merge this for now, as it fixes a breaking commit.

Issue #518, opened 2016-03-18
Rename author in history


adius:
It would be cool if there was a way to rename an author/committer in the history:
[help.github.com/articles/changing-author-info](https://help.github.com/articles/changing-author-info)

Maybe there should also be flags to switch between renaming an author, a committer or both (standard should be both).

hemanth:
PR welcomed :)

tardypad:
@nicolaiskogheim This one can be closed now that my PR is merged

nicolaiskogheim:
Yes 🎉

PR #548, opened 2016-07-13
Add git reauthor

tardypad:
Add new reauthor command: Rewrite history to change author's identity.
Will close issue #518
See the examples within the Commands.md file for some use cases

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -38,6 +38,7 @@
  - [`git psykorebase`](#git-psykorebase)
  - [`git pull-request`](#git-pull-request)
  - [`git rebase-patch`](#git-rebase-patch)
+ - [`git reauthor`](#git-reauthor)
  - [`git release`](#git-release)
  - [`git rename-tag`](#git-rename-tag)
  - [`git repl`](#git-repl)
@@ -347,6 +348,7 @@ Does the following:
   - Push the branch / tags
   - Executes _.git/hooks/post-release.sh_ (if present)
 
+
 ## git rename-tag
 
 Rename a tag (locally and remotely).
@@ -369,6 +371,37 @@ To git@myserver.com:myuser/myrepository.git
 $ git tag
 test2
 ```
+
+
+## git reauthor
+
+Rewrite history to change author's identity.
+
+Change manually the personal email and name of Jack to his work ones
+```bash
+$ git reauthor --old-email jack@perso.me --correct-email jack@work.com --correct-name 'Jack Foobar'
+```
+
+Change automatically the email and name of Jack to the ones defined in the Git config
+```bash
+$ git reauthor --old-email jack@perso.me --config
+```
+
+Fix only the email of Jack (keep the name already used)
+```bash
+$ git reauthor --old-email jack@perso --correct-email jack@perso.me
+```
+
+Change only the committer email of Jack (keep the author email already used)
+```bash
+$ git reauthor --old-email jack@perso.me --correct-email jack@work.com --type committer
+```
+
+Set Jack as the only author of the whole repository
+```bash
+$ git reauthor --all --correct-email jack@perso.me --correct-name Jack
+```
+
 
 ## git alias
 
--- a/Commands.md
+++ b/Commands.md
@@ -384,7 +384,7 @@ $ git reauthor --old-email jack@perso.me --correct-email jack@work.com --correct
 
 Change automatically the email and name of Jack to the ones defined in the Git config
 ```bash
-$ git reauthor --old-email jack@perso.me --config
+$ git reauthor --old-email jack@perso.me --use-config
 ```
 
 Fix only the email of Jack (keep the name already used)
--- a/Commands.md
+++ b/Commands.md
@@ -377,17 +377,17 @@ test2
 
 Rewrite history to change author's identity.
 
-Change manually the personal email and name of Jack to his work ones
+Replace the personal email and name of Jack to his work ones
 ```bash
 $ git reauthor --old-email jack@perso.me --correct-email jack@work.com --correct-name 'Jack Foobar'
 ```
 
-Change automatically the email and name of Jack to the ones defined in the Git config
+Replace the email and name of Jack to the ones defined in the Git config
 ```bash
 $ git reauthor --old-email jack@perso.me --use-config
 ```
 
-Fix only the email of Jack (keep the name already used)
+Replace only the email of Jack (keep the name already used)
 ```bash
 $ git reauthor --old-email jack@perso --correct-email jack@perso.me
 ```
@@ -397,9 +397,14 @@ Change only the committer email of Jack (keep the author email already used)
 $ git reauthor --old-email jack@perso.me --correct-email jack@work.com --type committer
 ```
 
-Set Jack as the only author of the whole repository
+Set Jack's identity as the only one of the whole repository
 ```bash
 $ git reauthor --all --correct-email jack@perso.me --correct-name Jack
+```
+
+Set Jack as the only committer of the whole repository (keeps authors)
+```bash
+$ git reauthor --all --correct-email jack@perso.me --correct-name Jack --type committer
 ```
 
 
--- a/bin/git-reauthor
+++ b/bin/git-reauthor
@@ -7,7 +7,7 @@ init_variables() {
   OLD_EMAIL=
   CORRECT_EMAIL=$( git config user.email )
   CORRECT_NAME=$( git config user.name )
-  TARGET='both'
+  TARGET='all'
 }
 
 
@@ -24,7 +24,7 @@ Optional options
     -n, --correct-name   <name>   override the correct name to set
                                   default is Git user.name config value
     -t, --target         <id>     define the target of the replacement
-                                  author, committer, both (default)
+                                  author, committer, all (default)
 EOF
 }
 
@@ -57,7 +57,7 @@ reauthor() {
   case "${TARGET}" in
     author) filter="${author}" ;;
     committer) filter="${committer}" ;;
-    both) filter="${author} ${committer}" ;;
+    all) filter="${author} ${committer}" ;;
   esac
 
   export OLD_EMAIL
@@ -111,7 +111,7 @@ validate_options() {
     error "${msg}"
   fi
 
-  if [[ "${TARGET}" != 'both' ]] \
+  if [[ "${TARGET}" != 'all' ]] \
      && [[ "${TARGET}" != 'author' ]] \
      && [[ "${TARGET}" != 'committer' ]]; then
     error "invalid target '${TARGET}'"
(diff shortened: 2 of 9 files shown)


qw3rtman:
I think this is a great addition!

However, whenever I try to run it, it errors with `error: invalid option 'ACe:hn:o:t:'` followed by the usage. I'm on OS X and `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)`.

/cc @hemanth @spacewander

qw3rtman:
Upon further investigation, it seems like this error is caused by the difference between `BSD getopt` (which comes on Macs) and `GNU getopt`. [This link](http://stackoverflow.com/questions/2721946/cross-platform-getopt-for-a-shell-script) suggests using `getopts` instead for proper compatibility.

nicolaiskogheim:
I'll just chime in here very quick: Yes, @qw3rtman, you are absolutely right. `getopt` should under no circumstance be used for anything.

tardypad:
I had doubt when I added `getopt` because I didn't find here any other command using it. I see why now. Unfortunately I don't want to use `getopts` either because it doesn't support long option names...
It's not a problem, I'll update this and do the parsing manually.

nicolaiskogheim:
:)

If you need inspiration on option parsing you can take a look at [this](https://github.com/tj/git-extras/blob/master/bin/git-effort#L143-L168).

tardypad:
I've just pushed the manual parsing.
Should be fine now hopefully.

hemanth:
LGTM!

bin/git-reauthor
```diff
@@ -0,0 +1,154 @@
...
+}
+
+
+usage() {
+  cat << EOF
+usage: git ${COMMAND} [<options>]
+
+Options
+    -A, --all                     rewrite all commits and tags
+    -C, --config                  define correct values from user Git config
```
nicolaiskogheim: This option should be called `--use-config`, or maybe even `--use-git-config`.
Also, why capital C?

bin/git-reauthor
```diff
@@ -0,0 +1,154 @@
...
+  TYPE='full'
+}
+
+
+usage() {
+  cat << EOF
+usage: git ${COMMAND} [<options>]
+
+Options
+    -A, --all                     rewrite all commits and tags
```
nicolaiskogheim: It must be evident that this option rewrites every single commit and tag in the repo. Can we write `all` in all caps? "rewrite ALL commits and tags"

bin/git-reauthor
```diff
@@ -0,0 +1,154 @@
...
+    -C, --config                  define correct values from user Git config
+    -e, --correct-email  <email>  define the correct email to set
+    -n, --correct-name   <name>   define the correct name to set
+    -o, --old-email      <email>  rewrite only commits and tags from old email
+    -t, --type           <id>     define the type of the rewrite
+                                  author, committer, full (default)
```
nicolaiskogheim: Is it more intuitive to say `both`, rather than `full`?

man/git-reauthor.md
```diff
@@ -0,0 +1,83 @@
+git-reauthor(1) -- Rewrite history to change author's identity
```
nicolaiskogheim: Suggestion:
```
Replace author and/or committer id on commits and tags.
```

man/git-reauthor.md
```diff
@@ -0,0 +1,83 @@
...
+==============================================================
+
+## SYNOPSIS
+
+`git reauthor [<options>]`
+
+## DESCRIPTION
+
+Rewrite history to change author's identity.
```
nicolaiskogheim: This line is the same as next to the `Name` field. Should be removed.

We have a tradition of sorts of repeating the short description (the one next to the name) in the description, but I have never seen anyone else do that. I think we should stop doing it.

man/git-reauthor.md
```diff
@@ -0,0 +1,83 @@
...
+
+Goes through all existing commits and tags to selectively modify the identities:  
+- modify name and/or email  
+- affect author and/or committer identities  
+- specify the new name/email as parameters or use the ones from the Git config  
+- change all objects or only replace those from a specific identity (defined by its email)
+
```
nicolaiskogheim: Describe how the utility works, rather than listing its features. You may have looked at some of our other man-pages, like the one for `git-effort`, but I think the description there is somewhat lacking too, so you should use man pages from git  as a guidelines instead. Unix utilites also have very good descriptions.

It's not easy, and require some thinking. Please ask if you get stuck, or push your suggestions and you will get a new review by someone.

Come to think of it, this is something we probably should mention in our upcoming Contributing Guidelines (#549).

man/git-reauthor.md
```diff
@@ -0,0 +1,83 @@
...
+  -n, --correct-name &lt;<name>&gt;
+
+    Define the correct name to set
+    Empty name '' is not allowed
+
+  -o, --old-email &lt;<email>&gt;
+
+    Rewrite only commits and tags from old email
+    Empty email '' is allowed
+    If the --all flag is specified as well, it takes precedence over this option
```
nicolaiskogheim: I think this can be slightly dangerous. Should we error out if both `--all` and `--old-email` are specified?
If someone specifies both it means they don't fully understand how the command works, and if it is the `--all` part they don't understand they have now rewritten more commits than they bargained for. If they then push to a remote, expecting the need of force pushing since they have rewritten some history, they might get into trouble.

bin/git-reauthor
tardypad: changed to `--use-config`, in commit 614d6ab (no need to specify git I think)
removed capital letters in short flags, in commit a6058f3 (was a convention of mine to distinguish flags)

bin/git-reauthor
tardypad: done as part of commit 68fb3d5

bin/git-reauthor
tardypad: After the rewrite of the descriptions, I came to agree to it :), in commit 4700320

man/git-reauthor.md
tardypad: I personally prefer my version:
- I don't want to put too many details in that short description. No need to mention a "committer id" there while many people don't know the existence of those 2 different type of identities. Same with mentioning there that it rewrites commits and tags.
- We better use the word rewrite I think, which is used within the Git documentation for filter-branch.

I've put those details you mention in the man page.

man/git-reauthor.md
tardypad: I agree. I did it like that because that's the way I found it done in many commands in the repo.
removed in commit 68fb3d5

man/git-reauthor.md
tardypad: Great advice :) By reviewing this and rewriting the docs, I found that many things were not so clear. I added a new example and rewrote many stuff.
Hopefully it's much better now.
Done in commit 68fb3d5

man/git-reauthor.md
tardypad: Good point! I've added this safeguard in commit 80a50b1

tardypad:
I've just replied to all the previous reviews made and pushed new commits fixing those.
Let me know if you want me to improve anything else.

man/git-reauthor.md
```diff
@@ -7,25 +7,26 @@ git-reauthor(1) -- Rewrite history to change author's identity
...
 
-Rewrite history to change author's identity.
+Lets you replace the author and/or committer identities in commits and tags.
 
-Goes through all existing commits and tags to selectively modify the identities:  
-- modify name and/or email  
-- affect author and/or committer identities  
-- specify the new name/email as parameters or use the ones from the Git config  
-- change all objects or only replace those from a specific identity (defined by its email)
+The command goes through all existing commits and tags in the history to selectively modify the identities present in those objects. All the other informations are preserved.
```
nicolaiskogheim: If it does this in all branches, then: `the history` -> `all branches`,
else if it's just the current branch, then: `the history` -> `the current branch`. (sorry, I don't remember how it worked, but you get the point.

> ~~All the other informations are preserved~~Commits not matching the query are left untouched.

man/git-reauthor.md
```diff
@@ -56,7 +56,7 @@ Change manually the personal email and name of Jack to his work ones
 
 Change automatically the email and name of Jack to the ones defined in the Git config
```
nicolaiskogheim: ~~Change automatically~~Replace

man/git-reauthor.md
```diff
@@ -56,7 +56,7 @@ Change manually the personal email and name of Jack to his work ones
...
 
-    $ git reauthor --old-email jack@perso.me --config
+    $ git reauthor --old-email jack@perso.me --use-config
 
 Fix only the email of Jack (keep the name already used)
```
nicolaiskogheim: ~~Fix~~Replace

man/git-reauthor.md
nicolaiskogheim: Good points. I agree.

nicolaiskogheim:
Fantastic work with the man-page! This might very well be the best one we've got.
I left some nits, but other than that, LGTM.

Commands.md
```diff
@@ -340,6 +341,42 @@ Does the following:
...
+## git reauthor
+
+Rewrite history to change author's identity.
+
+Change manually the personal email and name of Jack to his work ones
+```bash
+$ git reauthor --old-email jack@perso.me --correct-email jack@work.com --correct-name 'Jack Foobar'
+```
+
+Change automatically the email and name of Jack to the ones defined in the Git config
```
nicolaiskogheim: ~~Change automatically~~Replace

Commands.md
```diff
@@ -340,6 +341,42 @@ Does the following:
...
+```bash
+$ git reauthor --old-email jack@perso.me --use-config
+```
+
+Fix only the email of Jack (keep the name already used)
```
nicolaiskogheim: ~~Fix~~Replace

nicolaiskogheim:
Tested on OSX and FreeBsd, both with bash4.3. Works as expected.

man/git-reauthor.md
tardypad: > Commits not matching the query are left untouched.

Actually it's a bit different than this. A commit which is not directly concerned by a change of identity itself will still have its hash changed if its parent commit was modified.

I've changed the description to better explained what I meant there: "All the other informations such as dates, messages,.. are preserved"

tardypad:
Thanks :-)
I've just amended one commit to include the last tiny wording fixes

nicolaiskogheim:
We have had some additions to `Commands.md` on the same places that this PR adds to.
You will get a conflict in that file when you rebase onto `master`, but it shouldn't be to difficult to resolve. Ask if you need any help :)

tardypad:
Fixed

nicolaiskogheim:
Merged.

Thank you. Great work @tardypad! ✨

PR #552, opened 2016-07-25
fork: Fork 'origin' remote repo if repo is not given

npcode:


hemanth:
Thanks!

nicolaiskogheim:
Thanks. It seems only natural that we try and fork the repo we're in as a default, good catch.

There is some logic that need to change, though. The rest of the script assumes that it should clone the repo, cd into it and add the argument (here $url) as a remote called 'upstream'. See (this part)[https://github.com/npcode/git-extras/blob/76663c26db0a9c07f109d636a6f39af4cace0eaa/bin/git-fork#L38-L43]

What we want to do in this case is just rename `origin` to `upstream`, and add the fork as `origin`.

The manual pages also have to be updated to reflect that we will have a default value, and that `git fork` will do as I just described when ran without arguments.

nicolaiskogheim:
A little quick on the trigger there, @hemanth ;)

@npcode We'd appreciate a new PR with the changes I suggested. Please comment if anything was unclear, or if someone has other suggestions.

npcode:
@nicolaiskogheim I've sent the pull request. Your idea is just what I wanted!

PR #553, opened 2016-07-25
fork: Rename 'origin' to 'upstream' after forking

npcode:
Improve #552.


hemanth:
@nicolaiskogheim This time the trigger was bit slower ;)

bin/git-fork
```diff
@@ -34,10 +34,15 @@ fi
 curl -X POST -u "$user" "https://api.github.com/repos/$owner/$project/forks"
 [ $? = 0 ] || abort "fork failed"
 
-# clone forked repo into current dir
-git clone "https://github.com/$user/$project" "$project"
+if [ "$origin" = true ]; then
+    git remote rename origin upstream
+    git remote add origin "https://github.com/$user/$project"
```
nicolaiskogheim: We need to do a `git fetch --all` here to update the references and make the new `origin` "visible" so to speak.

man/git-fork.md
```diff
@@ -7,12 +7,18 @@ git-fork(1) -- Fork a repo on github
...
 
   1. forks the repo on github
   2. clones the repo into the current dir
   3. adds the original repo as a remote called upstream
 
+  If a url is not given and the current dir is a github repo, fork the repo.
+
+  1. forks the current repo
+  3. rename the 'origin' remote repo to 'upstream'
+  3. adds the forked repo as a remote called origin
```
nicolaiskogheim: Either single quote every occurence of `upstream` and `origin`, or none. I can't decide which I think is best, but I'm positive that they should all be written the same.

nicolaiskogheim:
@hemanth Hehe, yes. But I didn't have time to review.

@npcode Much better. Please see comments in the code. 
Also, the file `man/git-fork.html` dissapeared. It should have been generated by the man script. Any idea what happened?

man/git-fork.md
npcode: Now I think we should quote refs by backtick rather than single quote as [git does for man pages](https://github.com/git/git/commit/1249d8ad1c1b2d6fb42aabcaada2c0792d52232d). I'll send a pr to fix it.

bin/git-fork
npcode: Isn't `git fetch origin` enough?

I have several remote refs and some of them are accessible only in my office. I'm afraid that `git fetch --all` causes some kind of network error in my home.

npcode:
@nicolaiskogheim I ran `make docs` and the document was generated. I have no idea about why it disappeared in my branch. I've sent a new pr #554 to fix it and update this feature.

PR #554, opened 2016-07-26
Improve forking `origin` repo

npcode:
Update #552 and #553 

* Add git-fork.html removed by accident
* Quote refs by backtick rather than single quote
* Fetch origin after forking


nicolaiskogheim:
You need to run `make docs` again to generate the new versions of both the html and man pages 😁  To be sure you can just `rm man/git-fork.1 man/git-fork.html` before you `make docs`.

npcode:
Thanks! I've regenerated them.

PR #555, opened 2016-07-27
Add Contributing guideline

spacewander:
@hemanth @qw3rtman @nicolaiskogheim 
Let's have a look at this fresh guideline!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,16 @@
+## Your new git-extra command should support...
+
+* Should support Mac and Linux(You may need to browse their man page)
+* Should support Bash 3 or newer(If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
+* Should support Git 2.1+
+
+## To submit a new command, you should...
+
+Let's assume your new command is named `foo`.
+
+1. Write a bash script under `./bin` called `git-foo`.
+2. Read `./man/Readme.md` and write documentation for `git-foo`.
+3. Don't forget to introduce it in `Commands.md`.
+4. Update `./etc/git-extras-completion.zsh`. Just follow existing code.
+5. (Optional) Update `./etc/bash_completion.sh`.
+6. Run `./check_integrity.sh foo` to check if all is done.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,11 @@
 ## Your new git-extra command should support
 
-* OSX, Ubuntu, FreeBsd(You may need to browse their man page)
-* Bash 3.2+ (If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
+* OSX, Linux, BSD (You may need to browse their man page)<sup>*</sup>
+* Bash 3.2+ (If you aren't sure, see [the Bash changelog](http://tldp.org/LDP/abs/html/bash2.html))
 * Git 2.1+
+
+<sup>*</sup>If you aren't able to test your new command on a platform, 
+make that clear in your PR and someone else may be able to test it on their system.
 
 ## To submit a new command, you should
 
@@ -13,6 +16,6 @@ Let's assume your new command is named `foo`.
 3. Don't forget to introduce it in `Commands.md`.
 4. Update `./etc/git-extras-completion.zsh`. Just follow existing code.
 5. (Optional) Update `./etc/bash_completion.sh`.
-6. Run `./check_integrity.sh foo` to check if all is done.
+6. Run `./check_integrity.sh foo` to check if all done.
 
 You are welcome to open up an issue to discuss new commands or features before opening a pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,10 +1,10 @@
-## Your new git-extra command should support...
+## Your new git-extra command should support
 
-* Should support Mac and Linux(You may need to browse their man page)
-* Should support Bash 3 or newer(If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
-* Should support Git 2.1+
+* OSX, Ubuntu, FreeBsd(You may need to browse their man page)
+* Bash 3.2+ (If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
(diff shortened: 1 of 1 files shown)


nicolaiskogheim:
Could you split this PR in 3? :)
One for contributing guidelines, one for `check_integrity.sh`, and one for the rest.

spacewander:
Done

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+## Your new git-extra command should support...
```
nicolaiskogheim: `## Your new git-extra command should support`

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
...
+
+* Should support Mac and Linux(You may need to browse their man page)
+* Should support Bash 3 or newer(If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
+* Should support Git 2.1+
```
nicolaiskogheim: Remove `Should support` from all the items.

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+## Your new git-extra command should support...
+
+* Should support Mac and Linux(You may need to browse their man page)
```
nicolaiskogheim: I think we need to be more precise on this one.  We have users on i.e. *BSD and Solaris (i think), and one might forget that those may differ from your conventional Arch or Ubuntu in terms of the shell and maybe even available commands.

I don't actually know how to formulate this in a concise way, but our commands should run on all Ubuntu, Debian, Arch Linux, Linux Mint, Fedora/RedHat, and Bsd's and Solaris. The first few can probably go as Unix-like systems (along with OSX), but although *Bsd is Unix-like, technically, I think it deserves a mention because some of them have more restrictive shells.

I have rewritten the line one too many times now.. 
How about `OSX, Ubuntu, FreeBsd`.

So, if it runs on OSX, no one used Linux-specific stuff like reading from `/proc`, if it runs on Ubuntu it probably runs on most other Linux distros, and if it runs on FreeBsd there isn't any features considered 'unsafe' by FreeBsd and hopefully it runs on OpenBsd and the likes.

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
...
+* Should support Bash 3 or newer(If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
+* Should support Git 2.1+
+
+## To submit a new command, you should...
```
nicolaiskogheim: `## To submit a new command, you should`

nicolaiskogheim:
Thanks. Great work! I have added my two cents.

nicolaiskogheim:
At the bottom we could add something like:

```txt
You are welcome to open up an issue to discuss new commands or features before opening a pull request.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,16 @@
+## Your new git-extra command should support...
+
+* Should support Mac and Linux(You may need to browse their man page)
+* Should support Bash 3 or newer(If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
```
nicolaiskogheim: I don't actually think we will have much difficulties supporting Bash 3.0 (at least that's the impression I got from scanning over the changelogs), but it may be difficult for people to test it. I tried building Bash 3.0 on OSX, and I was unable to, even with all the patches.

Now, the default shell on my OSX El Capitan is 3.2 so that one shouldn't be hard to get hold of.

Should we say `Bash >= 3.2`?

spacewander:
What about the latest version of CONTRIBUTING.md?

CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
+## Your new git-extra command should support
+
+* OSX, Ubuntu, FreeBsd(You may need to browse their man page)
```
qw3rtman: Perhaps `OS X, Ubuntu, and FreeBSD (you may need to check the documentation for compatibility issues)`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
...
+* Bash 3.2+ (If you are not sure, [Bash versions](http://tldp.org/LDP/abs/html/bash2.html))
```
qw3rtman: Perhaps "Bash 3.2+ (if you are not sure, see [the Bash documentation](http://tldp.org/LDP/abs/html/bash2.html))"?

qw3rtman:
LGTM! I just mainly made a couple formatting changes along with one suggestion regarding what to do if you are unable to test the command on a particular OS.

On the topic of which OS's to support, should we explicitly include CentOS/RHEL as a requirement? Or would that be covered by Ubuntu already?

spacewander:
@qw3rtman 
I guess more people use ArchLinux as their desktop distribution than CentOS/RHEL...
Let's use `Linux` to simply represent all the distributions.

qw3rtman:
@spacewander

I'm good with that. So let's replace `Ubuntu` with `Linux` and keep `OS X` and `FreeBSD` (or just `BSD`?) (since `BSD != Linux`)

@nicolaiskogheim, what do you think?

CONTRIBUTING.md
nicolaiskogheim: I'm in favor of both of those suggestions.

CONTRIBUTING.md
nicolaiskogheim: Actually, that link goes to an unofficial changelog between the versions of Bash. There may be inaccuracies there.

The documentation is part of Advanced Bash Scripting Guide, which was written by a guy who wanted to learn to script with bash, and so it's pointed out regularly around the web that it's not the best resource for everything. It is a very good guide, and it has received a lot of feedback and suggestions from people like Stéphane Chazelas, who seems to know everything there is to know about bash and related topics.

Anyways, I think I'd prefer it'd said `Bash 3.2+ (If you aren't sure, see [the bash changelog])`.

nicolaiskogheim:
OSX, Linux, BSD is fine by me 👍

nicolaiskogheim:
Let's just hold this until #557 is in.

hemanth:
#557 is in, this is good to merge.

PR #77, opened 2012-02-11
Adding git-pull-all

hughsk:
Track working directories once, and then use `git pull-all` to pull each repository
in one go later. May be useful? Commands include `add`, `rm`, `ls` and `pull`.

PR #556, opened 2016-07-27
[WIP] Add git-pull-all command

nicolaiskogheim:
> Track working directories once, and then use `git pull-all` to pull each repository in one go later.

This is just #77 rebased onto master. The only edits I did was remove trailing whitespace.

I don't think I would use this command, but a couple of guys showed interest in the comments.
**If anyone has thoughts on whether `git pull-all` should be included or not, please share.**

I haven't reviewed or tested the code, hence the WIP label, but I will after we've decided to take it in. If we do, that is.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -35,6 +35,7 @@
  - [`git obliterate`](#git-obliterate)
  - [`git pr`](#git-pr)
  - [`git psykorebase`](#git-psykorebase)
+ - [`git pull-all`](#git-pull-all)
  - [`git release`](#git-release)
  - [`git repl`](#git-repl)
  - [`git reset-file`](#git-reset-file)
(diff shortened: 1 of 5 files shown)


qw3rtman:
Just to clarify, does this pull all branches from remote?

@hughsk

nicolaiskogheim:
> does this pull all branches from remote?

No.

There are two cases:
1. if `$# -eq 0` then `git pull` in each repository
2. else `git pull "$@"`, which means you can give arguments, but passing in branch names wouldn't make sense, except when you want to only pull the `master` branch in every repo.

qw3rtman:
My bad, I should have looked at the man page in this PR.

This looks like a useful feature to me (especially for those who keep all their repositories in a central directory).

PR #560, opened 2016-07-29
Fix incorrect comment in git-fork

lukechilds:


qw3rtman:
Good find! Thanks! :smil

Issue #540, opened 2016-06-06
Any way to persist Github username and password to avoid prompt for git fork?


Pysis868:
I don't want to enter it for every repository I want to fork.
This seems to be the only command that contacts Github.com[[1]](https://github.com/tj/git-extras/blob/master/Commands.md).
Would or could this read a .gitconfig file?  I might actually like this separate, as I have both a private and public use for git tools on the same computer.

spacewander:
Maybe use some environment variable like `GITHUB_USER_NAME` or `GITHUB_PASSWORD`, which is simpier.

nicolaiskogheim:
@Pysis868 Have you tried [github's hub](https://github.com/github/hub)? I've only tested it just now but it looks like it remembers credentials.

Hub is much like git-extras, and they actually have some of the same helper commands as we do, only you write `hub <command>`, so you can have both installed without getting conflicting commands.

lukechilds:
@nicolaiskogheim Couldn't you bypass the login prompt by setting the remotes using ssh rather than https? Then the ssh key will be used for auth in the background without any user interaction needed.

Happy to submit a PR for this, or is there a reason it uses https instead of ssh?

qw3rtman:
@lukechilds, AFAIK the SSH key can't be used to fork a repository on the remote as this would require authentication via the GitHub API.

One option that is possible would be creating a GiyHub application and authenticating via the app (while storing the OAuth token); however, I'm not sure this is the right way to go about doing this in terms of security and usability.

lukechilds:
@qw3rtman I think you may have misunderstood, I'm only proposing to use ssh for the remotes so you don't get prompted for each push, not the actual API request to do the fork.

See example in the referenced PR.

qw3rtman:
My bad! What you're saying makes sense. 👍

nicolaiskogheim:
> or is there a reason it uses https instead of ssh?
@lukechilds I answered in your PR :)

nicolaiskogheim:
Okay, so we have solved something with #563, but the initial fork still requires username and password/token, which means this issue still isn't solved.

I would still strongly suggest people who want to deal with github to use github/hub, and I still would prefer that we didn't deal with github api's. But, if there is a strong interest in being able to use `git fork` without having to type passwords, we will reconsider it. Pull requests are welcomed. Just comment here if you want to contribute and we will nudge you in the right direction.

rachmadaniHaryono:
[...]

PR #563, opened 2016-08-02
Set fork remotes using ssh instead of https

lukechilds:
Proposed in https://github.com/tj/git-extras/issues/540

This allows you to push changes to your fork without being prompted for your GitHub credentials every time.

Previous behaviour:

```
$ git fork tj/commander.js
Enter your github username
lukechilds
Enter host password for user 'lukechilds':
{
  ...
}
Cloning into 'commander.js'...
remote: Counting objects: 1791, done.
remote: Total 1791 (delta 0), reused 0 (delta 0), pack-reused 1791
Receiving objects: 100% (1791/1791), 381.24 KiB | 677.00 KiB/s, done.
Resolving deltas: 100% (977/977), done.
Checking connectivity... done.
From https://github.com/tj/commander.js
 * [new branch]      gh-pages   -> upstream/gh-pages
 * [new branch]      master     -> upstream/master

$ cd commander.js && git push
Username for 'https://github.com':
```

New behaviour:

```
$ git fork tj/commander.js
Enter your github username
lukechilds
Enter host password for user 'lukechilds':
{
  ...
}
Cloning into 'commander.js'...
remote: Counting objects: 1791, done.
remote: Total 1791 (delta 0), reused 0 (delta 0), pack-reused 1791
Receiving objects: 100% (1791/1791), 381.24 KiB | 309.00 KiB/s, done.
Resolving deltas: 100% (977/977), done.
Checking connectivity... done.
From github.com:tj/commander.js
 * [new branch]      gh-pages   -> upstream/gh-pages
 * [new branch]      master     -> upstream/master

$ cd commander.js && git push
Everything up-to-date
```

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -196,7 +196,7 @@ project  : git-extras
 
 ## git line-summary
 
-  WARNING: git line-summary has been replaced by git summary --line and will be removed in a future release. 
+  WARNING: git line-summary has been replaced by git summary --line and will be removed in a future release.
 
 ## git effort
 
@@ -215,7 +215,7 @@ $ git effort --above 5
 ```
 
   If you wish to see only the commits in the last month you may use `--since` (it supports the same syntax like `git log --since`):
-  
+
 ```
  $ git effort --since='last month'
 ```
@@ -319,7 +319,7 @@ Does the following:
 - forks the repo (prompts for github username and pass)
 - clones the repo into the current directory
 - adds the original repo as a remote so can track upstream changes
-- all remotes refs use git over ssh
+- all remotes refs use git over ssh if configured, otherwise https will be used
 
 
 ```Shell
@@ -427,7 +427,7 @@ To show just the global or just the local file's contents, you can use the follo
 
 * `-g` or `--global` to show just the global file
 * `-l` or `--local` to show just the local file
- 
+
 ```bash
 $ git ignore -g
 Global gitignore: /Users/foo/.gitignore_global
@@ -462,7 +462,7 @@ $ git ignore-io vim
     .netrwhist
     *~
 ```
-  
+
 To export it to `.gitignore` file you can use the following options:  
 
 * `-a` or `--append` to append the result to `.gitignore`
@@ -476,7 +476,7 @@ For efficiency, `git ignore-io` store all available types at `~/.gi_list`.
 To list all the available types:
 
 * `-l` or `-L` : These two options will show the list in different format. Just try it.
-  
+
 You can also search type from the list by:
 
 * `-s <word>` or `--search <word>`
@@ -804,7 +804,7 @@ Set up a git repository (if one doesn't exist), add all files, and make an initi
 
 A convenient way to copy files from the current working tree to the working directory of a remote repository. If a `<commits>...` is provided, only files that has changed within the commit range will be copied.
 
-Internally this script uses `rsync` and not `scp` as the name suggests. 
+Internally this script uses `rsync` and not `scp` as the name suggests.
 
 `git-rscp` - The reverse of `git-scp`. Copies specific files from the working directory of a remote repository to the current working directory.
 
@@ -1078,7 +1078,7 @@ Create pull request via commandline.
 
 ## git rebase-patch
 
-Given  you have a patch that doesn´t apply to the current HEAD, but you know it applied to some commit in the past, 
+Given  you have a patch that doesn´t apply to the current HEAD, but you know it applied to some commit in the past,
 `git rebase-patch` will help you find that commit and do a rebase.
 
 For example,
@@ -1108,4 +1108,3 @@ Sync local branch with origin/master
 ```bash
 $ git sync origin master
 ```
-
--- a/man/git-fork.md
+++ b/man/git-fork.md
@@ -16,8 +16,10 @@ git-fork(1) -- Fork a repo on github
   If a url is not given and the current dir is a github repo, fork the repo.
 
   1. forks the current repo
-  3. rename the `origin` remote repo to `upstream`
+  2. rename the `origin` remote repo to `upstream`
   3. adds the forked repo as a remote called `origin`
+
+  Remotes will use ssh if you have it configured with GitHub, if not, https will be used.
 
 ## EXAMPLE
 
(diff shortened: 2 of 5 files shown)


qw3rtman:
I'm good with this.

@nicolaiskogheim, @hemanth, @spacewander?

nicolaiskogheim:
This will make it less of a hassle for people who have ssh set up, but it will be a problem for those who haven't.

So this is actually a breakage for http users. The forking works, but cloning doesn't.

An alternative would be to use ssh if the argument to `git fork` is an ssh-url, and default to https. And we could have an option `--ssh/-s` to force ssh.


Either way, I'm not positive to this particular change.

lukechilds:
@nicolaiskogheim Fair enough, is this something you want me to work on then or are you not bothered?

I assume most people would use the shorthand `git fork user/repo` so checking for ssh/https in the argument probably wouldn't solve much.

The flag would work but seems kind of annoying to have to type every time. What do you reckon?

nicolaiskogheim:
I just got an idea.

We could run something like `ssh -T git@github.com`, and if it works it's safe to assume the ssh protocol.

Be sure to test it in an environment where ssh access isn't configured. I tend to pull out Vagrant for stuff like that.

If the user doesn't use ssh with github, we wouldn't want to add github to `.ssh/known_hosts`, and we don't want the user to answer any prompts from ssh.

You might get something like

```sh
The authenticity of host 'github.com (192.30.252.1)' can't be established.
RSA key fingerprint is 16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)?
```

from ssh, but to avoid that you can do (and redirect all output while we're at it):

```sh
yes no | ssh -T git@github.com >/dev/null 2>&1
```

There probably is a flag you can pass to ssh instead of the `yes no | ...` stuff.


> is this something you want me to work on

Yes :)

lukechilds:
Nice, I like it.

I'll have a play tomorrow.

lukechilds:
@qw3rtman @nicolaiskogheim Ok, I've updated this to check if we can login via ssh, if we can, ssh remotes are used, if not it falls back to https.

I've added the exit command to ssh and check for both 0 and 1 exit codes in bash so that if (for whatever reason) GitHub ever allow interactive ssh sessions in the future this will still work and won't hang the terminal.

The only caveat is that if the user has ssh set up for GitHub but they're on a new machine and don't have github.com in `known_hosts` this will treat it as a fail. This isn't a big deal as it's unlikely to happen and if it did it'll just revert to https so everything will still work ok.

I've done this on purpose, we could actually bypass `known_hosts` and not save the fingerprint by adding the args `-o StrictHostKeyChecking=no -o "UserKnownHostsFile /dev/null"`, however I don't think we should do this as we could potentially open users up to a MITM attack.

Also updated docs to reflect the new behaviour.

lukechilds:
Just a thought, I could remove `-o BatchMode=yes` from ssh and then `git fork` will prompt them to add github.com to `known_hosts` if it isn't already in there. All other ssh output will be hidden.

Commands.md
```diff
@@ -196,7 +196,7 @@ project  : git-extras
 
 ## git line-summary
 
-  WARNING: git line-summary has been replaced by git summary --line and will be removed in a future release. 
+  WARNING: git line-summary has been replaced by git summary --line and will be removed in a future release.
```
lukechilds: Oops, looks like atom removed a load of unnecessary white space when I saved. Happy to split this out into a separate commit (or just remove it) if you want.

nicolaiskogheim:
Looks great. I gave it a test run with and without ssh configured, and it worked as I expected.
When the man page is updated to reflect the change (especially the example there), this is good to go.

lukechilds:
Sweet, my lunch break is over now but I'll get that updated when I get home 👍

Commands.md
nicolaiskogheim: This is fine.

nicolaiskogheim:
> Just a thought, I could remove -o BatchMode=yes from ssh and then git fork will prompt them to add github.com to known_hosts if it isn't already in there.

My guess is that if github.com isn't in known_hosts, then they probably haven't configured ssh with github.com either, and a silent fail is appropriate.

lukechilds:
This should be all sorted now 🎉

man/git-fork.md
```diff
@@ -16,9 +16,11 @@ git-fork(1) -- Fork a repo on github
   If a url is not given and the current dir is a github repo, fork the repo.
 
   1. forks the current repo
-  3. rename the `origin` remote repo to `upstream`
+  2. rename the `origin` remote repo to `upstream`
   3. adds the forked repo as a remote called `origin`
 
+  Remotes will use ssh if you have it configured with GitHub, if not, https will be used.
```
qw3rtman: Maybe change this to:

The `ssh` remote will be used if you have used an `ssh` remote with GitHub in the past. Otherwise, the `https` remote will be used.

qw3rtman:
I think it looks great except for the one stylistic suggestion I made. @nicolaiskogheim, thoughts?

Also, could you regenerate the documentation pages ([see this](https://github.com/tj/git-extras/blob/master/man/Readme.md))?

lukechilds:
I thought I did regenerate the docs in https://github.com/tj/git-extras/pull/563/commits/f22d7bcc656fb53c70d6d7c2a024af3b0a026d75, or is something wrong?

It's late here and I'm tucked up in bed, I'll update the man page tomorrow morning 👍

qw3rtman:
No worries. It seems like `git-fork.html` has been deleted so you may need to regenerate that.

nicolaiskogheim:
The html file is missing :)

Otherwise, LGTM.

man/git-fork.md
nicolaiskogheim: >  if you have used an ssh remote with GitHub in the past
This isn't necessarily accurate.

I'm fine with the way it is.

nicolaiskogheim:
I wrote that last comment on an old version this page. Apart from the missing `git-fork.html` this looks good to me.

lukechilds:
Had to delete `git-fork.html` for `ronn` to pick it up ¯\_(ツ)_/¯

Should be sorted now ✨

nicolaiskogheim:
Thanks, @lukechilds!

PR #564, opened 2016-08-03
Remove a non-contributor from AUTHORS

sorbits:
Since there is hesitation about PR #544 this is a quick fix to at least remove my email from the master (google the email in question and you’ll find this repository is the only hit).


hemanth:
ok-go-never-OT1gQvc2HjkT6

Issue #532, opened 2016-04-28
`git rscp staging` doesn't sync remote unstaged files


tribou:
I was unable to get `git rscp staging` to sync remote unstaged files to my local repo.  Here's the output I get:

```
$ git rscp staging
receiving file list ... done

sent 8 bytes  received 13 bytes  14.00 bytes/sec
total size is 0  speedup is 0.00
```

I verified the remote is setup correctly (`git fetch` works).  As a workaround, I was able to just sync the entire remote repo with this:

```bash
$ git rscp staging .
```

Version info:
```
$ git-extras -v
4.1.0
```

spacewander:
Could [this version](https://github.com/spacewander/git-extras/tree/add_rscp_default_arg) fix the bug?

hemanth:
echo @tribou

tribou:
I copied the `git-scp` file from that branch to `/usr/local/bin`, but I'm not sure if it's being used (I installed git-extras with brew).  I'm still getting the same results.

spacewander:
@tribou What is the result of `which git-scp`?

tribou:
```
$ which git-scp
/usr/local/bin/git-scp
```

tribou:
`$ cat $(which git-scp) | pbcopy`

output:
```bash
#!/usr/bin/env bash

COLOR_RED()   { test -t 1 && echo -n "$(tput setaf 1)"; }
COLOR_GREEN() { test -t 1 && echo -n "$(tput setaf 2)"; }
COLOR_YELLOW(){ test -t 1 && echo -n "$(tput setaf 3)"; }
COLOR_BLUE()  { test -t 1 && echo -n "$(tput setaf 4)"; }
COLOR_RESET() { test -t 1 && echo -n "$(tput sgr 0)"; }

function _test_git_scp()
{
	command -v rsync    > /dev/null || _error requires rsync
	command -v git      > /dev/null || _error requires git
	command -v ssh      > /dev/null || _error requires ssh
	command -v php      > /dev/null || _info  optional php
	command -v dos2unix > /dev/null || _info  optional dos2unix
}

function set_remote()
{
	remote=$1
	if [ $(git remote | grep -c -- ^$remote$) -eq 0 ]
	then
		COLOR_RED
		echo "Remote $remote does not exist in your git config"
		COLOR_RESET
		exit 1
	fi
}

# Check file for PHP syntax errors
# takes a list of filenames
function php_lint()
{
	local error_count=()
	for i
	do
[...]

Issue #534, opened 2016-05-02
Time for a new release.


hemanth:
//cc @spacewander @qw3rtman

spacewander:
Maybe after https://github.com/tj/git-extras/issues/532 is closed?

hemanth:
Makes sense.

spacewander:
Close the issue because we have made a new release(4.2.0) in Oct 8th.

PR #565, opened 2016-08-06
git-missing: do proper argument parsing. Fix #562

### LINKED ISSUE #562: `git missing` using `--no-merges` doesn't seem to work as expected
Hello,

I'm trying to use the `git missing` with the `--no-merges` param but it doesn't work as (at least I)  expected: it returns the same list as calling it without the `--no-merges` param, listing all the commits including the merge ones.

I'm using `git missing branch1 branch2 --no-merges`.

If I try to run  `git missing branch2 --no-merges` or `git missing --no-merges branch2` it throws, respectively:

`fatal: unrecognized argument: --no-merges...branch2`

````
fatal: ambiguous argument 'branch2...--no-merges': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I don't know if it's a bug or I've just misunderstood the command, however in the man page it says that the same parameters as `git log` can be used.

Thanks! 😄

nicolaiskogheim:
Ref #562

This is a rewrite of `git-missing` to make it behave like the docs says it will.

I have tested by running these commands, which should be enough:
```sh
$ git missing feature
$ git missing master feature
$ git missing --no-merges feature
$ git missing --no-merges master feature
$ git missing --no-merges master feature --no-merges
```

That last one may seem stupid, but argument parsing, and especially passing arguments along to other commands in bash can be very tricky, so that one is important. `git log` doesn't care about duplicate arguments, so that's alright.

Only tested on OSX, with bash 4.3.46(1)-release.
I don't think there is anything that would make this crash on other systems, but I will test it on FreeBsd and Arch tomorrow.

___________
Possible future enhancements:

`git-missing` will get confused with a call looking like this: `git missing --option value firstbranch secondbranch`.  I don't know if `git log` accepts any argument on the form `--arg value`. I scanned `man git-log` and it doesn't look like it, but `git log` understands hundreds of different arguments it looks like, so I may be wrong.

### DIFF
--- a/bin/git-missing
+++ b/bin/git-missing
@@ -1,21 +1,41 @@
 #!/usr/bin/env bash
 
-# grab first two non-option arguments
-for arg in $*; do
-    if [ -n "${arg}" ]; then
-        test -z "$firstbranch" && firstbranch="${arg}" && continue
-        test -z "$secondbranch" && secondbranch="${arg}" && continue
-    else
-        # add anything else to a passthrough
-        passthrough="$passthrough $arg"
-    fi
-done
+usage() {
+    echo 1>&2 "usage: git missing [<first branch>] <second branch> [<git log options>]"
+}
 
-test -z "$firstbranch" && echo "at least one branch required" && exit 1
-
-if test -z "$secondbranch"; then
-    secondbranch=$firstbranch
-    firstbranch=
+if [ "${#}" -lt 1 ]
+then
+    usage
+    exit 1
 fi
 
-git log $passthrough $firstbranch...$secondbranch --format="%m %h %s" --left-right
+declare -a git_log_args=()
+declare -a branches=()
+for arg in "$@" ; do
+
+    case "$arg" in
+        --*)
+            git_log_args+=( "$arg" )
+            ;;
+        *)
+            branches+=( "$arg" )
+            ;;
+    esac
+done
+
+firstbranch=
+secondbranch=
+if [ ${#branches[@]} -eq 2 ]
+then
+    firstbranch="${branches[0]}"
+    secondbranch="${branches[1]}"
+elif [ ${#branches[@]} -eq 1 ]
+then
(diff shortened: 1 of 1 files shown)


nuragic:
Thanks so much for this one!

qw3rtman:
Tested and working on `FreeBSD 10.3` with `GNU bash, version 4.3.46(1)-release (amd64-portbld-freebsd10.1)` and `git version 2.9.0`. I also tested `git missing master feature --no-merges` and `git missing feature --no-merges`.

nicolaiskogheim:
Thanks, @qw3rtman.

------------
I found something that will break this, and that is the `-L <start>,<end>:<file>, -L :<funcname>:<file>` argument. I'll consider redoing the argument loop to tackle that later.

nuragic:
@nicolaiskogheim hi! now that this change is on `master`, it will be automatically available in the latest release... ?

nicolaiskogheim:
No, sorry. We will have to do a release first. When that will be I don't know.
@hemanth When do you think we'll be able to cut a new release?

hemanth:
After #534 ?

nicolaiskogheim:
Thanks, forgot about that one.

@nuragic We are waiting to resolve #532, which is partly blocked while waiting for a reply by the author of that issue. But after that, it shouldn't be too long.

nuragic:
Hi @nicolaiskogheim, 

Sorry in advance if I sound pedantic but, looking at that PR (#532) there are no news since June..

Wondering if you guys can just do a patch release...? (`v4.1.1`) 😉

nicolaiskogheim:
Yeah, you're right, it has been a long time. We do have a lot of stuff that should be released.

@spacewander @hemanth @qw3rtman

qw3rtman:
@nicolaiskogheim, made a new issue for further discussion: #579.

PR #568, opened 2016-08-11
Add built in 'exit' command to git-repl

wcmonty:
This just adds the ability to use `exit` to quit git-repl, in addition to the built-in `quit`.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -248,6 +248,8 @@ bin/git-release
 git> quit
 ```
 
+You can use `exit` instead of `quit`.
+
 ## git commits-since
 
 List commits since `date` (defaults to "last week"):
--- a/bin/git-repl
+++ b/bin/git-repl
@@ -28,7 +28,7 @@ while true; do
   case $cmd in
     ls) cmd=ls-files;;
     "") continue;;
-    quit) break;;
+    quit|exit) break;;
   esac
 
   if [[ $cmd == !*  ]]; then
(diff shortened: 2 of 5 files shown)


qw3rtman:
I'm good with this idea! For the actual code, instead of breaking `exit` and `quit` into two separate `case`s, you can combine them into one case like this:

```sh
quit|exit) break;;
```

Additionally, it would be nice to add this to the documentation [here](https://github.com/tj/git-extras/blob/master/Commands.md#git-repl) and update and regenerate the man page. 😄

wcmonty:
@qw3rtman I updated the script to combine the cases in 	554d779.  The documentation has been updated in d6f505f.

nicolaiskogheim:
The man page is not organised well now.

`Git read-eval-print-loop` belongs under `Description`, and the rest should go under `Examples`. The `Options` heading can be removed.

We should perhaps have a better description, but you don't need to do that now if you don't feel like it. The other changes is overdue so I would like that handled in this PR.

nicolaiskogheim:
Too late, haha.

qw3rtman:
@nicolaiskogheim Merged before I saw this. You can revert if you see fit.

nicolaiskogheim:
Nah, I'll open a new one :) Should be quick to do.

nicolaiskogheim:
Done https://github.com/tj/git-extras/pull/569

Issue #535, opened 2016-05-12
Git rename local and remote branch.


Dershowitz011:
If you could add a feature in the Git-Extras which can rename a branch properly, it'd be great! Thanks!

hozefaj:
Will like to take a crack at this~

nicolaiskogheim:
@Dershowitz011 Is #576 what you had in mind?

PR #576, opened 2016-09-11
add rename-branch command

hozefaj:
solve for #535

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -40,6 +40,7 @@
  - [`git rebase-patch`](#git-rebase-patch)
  - [`git reauthor`](#git-reauthor)
  - [`git release`](#git-release)
+ - [`git rename-branch`](#git-rename-branch)
  - [`git rename-tag`](#git-rename-tag)
  - [`git repl`](#git-repl)
  - [`git reset-file`](#git-reset-file)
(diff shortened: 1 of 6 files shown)


Commands.md
```diff
@@ -361,6 +362,17 @@ Does the following:
   - Executes _.git/hooks/post-release.sh_ (if present)
 
 
+## git rename-branch
+
+Rename a branch (locally and remotely).
```
spacewander: In my opinion, it is worth to clarify that this command doesn't really rename a remote branch. Seems it rename the upstream of local branch?

bin/git-rename-branch
```diff
@@ -0,0 +1,13 @@
...
+test -z $1 && echo "new branch name required." 1>&2 && exit 1
+
+if [ -z $2 ]
+  then
+    $2=${git symbolic-ref --short -q HEAD}
+fi
+
+git branch -m $2 $1
+git push origin :$2
+git push --set-upstream origin $1
```
spacewander: Use `branch -u/--set-upstream-to` if you just want to update the upstream.

Commands.md
hemanth: 🌵

spacewander:
@hozefaj 
I wonder why we need to push a branch with a comand named `rename-branch`...
Could you tell us the reason?

hozefaj:
@spacewander Idea being to keep remote and local in sync.

bin/git-rename-branch
hozefaj: Not just updating, but pushing the renamed branch. `--set-upstream` allows local branch to track the remote branch.

Commands.md
hozefaj: @spacewander Technically after renaming the local, the branch is pushed to remote. So maybe the documentation can be a bit more clear on that front. Will make the update.

- [ ] update docs clarifying `rename` functionality

PR #577, opened 2016-09-12
Update git-delete-branch

### LINKED ISSUE #514: git-delete-branch shouldn't assume origin
A lot of these extra commands don't have any help text and the only way to figure out what they're doing is by looking at the source. If I run `git delete-branch` and want it to delete on the remote, it's because the local branch is set up to track it or I've manually specified it. The behavior of assuming `origin` if there is no tracking branch seems like an unintuitive default. I would expect it to give an error because the local branch isn't set up to track a remote branch, or because the remote isn't specified. 

I have a similar complaint about `create-branch` which seems to assume `origin` by default.

I'll propose a fix for `delete-branch` at least:
- Add a new optional argument which is the remote to delete from. This could optionally also have syntax to specify a remote branch name in case it doesn't have the same name. 
- Error if delete-branch is called without specifying a remote and no upstream remote is configured.
- Add some help text.

If it's preferred that `delete-branch` also work for deleting local branches only, then a return value / warning can be setup to indicate that possibly a remote was meant to be deleted but one couldn't safely be found.

Thoughts?

bturrubiates:
Instead of assuming 'origin' as the remote, default to using the remote
tracking configuration. This can be overridden by using the `-r` option to
provide a remote name and a remote branch name. This also removes the ability
to delete multiple branches at once.

Closes #514.

@hemanth, apologies this took so long. I haven't updated the completions file yet,
need to figure that out.

Thoughts?

### DIFF
--- a/bin/git-delete-branch
+++ b/bin/git-delete-branch
@@ -1,16 +1,94 @@
 #!/usr/bin/env bash
 
-# Assert there is at least one branch provided
-test -z $1 && echo "branch required." 1>&2 && exit 1
+set -u
 
-for branch in "$@"
-do
-  remote=$(git config branch.$branch.remote)
-  test -z $remote && remote="origin"
-  ref=$(git config branch.$branch.merge)
-  test -z $ref && ref="refs/heads/$branch"
+usage()
+{
+    local command="git delete-branch"
+    cat << EOF
+Usage:
+    ${command} [-r <remote[:remote branch name]>] <branch>
+    ${command} -h
 
-  git branch -D $branch
-  git branch -d -r $remote/$branch
-  git push $remote :$ref
+Delete local and remote branches. If the remote switch (-r) is provided, then
+use the given remote and remote branch. Otherwise, the remote branch is found
+by first trying the upstream tracking branch. If there is no tracking branch
+configured, then deletion will not succeed unless provided explicitly. If the
+remote branch and local branch do not share a name, the remote branch name can
+be included with the remote switch.
+EOF
+}
+
+remote=
+remote_branch=
+
+while getopts "r:h" opt; do
+    case "$opt" in
+        r)
+            remote="$(echo ${OPTARG} | cut -d ':' -f 1)"
+            remote_branch="$(echo ${OPTARG} | cut -s -d ':' -f 2)"
+
+            # Remote must not be empty string, or anything of that sort. The
+            # remote branch need not be provided here.
+            if [[ -z "$remote" ]]; then
+                echo "error: Provided empty remote name"
+                usage
+                exit 1
+            fi
+            ;;
+        h)
+            usage
+            exit 0
+            ;;
+        \?)
+            usage
+            exit 1
+            ;;
+    esac
 done
+
+shift $((OPTIND-1))
+
+if [[ $# -ne 1 ]]; then
+    echo "error: Need a single branch as a positional argument"
+    usage
+    exit 1
+fi
+
+branch="$1"
+
+# If the -r option was not passed, then both the remote and the remote branch
+# have to be assumed using the tracking configuration. Need to make sure that
+# these values are set as both can be set/unset and both are necessary in this
+# case.
+if [[ -z "$remote" ]]; then
+    remote="$(git config --get branch.${branch}.remote)"
+    if [[ $? -ne 0 ]]; then
+            echo "error: $branch not tracking a repository"
+            usage
+            exit 1
+    fi
+
+    remote_branch="$(git config --get branch.${branch}.merge)"
+    if [[ $? -ne 0 ]]; then
+            echo "error: $branch not tracking a remote branch"
+            usage
+            exit 1
+    fi
+elif [[ -z "$remote_branch" ]]; then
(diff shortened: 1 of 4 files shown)


hemanth:
LGTM 

@spacewander ^

bin/git-delete-branch
```diff
@@ -1,16 +1,94 @@
...
+}
+
+remote=
+remote_branch=
+
+while getopts "r:h" opt; do
+    case "$opt" in
+        r)
+            remote="$(echo ${OPTARG} | cut -d ':' -f 1)"
+            remote_branch="$(echo ${OPTARG} | cut -s -d ':' -f 2)"
```
spacewander: Better to use parameter expansion instead.

bin/git-delete-branch
```diff
@@ -1,16 +1,94 @@
...
+
+            # Remote must not be empty string, or anything of that sort. The
+            # remote branch need not be provided here.
+            if [[ -z "$remote" ]]; then
+                echo "error: Provided empty remote name"
+                usage
+                exit 1
```
spacewander: It will be great if you can extract all `usage... exit ...` into a function.

bin/git-delete-branch
```diff
@@ -1,16 +1,94 @@
...
+    fi
+elif [[ -z "$remote_branch" ]]; then
+    # In the event that a remote is specified but no remote branch is
+    # specified, then assume the branch name is the same as the local branch.
+    remote_branch="$branch"
+fi
+
+git branch -d "$branch"
+ret=$?
+if [[ $ret -ne 0 ]]; then
```
spacewander: Why not `git branch ... || exit $?`

bin/git-delete-branch
```diff
@@ -1,16 +1,94 @@
...
+    exit $ret
+fi
+
+git push "${remote}" ":${remote_branch}"
+ret=$?
+if [[ $ret -ne 0 ]]; then
```
spacewander: The exit status of the final command will always be the exit status of the script. No need to check it.

man/git-delete-branch.md
```diff
@@ -3,29 +3,44 @@ git-delete-branch(1) -- Delete branches
...
 
 ## OPTIONS
 
-  &lt;branchname&gt;
+-r
 
-  The name of the branch to delete.
-  If multiple branches are provided, then they will all be deleted.
+Specify a remote name and a remote branch name to use during the remote
+branch deletion.
```
spacewander: At my first glance, it seems we need to provide two arguments, one is remote name and the other is remote branch name. Maybe we could mention the format of arguments to make it clearer.

bin/git-delete-branch
```diff
@@ -1,16 +1,94 @@
...
+            usage
+            exit 1
+    fi
+elif [[ -z "$remote_branch" ]]; then
+    # In the event that a remote is specified but no remote branch is
+    # specified, then assume the branch name is the same as the local branch.
+    remote_branch="$branch"
+fi
+
+git branch -d "$branch"
```
spacewander: Could you tell me the reason to replace `branch -D` to `branch -d`?

bturrubiates: > Could you tell me the reason to replace branch -D to branch -d?

I had replaced it with `-d` because it warns about unmerged branches, but I could use `-D`.

bin/git-delete-branch
bturrubiates: > The exit status of the final command will always be the exit status of the script. No need to check it.

Thanks, will fix.

bin/git-delete-branch
bturrubiates: > Why not git branch ... || exit $?

Can do

bin/git-delete-branch
bturrubiates: > Better to use parameter expansion instead.

Hmm I thought using cut would be more clear. So is this what you're suggestion: 

So instead of: 
```sh
remote="$(echo ${OPTARG} | cut -d ':' -f 1)"
remote_branch="$(echo ${OPTARG} | cut -s -d ':' -f 2)"
```

you want: 

```sh
remote=${OPTARG%%:*}
remote_branch=${OPTARG##*:}
```

bin/git-delete-branch
bturrubiates: > It will be great if you can extract all usage... exit ... into a function.

Can do

man/git-delete-branch.md
bturrubiates: > At my first glance, it seems we need to provide two arguments, one is remote name and the other is remote branch name. Maybe we could mention the format of arguments to make it clearer.

What about: 

> Specify a remote name and an optional remote branch name to use during the remote
> branch deletion.

spacewander: I think `a remote name with an optional remote branch name` is a little better. Just change a little from your idea.

bin/git-delete-branch
spacewander: In my opinion, it is import to persist the original behaviour, unless the original one is wrong or not up-to-date.
Glad to see the old one back 😄

PR #578, opened 2016-09-12
count: remove usage of extra arguments for detailed display

tardypad:
When viewing detailed count information through the usage of the
--all argument, extra arguments would be also passed to the
git shortlog command (probably to be used as revisions list)
This behaviour was introduced in commit 7a60cd1.

Unfortunately the --all argument itself is passed along and results
in the counts of all refs to be displayed
This commit removed the usage of extra arguments and thus only
display the counts of the current branch

### DIFF
--- a/bin/git-count
+++ b/bin/git-count
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if test "$1" = "--all"; then
-  git shortlog -n -s $@ | awk '{print substr($0,index($0,$2)) " (" $1 ")"}'
+  git shortlog -n -s | awk '{print substr($0,index($0,$2)) " (" $1 ")"}'
   echo
 fi
 


tardypad:
Let me know if you think that it's worth keeping this behavior (being able to pass a revisions list as extra arguments after --all)
I will just make sure that "--all" is not passed as well
This behaviour was not documented so far, so we will need to update that too then.

tardypad:
I just saw that the initial commit was actually made with a `shift` before which then prevented this issue.
But this was surprisingly removed "on purpose" in aefe5c0 :confused:

tardypad:
Any thoughts on this?

If it was not clear, I'll put a more concrete example here
This shows the behavior issue of the current git count command
```
repo$ git init
repo (master) $ git commit --allow-empty --author "Foo <foo>" -m "commit 1"
repo (master) $ git commit --allow-empty --author "Foo <foo>" -m "commit 2"
repo (master) $ git count --all
Foo (2)
total 2
repo (master) $ git checkout -b new_branch
repo (new_branch) $ git commit --allow-empty --author "Foo <foo>" -m "commit 3"
repo (new_branch) $ git checkout master
repo (master) $ git count --all
Foo (3)
total 2
```
As you can see the result of the last command is problematic
The first part is the commits count calculated among all the branches,
whereas the total is only about the commits of the current branch

spacewander:
Merged. Thanks!

PR #581, opened 2016-09-20
update docs to clarify rename-branch

spacewander:
For the discussion in #576 
To catch up with the release cycle, I submit this pr.
Thanks for the idea from @hozefaj


hozefaj:
@spacewander Thanks. Looks good.

PR #584, opened 2016-09-23
fixes #583

### LINKED ISSUE #583: super easy to clear all work with `git clear`
I often type `clear` in my terminal to clear it out. Sometimes I might even hit enter twice. Today I typed `git clear` and hit enter twice. It proceeded to undo all of my work. What the heck. If this command needs to exist can it at least require you to type in "yes" or anything at all. It would have saved me many hours of doing work twice. Thanks.

`git-extras/bin/git-clear`
```

#!/usr/bin/env bash

echo "Sure? - This command may delete files that cannot be recovered, including those in .gitignore [Y/n]"
read ans
if [ "$ans" != "n" ] 
    then git clean -d -f -x && git reset --hard
fi
```

grindhold:
put default option in security-answer to 'no'

### DIFF
--- a/bin/git-clear
+++ b/bin/git-clear
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-echo "Sure? - This command may delete files that cannot be recovered, including those in .gitignore [Y/n]"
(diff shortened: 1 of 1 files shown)


spacewander:
Thanks for the quick fix.

PR #585, opened 2016-10-06
Add git stamp

tardypad:
Continuing the moving of some of my local git aliases and functions
in order to make them more maintainable and so that other people can use them too :grin:

This adds a new stamp command: Stamp the last commit message
See the [manual](https://github.com/tardypad/git-extras/blob/stamp/man/git-stamp.md) for the use cases and examples

The command is best used afterwards with personal aliases such as `issue = stamp -r Issue`
The usage is then `git issue FOO-142` to stamp the commit with an issue number

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -997,16 +997,17 @@ Issue FOO-456 #close
 Review https://reviews.foo.org/r/4567/
 ```
 
-Replace previous issues with a new one
+Replace previous issues with a new one  
+(Note that the identifier is case insensitive)
 
 ```bash
-$ git stamp --replace Issue BAR-123
+$ git stamp --replace issue BAR-123
 
 Fix timezone bug
 
 Review https://reviews.foo.org/r/4567/
 
-Issue BAR-123
+issue BAR-123
 ```
 
 
--- a/Commands.md
+++ b/Commands.md
@@ -974,13 +974,26 @@ Reference the issues numbers from your bug tracker
 
 ```bash
 $ git stamp Issue FOO-123
+
+commit 787590e42c9bacd249f3b79faee7aecdc9de28ec
+Author: Jack <jack@work.com>
+Commit: Jack <jack@work.com>
+
+    Fix timezone bug
+
+    Issue FOO-123
+
 $ git stamp Issue FOO-456 \#close
 
-Fix timezone bug
+commit f8d920511e052bea39ce2088d1d723b475aeff87
+Author: Jack <jack@work.com>
+Commit: Jack <jack@work.com>
 
-Issue FOO-123
+    Fix timezone bug
 
-Issue FOO-456 #close
+    Issue FOO-123
+
+    Issue FOO-456 #close
 ```
 
 Link to its review page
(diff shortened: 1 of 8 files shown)


Commands.md
```diff
@@ -928,6 +929,58 @@ For example, running `git show-tree` will display:
...
+$ git stamp --replace issue BAR-123
+
+Fix timezone bug
+
+Review https://reviews.foo.org/r/4567/
+
+issue BAR-123
+```
+
+
```
spacewander: In my opinion,  it will be greater if you also show the change of last commit via `git log` in this introduction.

bin/git-stamp
```diff
@@ -0,0 +1,92 @@
...
+  local stamp_msg
+  [[ -n "${MSG}" ]] && stamp_msg="${ID} ${MSG}" || stamp_msg="${ID}"
+  
+  if ${REPLACE}; then
+    # remove previous stamps with same ID from the commit message
+    commit_msg=$(
+      echo "${commit_msg}" \
+        | grep --ignore-case --invert-match "^${ID}\b" \
+        | cat --squeeze-blank
+    )
```
spacewander: Here is a corner case:
What will happen if commit message without stamp shares same prefix with `$ID`?
For example, run `git stamp 'stamp:' blahblah` on your branch.

spacewander: Maybe this case is unavoidable... Warn it in the document?

Commands.md
tardypad: You mean showing the exact result of the commands as it they were executed?
Meaning with the commit hash, author name and date as well

bin/git-stamp
```diff
@@ -0,0 +1,92 @@
...
+
+Options:
+    -r, --replace    replace all previous stamps with same id
+EOF
+}
+
+
+error() {
+  if [[ -n "$1" ]]; then
+    local msg=$( echo "error: $1" | sed 's/\\n/\\n       /g' )
```
tardypad: Ah that a copy paste from my previous `git reauthor` command
https://github.com/tardypad/git-extras/blob/stamp/bin/git-reauthor#L32
This was for multiple lines error message so that the second line is aligned with the `:` as in
``` 
error: missing target of the rewrite
       use either --old-email option or --all flag
```
It's not needed here, I better remove it then I guess

bin/git-stamp
tardypad: Indeed this is "unavoidable" the way it is
That's why I show the new commit message as a result, so that people can check that nothing went wrong.
But it's better to be explicit and I'll put a warning in the documentation

etc/bash_completion.sh
```diff
@@ -134,6 +134,10 @@ _git_scp(){
   __git_complete_remote_or_refspec
 }
 
+_git_stamp(){
+  __gitcomp '--replace -r'
+}
+
```
tardypad: Alright I'll try to add it since that should be a simple one
I am not sure I'll be able to test it though
Configuring and switching to ZSH is still on my TODO list :)

Commands.md
spacewander: @tardypad 
Yes.

spacewander:
Waiting for your update...
Since we are making a new release this weekend, the `git stamp` will be included in next release.

tardypad:
Every review comment should be fixed within those last commits

nicolaiskogheim:
@spacewander Any final thoughts?

spacewander:
@tardypad 
Sorry for the delay... Could you rebase the pull request so that we can merge it directly?

tardypad:
@spacewander I've rebased my branch against master yesterday
I'm not sure if you receive a notification in that case :thinking:
So here is this message

spacewander:
Now I received the notification via email, I think it is time to merge the
pr. :)

2016-12-05 17:32 GMT+08:00 Damien Tardy-Panis <notifications@github.com>:

spacewander:
Just introduced a new command to `git-extras`. Thank you!

PR #595, opened 2016-11-14
fix: git-squash docs. Replaced 'actual' with 'current'

aaguilera:
Fixed a small typo in git-squash documentation.


spacewander:
Thanks!

PR #600, opened 2016-11-19
git-fresh-branch: fix handling of 'yes' response when there are changes

richardfearn:
`break` isn't valid in its current location in the script, resulting in an error like:

    /usr/bin/git-fresh-branch: line 42: break: only meaningful in a `for', `while', or `until' loop

Since there's a call to `clean` right at the end of the script, do nothing if anything beginning `Y` or `y` is entered; execution will continue to the end of the script, and `clean` will be called.


spacewander:
Thanks for your fix!

PR #601, opened 2016-11-19
git-sed: use "grep -z" and "xargs -0" to cope with spaces in filenames

richardfearn:
Otherwise filenames with spaces result in errors:

    $ cat "file with spaces in name"
    old

    $ git sed old new
    sed: can't read file: No such file or directory
    sed: can't read with: No such file or directory
    sed: can't read spaces: No such file or directory
    sed: can't read in: No such file or directory
    sed: can't read name: No such file or directory

### DIFF
--- a/bin/git-sed
+++ b/bin/git-sed
@@ -64,6 +64,6 @@ actual command:
     shift
 done
 
-command="git grep -l '$search' | xargs sed -i 's/$search/$replacement/$flags'"
-git grep -l "$search" | xargs sed -i "s/$search/$replacement/$flags"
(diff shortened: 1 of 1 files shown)


spacewander:
Well, it seems BSD grep's `-z` option is different from the GNU one: http://www.unix.com/man-page/FreeBSD/1/grep/
Therefore, we need to use `--null` instead.

richardfearn:
Sorry, my commit message was slightly wrong and should have said "**git** grep -z" instead of "grep -z". That's now fixed.

Since it's the `git grep` command as opposed to regular `grep`, I think `-z` might be OK...

spacewander:
Merged. Thanks!

PR #603, opened 2016-11-19
git-clear-soft: change default yes/no option to "no" (as with git-clear)

richardfearn:
6d49449 changed the `git-clear` default option to "no" (see #583).

Make the `git-clear-soft` default the same - it's safer.


hemanth:
Thank you!

PR #606, opened 2016-12-07
fix git-extras update and tweak Windows detection

spacewander:
Replace expr with bash substring expansion, which fixes some syntax error.
Tested under:
GNU bash, version 4.3.42(5)-release (x86_64-pc-msys)
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)


hemanth:
It's just P.E it shall be fine across.

Issue #609, opened 2016-12-25
Misspelling in man page about git-archive-file(1)


loovjo:
When running `git extras --help`, on the line about `git-archive-file`, it says `...to a archive.`. This should be `to an archive`.

nicolaiskogheim:
It appears that there are a lot of that particular spelling.

```sh
git-extras$ grep -n "to a archive" **/* 2>/dev/null
etc/git-extras-completion.zsh:359:    archive-file:'export the current head of the git repository to a archive' \
man/git-archive-file.1:7:\fBgit\-archive\-file\fR \- Export the current HEAD of the git repository to a archive
man/git-archive-file.1:13:Export the current HEAD of the repository into a archive with a identifiable and unique name\.
man/git-archive-file.html:6:  <title>git-archive-file(1) - Export the current HEAD of the git repository to a archive</title>
man/git-archive-file.html:74:  <code>git-archive-file</code> - <span class="man-whatis">Export the current HEAD of the git repository to a archive</span>
man/git-archive-file.html:83:<p>Export the current HEAD of the repository into a archive with a identifiable and unique name.</p>
man/git-archive-file.md:1:git-archive-file(1) -- Export the current HEAD of the git repository to a archive
man/git-archive-file.md:10:Export the current HEAD of the repository into a archive with a identifiable and unique name.
man/git-extras.1:36:\fBgit\-archive\-file(1)\fR Export the current HEAD of the git repository to a archive
man/git-extras.html:98:<li> <strong><a class="man-ref" href="git-archive-file.html">git-archive-file<span class="s">(1)</span></a></strong> Export the current HEAD of the git repository to a archive</li>
man/git-extras.md:25:   - **git-archive-file(1)** Export the current HEAD of the git repository to a archive
```

spacewander:
@loovjo @nicolaiskogheim 
Would you send a pr to substitute the spelling, and regenerate html files and man pages? Thanks!

nicolaiskogheim:
Coming up!

nicolaiskogheim:
#610

PR #610, opened 2016-12-26
spelling: {a,an} archive

nicolaiskogheim:
#609 and an occurence of `a identifiable`. 

The `git extras` man page also got updated with an entry for `git reauthor` which probably just was forgotten when it was added.


spacewander:
Merged. Thanks!

PR #617, opened 2017-01-23
[Git Summary] Show email address

Dark-Water:
I need the email address on the summary to work easier with `git reauthor`

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -49,7 +49,7 @@ file_count() {
 #
 
 authors() {
-  git shortlog -n -s $commit | LC_ALL=C awk '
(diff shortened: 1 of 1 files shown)


hemanth:
I would vote for passing a flag to `author` rather than having the `-e` flag hardcoded.

nicolaiskogheim:
> I would vote for passing a flag to author

Just my thoughts as well.

bl-ue:
I think we should close this PR.  It's old, has conflicts, and the author is obviously not coming back...
And there are clearly no intentions to fix it per the collaborators comments above.

Issue #614, opened 2017-01-13
[Feature request] ReAuthor replace by name


TobiasSekan:
Hello

It is possible to add search and replace by name for`git reauthor` ?

`$ git reauthor --old-name 'J. F.' --correct-email jack@work.com --correct-name 'Jack Foobar'`

My problem:
After importing repository from old SVN and HG, I miss a few email addresses on different authors and it is replaced by `devnull@localhost`

Sincerely yours

PR #621, opened 2017-01-24
[Git Summary] Replace by name

Dark-Water:
Added option to search for old user name instead of old email address 
* Help to fix repository commits that have identical emails but different users
* Help to clean up repository commits after (broken) imports from HG and SVN

### DIFF
--- a/bin/git-reauthor
+++ b/bin/git-reauthor
@@ -7,6 +7,7 @@ init_variables() {
   CONFIG=false
   ALL=false
   unset OLD_EMAIL
+  unset OLD_NAME
   unset CORRECT_EMAIL
   unset CORRECT_NAME
   TYPE='both'
@@ -22,7 +23,8 @@ Options
     -c, --use-config              define correct values from user Git config
     -e, --correct-email  <email>  define the correct email to set
     -n, --correct-name   <name>   define the correct name to set
-    -o, --old-email      <email>  rewrite identities matching old email in commits and tags
+    -oe, --old-email     <email>  rewrite identities matching old email in commits and tags
+    -on, --old-name      <name>   rewrite identities matching old name in commits and tags
     -t, --type           <id>     define the type of identities affected by the rewrite
                                   author, committer, both (default)
 EOF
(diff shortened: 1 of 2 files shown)


Dark-Water:
Fix for #614

bin/git-reauthor
```diff
@@ -22,7 +23,8 @@ Options
     -c, --use-config              define correct values from user Git config
     -e, --correct-email  <email>  define the correct email to set
     -n, --correct-name   <name>   define the correct name to set
-    -o, --old-email      <email>  rewrite identities matching old email in commits and tags
+    -oe, --old-email     <email>  rewrite identities matching old email in commits and tags
```
spacewander: It would be great to keep the old option. For example, `-oe, --old-email, -o`.
And don't forget to update the man page and Commands.md, so that everyone can find your new feature!

tardypad:
The changes looks fine to me but there is a case that needs to be cleared
What happens in case both arguments `--old-email` and `--old-name` are set?
Should the script update the commits with match both or only one of them?

From what I see in the current implementation, the commit is updated if any of the email or name matches.
To me, it would makes more sense if the match would be done on both of them
But that'll indeed makes the code a bit more complex (that's why I didn't include this option initially :stuck_out_tongue:)

In any case this needs to be made clear in the docs.

Dark-Water:
Yes you are right,
I will see how can I implement this for both options
(if inside bash is a litte bit horrible :alien:)

tiemonl:
Is there any update on this? This is a feature I would very much like.

spacewander:
@tiemonl 
It seems that the original author no longer works on this PR. Will you take over the code, and implement the feature that @tardypad mentioned?

tiemonl:
@spacewander I was messing with it a bit yesterday and I found out a code change isn't completely necessary, except for clarity. For example, you have these authors:
```
John Smith <john.smith@email.com>
smithj <john.smith@email.com>
```
You can see it's the same person but for some commits it used their name and for the others it used their username. An example command to change just the name would be: 
```
git reauthor --old-email john.smith@email.com --correct-email john.smith@email.com --correct-name smithj
```
Resulting AUTHORS file:
```
smithj <john.smith@email.com>
```
Explanation: The command gets all the users with the specified email, keeps the email but changes the name associated to the email.

Easy fix: change git-author.md to include this example.
Hard fix: Add logic to have a name variable.

Which fix would be preferred?

spacewander:
@tiemonl 
I think you could simply update the documentation.
If you have much free time, you could also introduce a new option to deduplicate via name instead of email.

tiemonl:
I will make the documentation change for now, and possibly open another pull request to introduce the new option if I have time/can get it working

spacewander:
@tiemonl 

> introduce a new option to deduplicate via name instead of email

I need to clarify myself that I mean to add a new option in `git author` to deduplicate the authors via their name instead of the email when generating the `AUTHORS` file.

tiemonl:
@spacewander 
My mistake. I misspoke. I used git AUTHORS as an example, but you are correct in the fact that authors will need to be changed too. My original intention was to change the output in `git summary`.

For example running `git summary` in one of my repos returns this: 
Screen Shot 2019-05-22 at 8 27 21 AM
Liam and tiemonl is me and it bothers me that I am showing twice with the same email, but different name. The quick workaround for now was the documentation change I made in #753 which results in this:
Screen Shot 2019-05-22 at 8 30 59 AM

`git author` I've noticed only shows one username per email, and it won't duplicate and show multiple names with the same email, it will only show the first username attached to that email. So looking at the first screenshot I've attached. Running `git authors` would only return:
```
tiemonl <myemail@email.com>
```
and not:
```
tiemonl <myemail@email.com>
Liam <myemail@email.com>
```

For the permanent fix, I would need an `--old-user` option for `git reauthor`, but I was just adding a temporary workaround for people who just wanted to rename for the time being, until I have some time to look into implementing `--old-user`

A fix could also be made for `git authors` to show multiple user names with the same email attached. I'm hoping to have some time this weekend to look into the `reauthor fix` and possibly the `author` fix if you think it is needed? Let me know.

I apologize for the confusion I made with my first example.

spacewander:
@tiemonl 

OK, now I understand what you mean.

> A fix could also be made for git authors to show multiple user names with the same email attached. I'm hoping to have some time this weekend to look into the reauthor fix and possibly the author fix if you think it is needed? Let me know.

It's up to you. Feel free to do contribution when you have spare time. PR is always welcome!

PR #622, opened 2017-01-27
effort: handling of $'s and \n's, plus make shellcheck happy

nicolaiskogheim:
This fixes an issue that was reported this week, with `$`'s in a filename breaking `git log` and take `git effort` with it. `git effort` now understands how to treat files with actual newlines in them as well.

[Shellcheck](https://github.com/koalaman/shellcheck) had some complaints, and I agreed to all of them. Because of that, there is a lot of code to review, so please look at each commit individually, if that makes it easier.

While I were at it, I expanded upon the docs, and made some small rewordings. If there are many comments on the doc edits, I'll split that out in it's own PR.

______

**Please help test this**

- [x] macOS
- [x] Linux
- [ ] BSD

You can use [this repo](https://github.com/nicolaiskogheim/testrepogiteffort) as a test repo. It contains files with names that should produce an error if the issue isn't fixed.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -17,7 +17,10 @@ usage() {
 # get dates for the given <commit>
 #
 dates() {
-  eval "git log $args_to_git_log --pretty='format: %ad' --date=short \"$1\""
+  path="$1"
+  # Escape dollars so they aren't interpreted in eval
+  path="${1/\$/\\$}"
+  eval "git log $args_to_git_log --pretty='format: %ad' --date=short \"${path}\""
 }
 
 # tput, being quiet about unknown capabilities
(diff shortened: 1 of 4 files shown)


bin/git-effort
```diff
@@ -80,41 +83,63 @@ color_for() {
...
 effort() {
     path=$1
+
+    # Exit if path == ""
+    # This is not an error
+    # shellcheck disable=SC2116
+    if [ -z "$(head -c 1 <<<"$(echo "$path")")" ]
+    then
+      exit 0
+    fi
```
nicolaiskogheim: It looks like printf (that produces output to `effort`) sends a final empty line when we call `effort`. That's the reason for this check.

spacewander:
Sorry for my missing of this.
Pass under Linux with bash 4.3.30.

@nicolaiskogheim 
Could we merge the pr now?

nicolaiskogheim:
I'll fire up a FreeBSD box tomorrow and run the script just to be sure.

PR #627, opened 2017-02-02
Added Github Enterprise support

hickey:
Added support for forking on instances of Github other than github.com. 

In addition added ability to pull stored settings from git config so that one is not prompted each time for things like username and such. 

Also introduced a self documenting way of including helper scripts into the generated scripts. See the Makefile and helper/config-value for details. Also helps keeping one from forgetting to update the needs_* files.


hickey:
Found that some Github enterprise installations need to have a path prepended to /repos/..... Will update code and create a new pull request.

PR #628, opened 2017-02-03
Feature/git fork

hickey:
Added support for forking on instances of Github other than github.com.

In addition added ability to pull stored settings from git config so that one is not prompted each time for things like username and such.

Also introduced a self documenting way of including helper scripts into the generated scripts. See the Makefile and helper/config-value for details. Also helps keeping one from forgetting to update the needs_* files.

### DIFF
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -44,8 +44,6 @@ if [[ -z "$user" ]]; then
 fi
 
 if [[ -z "$token" ]]; then
-  echo "Enter github password"
-  read password
   echo "Enter github two-factor authentication code (leave blank if not set up)"
   read MFA_CODE
 fi
@@ -55,10 +53,10 @@ header_info=''
 if [[ -n "$token" ]]; then
   header_info="-H \"Authorization: token ${token}\""
 elif [[ -n "$MFA_CODE" ]]; then
-  auth_info="-u \"$user:$password\""
+  auth_info="-u \"$user\""
   header_info="-H \"X-GitHub-OTP: $MFA_CODE\""
 elif [[ -n "$password" ]]; then
-  auth_info="-u \"$user:$password\""
+  auth_info="-u \"$user\""
 else
   echo "No login credentials specified."
   exit 1
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -64,9 +64,9 @@ fi
 
 # create fork
 IFS="'" cmd="curl -qs -X POST $auth_info $header_info https://$api_server/${api_prefix}repos/$owner/$project/forks"
-eval $cmd >/dev/null 
+eval $cmd  | grep "message" >/dev/null
 
-[ $? = 0 ] || abort "fork failed"
+[ $? = 0 ] && abort "fork failed"
 
 use_ssh=$(get_config_value "$server.use-ssh")
 if [[ -z "$use_ssh" ]]; then
@@ -103,10 +103,11 @@ else
       git clone "${remote_prefix}${user}/${project}.git" "$project" 2>/dev/null
     done
 
-    if [[ -d "$proejct" ]]; then
+    if [[ -d "$project" ]]; then
       # add reference to origin fork so can merge in upstream changes
       cd "$project"
-      git remote add upstream "${remote_prefix}${owner}/${project}.git"
+      git remote add upstream "${remote_prefix}${owner}/${project}.git" 2>/dev/null
+      [[ $? > 0 ]] && echo "WARN: upstream reference already exists"
       git fetch upstream
     else
       echo "Github did not fork within 60 seconds or issue cloning repo."
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -10,42 +10,73 @@ test -z "$url" && url=$(git remote get-url origin 2> /dev/null) && origin=true
 # validate repo url
 test -z "$url" && abort "github repo needs to be specified as an argument"
 
-# validate user
-echo "Enter your github username"
-read user
-[ -n "$user" ] || abort "git username required"
-echo "Enter github two-factor authentication code (leave blank if not set up)"
-read MFA_CODE
-
 # extract owner + project from repo url
 project=${url##*/}
 owner=${url%/$project}
 project=${project%.git}
-if [[ $owner == git@* ]]; then
-  owner=${owner##*:}
-else
-  owner=${owner##*/}
-fi
+owner=${owner##*[/:]}
+# Yes, the following extracts server for both SSH and https references
+server=${url##*@}
+server=${server##*://}
+server=${server%%[/:]*}
 
 # validate
 [ -z "$project" -o -z "$owner" ] && abort "github repo needs to be specified as an argument"
 
+# determine github credentials
+user=$(get_config_value "$server.user")
+token=$(get_config_value "$server.token")
+if [[ $(get_config_value "$server.add-api") == "true" ]]; then
+  api_server="api.$server"
+else
+  api_server=$server
+fi
+
+if [[ -z "$user" ]]; then
+  # validate user
+  echo "Enter your github username"
+  read user
+  [ -n "$user" ] || abort "git username required"
+fi
+
+if [[ -z "$token" ]]; then
+  echo "Enter github password"
+  read password
+  echo "Enter github two-factor authentication code (leave blank if not set up)"
+  read MFA_CODE
+fi
+
+auth_info=''
+header_info=''
+if [[ -n "$token" ]]; then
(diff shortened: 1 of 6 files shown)


bin/git-fork
```diff
@@ -10,42 +10,77 @@ test -z "$url" && url=$(git remote get-url origin 2> /dev/null) && origin=true
...
+if [[ -z "$user" ]]; then
+  # validate user
+  echo "Enter your github username"
+  read user
+  [ -n "$user" ] || abort "git username required"
+fi
+
+if [[ -z "$token" ]]; then
+  echo "Enter github password"
+  read password
```
spacewander: I prefer let curl ask for my password instead of passing it in the option.

bin/git-fork
```diff
@@ -55,6 +90,12 @@ if [ "$origin" = true ]; then
 else
     # clone forked repo into current dir
     git clone "${remote_prefix}${user}/${project}.git" "$project"
+    until [[ -d "$project" ]]; do
+      echo "Github is being slow today. Waiting 10 secs to attempt clone."
```
spacewander: There are various causes that should be responsible for the failure of `git clone`.
Fortunately, the error message from `git clone` is clear enough.
Better shorten the message and remove our guess.

bin/git-fork
```diff
@@ -55,6 +90,12 @@ if [ "$origin" = true ]; then
 else
     # clone forked repo into current dir
     git clone "${remote_prefix}${user}/${project}.git" "$project"
```
spacewander: Maybe we could check if $project exists before cloning into it?

bin/git-fork
hickey: Understandable. Although one is already being asked for the MFA code. Does that make sense to be asked for the MFA code before the password?

bin/git-fork
hickey: Git clone will not allow one to clone over an already existing repo directory. 
````
0:0 ᐅ git clone git@github.com:hickey/git-extras.git git-extras
fatal: destination path 'git-extras' already exists and is not an empty directory.
````

File: bin/git-fork (hunk 2 above)
Comment (hickey) [inline bin/git-fork]: The issue I see there is that git clone is **very** verbose and just adds a lot of cruf to the terminal that does not need to be there (see below). Although I just realized that I should be sending stderr to /dev/null to avoid the excess cruf. 

Since the fork is asynchronous there is no way to determine when the fork is complete other than trying it. If I execute the fork command against github.com, the repo exists by the time that the clone is attempted. But if I try the fork command against the internal Github Enterprise instance, I end up waiting at least once for the fork to complete. (probably the fast LAN vs 200 ms across the net). 

The more that I think about it, the attempt to clone should probably only be attempted for about a minute or so. After that point the command should exit with an error code and a message that the repo could not be cloned locally. 

````
0:128 ᐅ git clone git@github.com:hickey/fioo.git foo  
Cloning into 'foo'...
ERROR: Repository not found.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
````

File: bin/git-fork (hunk 1 above)
Comment (spacewander) [inline bin/git-fork]: @hickey 
I think the order is not critical. Just because let curl ask for your password is simpler.

File: bin/git-fork (hunk 3 above)
Comment (spacewander) [inline bin/git-fork]: @hickey 
Ah ha, this idea is inspired by your `-d $project` check.
If there is `$project` already, we can directly exit the script, skip `git clone`, `cd`, and so on.

File: bin/git-fork (hunk 1 above)
Comment (spacewander) [inline bin/git-fork]: @hickey
I think the order is not critical. Just because let curl ask for your password is simpler.

**Comment by hickey** [comment]
Added better detection when the fork fails (which is really a comms error with talking to Github). 

If directory already exists, clone will silently fail and the upstream references will be created. If they already exist, then a warning message is printed that it exists.

**Comment by hickey** [comment]
If you would prefer, I can squash all the commits and resubmit the PR.

File: bin/git-fork [hunk 4]
```diff
@@ -54,9 +87,30 @@ if [ "$origin" = true ]; then
     git fetch origin
 else
     # clone forked repo into current dir
-    git clone "${remote_prefix}${user}/${project}.git" "$project"
-    # add reference to origin fork so can merge in upstream changes
-    cd "$project"
-    git remote add upstream "${remote_prefix}${owner}/${project}.git"
-    git fetch upstream
+    git clone "${remote_prefix}${user}/${project}.git" "$project" 2>/dev/null
```
spacewander: I agree that the output of `git clone` is cumbersome, but still think we should not hide the error message. They are like low level log, which may be useful sometimes.

bin/git-fork
```diff
@@ -54,9 +87,30 @@ if [ "$origin" = true ]; then
...
+
+    # Check to make sure we cloned the repo. Backoff exponetially and try again
+    timeout=( 2 2 2 6 6 6 20 20 )
+    total_wait=0
+    until [[ (( "$total_wait" > 60 )) || -d "$project" ]]; do
+      echo "Github is being slow today. Waiting $timeout secs to attempt clone."
+      sleep ${timeout[0]}
+      total_wait=$(expr $total_wait + ${timeout[0]})
```
spacewander: Better to use `total_wait=$((total_wait + timeout[0]))`, which is more modern.

bin/git-fork
```diff
@@ -54,9 +87,30 @@ if [ "$origin" = true ]; then
...
-    # add reference to origin fork so can merge in upstream changes
-    cd "$project"
-    git remote add upstream "${remote_prefix}${owner}/${project}.git"
-    git fetch upstream
+    git clone "${remote_prefix}${user}/${project}.git" "$project" 2>/dev/null
+
+    # Check to make sure we cloned the repo. Backoff exponetially and try again
+    timeout=( 2 2 2 6 6 6 20 20 )
+    total_wait=0
+    until [[ (( "$total_wait" > 60 )) || -d "$project" ]]; do
```
spacewander: Actually, it will wait for 64(44 plus the last 20) seconds.

man/git-fork.1
```diff
@@ -40,7 +40,71 @@ adds the forked repo as a remote called \fBorigin\fR
 .P
 Remotes will use ssh if you have it configured with GitHub, if not, https will be used\.
 .
+.P
+Your Github settings can not be saved as git config values instead of specifying them each time\. To enable this you need to execute a few git config commands like the following\.
```
sacres: I believe you meant now rather than "not be saved" here.

man/git-fork.html
```diff
@@ -99,8 +99,52 @@ <h2 id="DESCRIPTION">DESCRIPTION</h2>
 
 <p>  Remotes will use ssh if you have it configured with GitHub, if not, https will be used.</p>
 
+<p>  Your Github settings can not be saved as git config values instead of
```
sacres: And here as well: s/not be saved/now be saved/

man/git-fork.md
```diff
@@ -21,8 +21,49 @@ git-fork(1) -- Fork a repo on github
 
   Remotes will use ssh if you have it configured with GitHub, if not, https will be used.
 
+  Your Github settings can not be saved as git config values instead of
```
sacres: And here as well: s/not be saved/now be saved/

bin/git-fork
hickey: I thought about that, but some of the older shells don't understand the double parenthesis. So I opted for the more traditional form.

hickey:
Sorry, I have been busy and this PR fell off the list. Updated wording in the documentation and surprised that I missed that typo. Thanks.

hickey:
Let me know if you would like this PR squashed.

helper/config-value
```diff
@@ -0,0 +1,11 @@
+# needed_by: git-fork
+
+get_config_value() {
+  echo $(git config git-extras.$1)
+}
+
+# set_config_value "key" "value" "global|system" (global or system not required)
+set_config_value() {
+  $(git config ${3+--$3} git-extras.$1 $2)
+  echo $?
```
spacewander: IMHO, I prefer to use `git config ${3+--$3} git-extras.$1 $2` directly.
When writing shell script, people are more relied on  `$?` that the return value to check if the previous command succeeded.

bin/git-fork
spacewander: Still insist the need for the output of `git clone`.

adriaanzon:
You may want to use the curl's `-f` flag like I did in #649, instead of checking if the response contains "message"

bl-ue:
We should close this — it's out-of-date and 4 years old, and has no votes.

PR #633, opened 2017-03-04
Columns in git-effort adopt to file length

nschlimm:
The columns in git-effort end up in a mess when you have longer file names.
I have fixed that so the column width matches the longest file in the current git repo.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -5,6 +5,7 @@ above=0
 # if the output won't be printed to tty, disable the color
 test -t 1 && to_tty=true
 color=
+columns=45
 
 #
 # print usage message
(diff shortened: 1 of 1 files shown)


bin/git-effort
```diff
@@ -5,6 +5,7 @@ above=0
 # if the output won't be printed to tty, disable the color
 test -t 1 && to_tty=true
 color=
+columns=45
```
spacewander: `colums=45` seems duplicate now, because you will calculate it from longest filename later.

nschlimm: taklen that out. I have another branch where you can set the columns as option on the command line when calling git effort, this was the default setting. however it's not necessary here ...

spacewander:
@nschlimm 
Merged. Thanks for your contribution!

Issue #634, opened 2017-03-06
git-rename-branch


lagden:
```
OSX
fish, version 2.3.1
git version 2.9.0
```

I try rename a branch:
```
(branch-a) $ git rename-branch branch-a
/usr/local/bin/git-rename-branch: line 31: $2=${git symbolic-ref --short -q HEAD}: bad substitution
```

spacewander:
@lagden
Hi, I just submit a pull request #635 to fix your issue.
Could you try the code there and tell us whether the fix works for you?
Thanks!

lagden:
@spacewander 
Works fine now!! tks! 😃 

**Sugestion**

Avoid rename remote if not exist the `origin`

```
~/t (abc.4.5.6) $ git rename-branch qwe.123 abc.4.5.6
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: 'origin' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
~/t (qwe.123) $
```

PR #636, opened 2017-03-08
use remote detected instead of 'origin' in rename-branch

spacewander:
This change will deal with the case in #634 smoothly.
@lagden
Is this what you prefer?


spacewander:
Merged because this pr solves same problem as #641 in `git-rename-branch`.

Issue #645, opened 2017-04-09
gpg-signed commits in git-standup


sbriskin:
In the kamranahmedse/git-standup#40 was made a nice feature to show gpg-signed commits 
with a `-g` flag. Is it possible to make the same in git-extras?

spacewander:
Closed because #646 has been merged.

PR #646, opened 2017-04-12
Upgrade git-standup to new interface

spacewander:
For issue #645 
Now we can use `-g` option to show GPG info.

@sbriskin
Maybe it is something your want?
Remember to run `git standup` with new style options, for example,
`git standup -a $author -d $days_ago -g`.


sbriskin:
That's exactly what i need! Thank you!

PR #649, opened 2017-04-27
git-fork: Let curl exit with error code on fail

adriaanzon:
The next line checks if curl failed, and aborts the script if it has.

This adds the `-f` flag to the curl call, so when the request failed (i.e. when you typed a wrong password), the script aborts appropriately, instead of creating a branch for a non-existent remote.


spacewander:
@adriaanzon 
LGTM. Merged!
Thank you for your patch!

PR #656, opened 2017-05-16
Fix typo in git-extras

dueringa:
Output said "Setting up 'git-exras" instead of "Setting up 'git-extras"


spacewander:
Good catch. Merged!

PR #658, opened 2017-06-13
Fix sample install without root

ilyar:


spacewander:
`make install PREFIX=$HOME/software` is just an example, doesn't mean people should install `git-extras` under `$HOME/software`.
Moreover, `.local` is only part of XDG standard. It's not required to be existed in every unix environment.
Compared with `.local`, `software` is more likely to be a placeholder, instead of a real destination.

Issue #186, opened 2013-09-05
git-feature/bug/refactor does not create remote branch


stuplum:
I would expect using the git-feature/bug/refactor alias to create both a local and remote branch, like git-create-branch does. Is this by design?

gphat:
I would love for this to be an option!

nicolaiskogheim:
Hi,

I think this is a valid feature request, and I'll take care of it after we agree on how this should look, IF we do, that is.

> Is this by design?

I would guess it just didn't occur to the creator of this command.


Anyway, I propose to add a flag to create the remote branch.
It could be called `-c/--create-remote-branch`, or `-p/--push(-branch)`.

Now, people tend to have multiple remotes with all kinds of names, and we cannot assume that the created branch should use the same remote as the branch it was based on.

For example, if I fork a repo and want to create a feature based off `master`, I would like the branch pushed to my fork, and not to `upstream` (which may or may not be the remote that `master` is pointing to).

And sometimes I might want to push to the `upstream` remote, so we cannot assume `origin` either.

As the examples show, we might want to let our new option take an argument. We should add autocompletion on remote names in that case.


We need to figure out how this is supposed to work. Feedback is much appreciated.

mapitman:
Any movement on this? This seems broken to me as when I type:
```
git feature finish my-feature-branch
```
I receive this:
```
Deleted branch feature/my-feature-branch (was ea6e433).
error: remote-tracking branch 'origin/feature/my-feature-branch' not found.
remote: warning: Deleting a non-existent ref.
```

It seems as though the `finish` command is expecting that the `feature` command created a remote branch.

mapitman:
PR #664 should close this issue.

PR #664, opened 2017-07-08
Fix Issue #186

mapitman:
Use `git create-branch` instead of `git checkout -b` in:
git-bug
git-chore
git-feature
git-refactor

Since `git-finish` uses `git-delete-branch` (which expects the branch to
have been pushed to origin) we should use `git-create-branch` to create
it.

### DIFF
--- a/bin/git-bug
+++ b/bin/git-bug
@@ -10,5 +10,5 @@ else
     echo "too many arguments; if not finishing a bug, only <name> of new bug is expected" && exit 1
   fi
   branch=bug/$1
-  git checkout -b $branch
+  git create-branch $branch
 fi
--- a/bin/git-chore
+++ b/bin/git-chore
@@ -10,5 +10,5 @@ else
     echo "too many arguments; if not finishing a chore, only <name> of new chore is expected" && exit 1
   fi
   branch=chore/$1
-  git checkout -b $branch
+  git create-branch $branch
 fi
--- a/bin/git-feature
+++ b/bin/git-feature
@@ -37,5 +37,5 @@ else
   else 
     branch="$branch_prefix"/"${argv[0]}"
   fi
-  git checkout -b $branch
+  git create-branch $branch
 fi
--- a/bin/git-refactor
+++ b/bin/git-refactor
@@ -10,5 +10,5 @@ else
     echo "too many arguments; if not finishing a refactor, only <name> of new refactoring is expected" && exit 1
   fi
   branch=refactor/$1
-  git checkout -b $branch
+  git create-branch $branch
 fi


spacewander:
@mapitman 
Personally speaking, I perfer to change the behavior of `git-delete-branch`.
When remote could not be found, we could print a message and skip deleting remote branch, instead of falling back to 'origin'.

The `git-create-branch` will set `origin` as upstream, which is not always the favorite.

mapitman:
I'd like to understand this a little better.

1. When you use `git create-branch somebranch` it creates the branch and pushes to origin, with no option to change this behavior.
1. When you use `git delete-branch somebranch` it deletes the branch locally and pushes the delete to origin, with no option to change this behavior.
1. When you use `git feature somebranch` it currently creates the branch locally but does not push to origin.
1. When you use `git feature finish somebranch` it currently merges the feature branch, deletes the local feature branch and attempts to delete the branch at origin as well and errors if it does not exist there.

I don't understand why it is OK for 1 to push the branch to origin but not OK for 3 to do so as well. As a matter of fact, it makes more sense to me for the `git feature` commands to automatically push to origin since they make up a workflow and automate several commands.

spacewander:
@mapitman 
It is my wrong not to explain it clearly.

First of all, push to `origin` after creating a new branch is not always what people want. Sometimes they want to push to other upstream.
Actually, `git delete-branch` doesn't push the delete to origin directly. It will try to find the correct upstream, and use `origin` as fallback. Therefore, in most cases it works as expected. Sometimes it will try to delete non-existent remote-tracking branch because the fallback.

Secondly, we now have a plan to support pushing to a specific upstream via extra arguments. (It would be great if you would like to send a pr for this!)

mapitman:
OK, I see what you mean about the delete. So, what if we add an argument to `create-branch` that allows you to specify the remote? Then also add it to `git feature` (and all the other variants). Something like this:

```
git create-branch -r upstream mybranch
git create-branch --remote upstream mybranch
git feature -r upstream mybranch
git feature --remote upstream mybranch
```
It would default to origin if it is not specified.

I think that would make everyone happy.

spacewander:
Great!
One modification: For `git feature`, if no `-r` is given, we should avoid pushing to remote upstream. So that we could keep the behavior the same as before.

mapitman:
@spacewander I tested this code, but I'd appreciate it if someone else tested it too!

bin/git-create-branch
```diff
@@ -1,8 +1,35 @@
...
+for param in $*
+do
+	case $param in
+		-r|--remote)
+			if [ $# -eq 2 ]
+			then
+				BRANCH=$2
+			else			
+				REMOTE=$2
+				BRANCH=$3
```
spacewander: If we run `git-create-branch xx -r origin` :
The remote would be `-r`, while the branch would be `origin`.

bin/git-create-branch
```diff
@@ -1,8 +1,35 @@
...
+
+test -z $BRANCH && echo "branch required." 1>&2 && exit 1
+if [[ -n $REMOTE ]]
+then
+	git ls-remote --exit-code $REMOTE &>/dev/null
+	if [ $? = 0 ]
+	then
+		git push $REMOTE HEAD:refs/heads/$BRANCH
+		git fetch $REMOTE
+		git checkout --track -b $BRANCH $REMOTE/$BRANCH
```
spacewander: IMHO, if we could add `exit $?` after here, the code structure would be more concise.

bin/git-bug
```diff
@@ -1,14 +1,55 @@
...
+	else 
+		branch="$branch_prefix"/"${argv[0]}"
+	fi
+	if [[ -n $remote ]]
+	then
+		git create-branch -r $remote $branch
+	else
+		git create-branch $branch
+	fi
 fi
```
spacewander: Wow! Now `git bug|chore|feature|refactor` are almost the same!
Shall we create a new inner command, and extract the common logic into it?
Just like [git-line-summary](https://github.com/tj/git-extras/blob/master/bin/git-line-summary) for `git summary --line).
It would be helpful for the future maintaining.

mapitman: Yeah, I noticed that before I even started on this. bug chore and refactor were almost identical. I'll refactor them.

mapitman: I'm trying to come up for a name for the generalized version of the file. Any suggestions?

mapitman:
@spacewander Take a look and see what you think.

bin/git-create-branch
```diff
@@ -26,10 +34,10 @@ then
...
-	else
-		git checkout -b $BRANCH
+		exit $?
 	fi
-else
-	git checkout -b $BRANCH
 fi
 
+git checkout -b $BRANCH
+exit $?
```
spacewander: IMHO, the last `exit $?` could be eliminated. Because shell script will return with the exit code of the final command by default.

bin/git-create-branch
```diff
@@ -1,22 +1,30 @@
...
+			if [[ -n $2 ]]
 			then
-				BRANCH=$2
-			else			
 				REMOTE=$2
-				BRANCH=$3
+				shift
+			else			
+				REMOTE=origin
+				shift
```
spacewander: If I am right, the last argument would be shifted twice. Maybe we could remove it?

bin/git-bug
spacewander: @mapitman 
It's your privilege to name it and give the power. The only suggestion from me is that it could be started with a '_', to avoid invoking accidently. For instance, `git-_name_you_like`.

One more thing, remember to update the hard code there:
https://github.com/tj/git-extras/blob/master/check_integrity.sh#L67

mapitman: I started on these changes, but work, life, etc. are getting in the way of finishing. I'll try to push more commits sometime this week!

bin/git-create-branch
mapitman: fixed

bin/git-create-branch
mapitman: fixed

bin/git-bug
mapitman: Since the code already accepted an `--alias` argument, I kept the main script as git-feature and the others call it with the `-a` parameter

bin/git-bug
```diff
@@ -1,14 +1,4 @@
...
-else
-  test -z $1 && echo "bug <name> required." 1>&2 && exit 1
-  if test -n "$2"; then
-    echo "too many arguments; if not finishing a bug, only <name> of new bug is expected" && exit 1
-  fi
-  branch=bug/$1
-  git checkout -b $branch
-fi
+ALIAS=bug
+git feature -a $ALIAS $@
```
spacewander: A little suggestion:
Use `"$@"` instead of `$@`, to avoid re-splitting arguments with whitespaces.
Although we don't need it now, it could be helpful for future change.

mapitman: I'll update that in `git-chore` and `git-refactor` as well

mapitman: If you haven't guessed, I haven't done a lot of shell scripting, so thanks for bearing with me. I'm learning a lot.

mapitman:
Do we need to call out somewhere that the default behavior of `git create-branch` has changed since it no longer sets up a remote tracking branch in `origin`? Would this change require a bump of the major version since it is a breaking change?

mapitman:
Also, before merging this, would you like me to squash these commits into one?

mapitman:
I just noticed there are separate markdown files for generating the man pages. I'll update those too.

spacewander:
@mapitman 
Thank you for the updated documentation!

I think a bump of minor version is enough.
It would be great to write down the break change since `4.3.0` in the man page of `git create-branch`.

spacewander:
Merged. Thanks!

mapitman:
Woot!

PR #669, opened 2017-07-26
add git-wip and git-unwip

acant:
* **git-wip** will commit all the changes in the working directory into a WIP commit
* **git-unwip** will undo a WIP commit using git-undo

I have been using these commands for a while, to save interim changes when switching between feature branches. This ideas in other places, but it would be nice to included it here too. I have added links to files in https://github.com/ddollar/git-utils which are similar, but the scripts are not exactly the same.

Any feedback is welcome. I went through the contributing guidelines, but if I have missed anything I am happy to updated it.

Thanks in advance.


eMPee584:
Hm, kinda sounds like `git stash` to me.. Maybe you can check whether this is still useful today?

acant:
Rebased onto `main` instead of `master` to clean up the conflicts.

bin/git-unwip
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+# Based on scripts from git-utils:
+# * https://github.com/ddollar/git-utils/blob/master/git-unwip
+
+# Check if the last commit is a 'WIP' commit
+LAST_COMMIT=`git log -1 --pretty=%B | tr --delete '[:space:]'`
```
hyperupcall, line 7:
```suggestion
LAST_COMMIT=`git log -1 --pretty=%B | tr -d '[:space:]'`
```

[BSD tr](https://man.freebsd.org/cgi/man.cgi?tr) doesn't seem to support `--delete`

hyperupcall (changes requested):
Thanks! Generally looks good, had two suggestions

bin/git-unwip
```diff
@@ -0,0 +1,14 @@
...
+
+if [ 'WIP' != $LAST_COMMIT ]; then
+  echo 'Last commit is not a WIP commit, so it will not be unWIP-ed.'
+  exit
```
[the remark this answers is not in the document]
acant: That makes sense I will update the commit. :+1:

bin/git-unwip
acant: Well, that is unfortunate. I have been trying to use the full flag names in scripts, but you are right that BSD does not support it. :frowning_face: 

I will update to use `-d` as you suggested.

acant:
@hyperupcall  thanks for approving the workflow execution.

I have tried to fix the various failures in:
* `build`  by adding git-wip/unwip to the `man/git-extras.md` and re-generating with the make command.
* `typo`  by adding codespell inline ignore comments. (My last name is "Cant". :smile:)

* https://github.com/codespell-project/codespell?tab=readme-ov-file#ignoring-words
* https://github.com/codespell-project/codespell/issues/3396

acant:
@hyperupcall thanks again for running the workflow.
Looks like I successfully fixed the `build` stage, but not the `typo`.

After some investigation is seems that the problem is the version of codespell, which does not include the inline ignore support.

I have created a separate commit to upgrade codespell, which updates `tests/pyproject.tom` and `tests/poetry.lock`.
I also ran the codespell command locally and it works on my machine.

Hopefully it will work in Github Actions too. :crossed_fingers:

hyperupcall:
Thanks! Inline ignoring seems to work fine, although CI still unfortunately fails on spellcheck. I haven't properly been able to look into it because of my schedule at the moment - though I will later

acant:
@hyperupcall  I fixed one problem, can caused another. :laughing:

I just updated the `codespell` commit to include the spelling fix, and updated the commit message. I re-ran codespell locally, and it seemed to report no errors. Hopefully it will run green now.

Thanks again for your time and attention, it is much appreciated.

hyperupcall:
Fantastic! It's green now. Everything LGTM, I just have to pass it off to @spacewander now

hyperupcall (approved):

Commands.md
```diff
@@ -1628,3 +1630,21 @@ Commits changes with a generated message.
...
+Create a Work In Progress(WIP) commit, which will include all changes in the
+working directory. (i.e., changes to existing files, new files, removed files)
+
+```bash
+$ git wip
+```
+
+## git unwip
+
+Undo A Work In Progress(WIP) commit and put all of those changes back into the
```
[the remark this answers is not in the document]
spacewander: Better to also mention that the undo changes will be left in the staging area.

spacewander (commented):
Thanks! Just a few comments.

man/git-extras.1
```diff
@@ -1,6 +1,6 @@
-.\" generated with Ronn-NG/v0.9.1
-.\" http://github.com/apjanke/ronn-ng/tree/0.9.1
-.TH "GIT\-EXTRAS" "1" "September 2024" "" "Git Extras"
+.\" generated with Ronn-NG/v0.10.1
```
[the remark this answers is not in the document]
acant: Absolutely, I will update and push a new commit with just those changes.
Thanks for your feedback, and for taking the time to review this PR.

Commands.md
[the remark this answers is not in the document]
acant: Thanks, that is a good correction.

I looked over the documentation for `git-undo` and found the `--soft` flag which will put the changes back into the working directory and not into the stage. That is the behaviour that I actually want it to have.

I will update the script with `--soft` and leave the documentation as it is. That should make them match.

acant:
@spacewander I have updated the commit with `git undo --soft` and dropping the `man/git-extras.*` changes , and it is ready for review.
Thanks again for your time. And please let me know if there are any more changes to make.

man/git-extras.1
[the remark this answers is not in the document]
acant: @spacewander no worries, I was also just guessing at what was required when I added this.
But I also see that the CI systems know better than either of us, because it seems to be failing on a documentation check. :rofl: 
I will re-run the `make -C .. man/git-extras.{1,html}` command and add these files back in.

man/git-summary.1
```diff
@@ -1,6 +1,6 @@
-.\" generated with Ronn-NG/v0.9.1
```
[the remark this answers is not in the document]
acant: This was updated because of a spelling error which was noticed by the latest version of `codespell`.
I upgraded `codespell` so that I would be able to use the inline disabled comments.
I need this because my last name "Cant" gets flagged as a spelling error.

acant:
@spacewander when you are ready to re-run the workflow, I have updated the `git-extras.*` files added the `make -C .. man/git-extras.{1,html}` command back into the commit message.

Please let me know if there is anything which needs to be changed. Thanks again for your time.

(P.S.:  Would it be helpful to add `make -C .. man/git-extras.{1,html}` and instructions about updating `Commands.md`/`man/index.txt`/`man/git-extras.md` to the `man/Readme.md`? If yes, then I can open a separate PR.)

Commands.md
spacewander, line 1645:
```suggestion
Undo a Work In Progress(WIP) commit and put all of those changes back into the
```

spacewander (approved):

spacewander:
> (P.S.: Would it be helpful to add make -C .. man/git-extras.{1,html} and instructions about updating Commands.md/man/index.txt/man/git-extras.md to the man/Readme.md? If yes, then I can open a separate PR.)

Thanks. Let's submit a new PR for it.

Commands.md
acant: I have updated the commit with this change.

etc/git-extras-completion.zsh
```diff
@@ -441,3 +441,5 @@ zstyle ':completion:*:*:git:*' user-commands $existing_user_commands \
     undo:'remove latest commits' \
     unlock:'unlock a file excluded from version control' \
     utimes:'change files modification time to their last commit date'
```
spacewander, line 443: Missing a `\` at the end?

acant: Yes, you are correct. I missed that when I re-based.
I just updated the commit, and that is now fixed.

man/git-wip.html
```diff
@@ -0,0 +1,122 @@
...
+</code></pre>
+
+<p>Later on, undo the commit and continue making changes.</p>
+
+<pre><code>$ git unwip
+</code></pre>
+
+<h2 id="AUTHOR">AUTHOR</h2>
+
+<p>Written by Andrew Sullivan Cant &lt;<a href="mailto:mail@andrewsullivancant.ca" data-bare-link="true">mail@andrewsullivancant.ca</a>&gt; <!-- codespell:ignore Cant --></p>
```
spacewander, line 104: It seems the generated HTML files don't handle the `<!--` correctly. What about adding the word to https://github.com/tj/git-extras/blob/main/.github/.ignore_words?

acant: Hi @spacewander,

Sorry for the long delay. I finally got to have a look at the codespell ignore issue, and moved "Cant" into the `.ignore_words` file as you suggested. I re-ran the codespell command locally and re-generated the HTML, so hopefully this is all ready to go.

Thanks for your time and attention.

spacewander:
@acant 
Would you solve the conflict so we can rerun the CI?

spacewander (approved):

spacewander:
@acant 
Merged. Thanks!

PR #671, opened 2017-10-10
Improve ignore-io

sambostock:
This makes two improvements to `git ignore-io`:

- Fixes a bug where `git ignore-io -L` omitted entries which did not start with `a..z`
- Removes the arbitrary choice of 5 columns on `git ignore-io -L`, in favour of `column`'s default

### DIFF
--- a/bin/git-ignore-io
+++ b/bin/git-ignore-io
@@ -16,8 +16,7 @@ print_in_alphabetical_order() {
 }
 
 print_in_table_format() {
-    table_width=5
-    echo "$gi_list" | xargs -n $table_width | column -t
+    echo "$gi_list" | column
 }
 
 search() {
--- a/bin/git-ignore-io
+++ b/bin/git-ignore-io
@@ -8,11 +8,22 @@ update_gi_list() {
 }
 
 print_in_alphabetical_order() {
-    for i in {a..z};
+    first=true
+    for ignorable in $(echo "$gi_list" | sort);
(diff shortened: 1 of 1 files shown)


bin/git-ignore-io
```diff
@@ -8,16 +8,26 @@ update_gi_list() {
...
 print_in_alphabetical_order() {
-    for i in {a..z};
+    first=true
+    for ignorable in $(echo "$gi_list" | sort);
     do
-        echo "$gi_list" | grep "^$i" | tr "\n" " "
-        echo
+        first_character=${ignorable:0:1}
+        if [[ $first_character = $previous_first_character ]]; then
+            printf " $ignorable"
```
spacewander: Better to use `printf " %s" "$ignoreable"` to avoid format string injection.

bin/git-ignore-io
```diff
@@ -8,16 +8,26 @@ update_gi_list() {
 }
 
 print_in_alphabetical_order() {
-    for i in {a..z};
+    first=true
```
spacewander: IMHO, it is better to declare `local first ignorable first_character previous_first_character` before using those variables in function, to avoid polluting outer variable scope.

sambostock:
@spacewander Thanks for the tips! Updated with those in mind.

spacewander:
@sambostock 
Merged. Thanks!

PR #675, opened 2017-10-27
git-effort: restore normal (visible) cursor properly

quite:
If I understand GNU* correctly, cnorm should be used to undo `civis` (make
invisible). `cvvis` is just another abnormal mode, namely "very visible".

*) https://www.gnu.org/software/termutils/manual/termutils-2.0/html_chapter/tput_1.html


spacewander:
@quite 
`man terminfo` also support your opinion.
Merged. Thanks!

PR #676, opened 2017-11-03
Fix 'Invalid code page' error when restoring code page at end of script for Windows

gormac:
On Windows before setting the code page to 850 the install.cmd script stores the current code page. The current code page is however stored minus the last character, so a current code page '850' will be stored as '85'. This results in the 'Invalid code page' error when trying to restore the code page at the end of the script.


spacewander:
LGTM. Thanks!

PR #678, opened 2017-11-09
Docs> Add -- in front of cmd for git log

timhwang21:
The current docs are incorrect, using the command gives an error msg:

```
error: unknown argument --since=1 week ago
error: if that argument was meant for git-log,
error: please put it after two dashes ( -- ).
```


spacewander:
Merged. Thanks!

Issue #604, opened 2016-11-26
how to extend user-commands


vinurs:
I defined another `zstyle ':completion:*:*:git:*' user-commands` in another file, then source it in zshrc, but the second will replace the original one, it discard original user-commands, then i found this:
` zstyle ':completion:*:*:git:*' user-commands \`
`	   ${${(k)commands[(I)git-*]}#git-} \`
`         extra-example:'extra-example' `
but it has no descriptions.
how to achieve fuction like this: append to user-commands

nicolaiskogheim:
> then i found this:
zstyle ':completion:*:*:git:*' user-commands \
${${(k)commands[(I)git-*]}#git-} \
extra-example:'extra-example'
but it has no descriptions.

Where did you find this?

This sound like a question for Stack Overflow.

lhernanz:
Hi,

IMHO, overriding the user-commands value is a bad practice as it is going to cause confusion to the end user when using conflicting packages or local commands. I would suggest that you fix the use of the command with something like the following. I can create a pull request if you prefer.
`
zstyle -g existing_commands ':completion:*:*:git:*' user-commands
zstyle ':completion:*:*:git:*' user-commands $existing_commands flow:'provide high-level repository operations'
`
Regards

spacewander:
@lhernanz 
Your pr is welcome.

lhernanz:
Hi,
sorry for the delay. I have just added the PR.
Regards

spacewander:
The PR is merged. Reopen this issue if need.

PR #680, opened 2017-11-15
Preserve existing user commands in zsh completion

lhernanz:
Hi,

this PR addresses #604. Instead of completely overwriting the user-commands variable, we retrieve its value first and we add the additional completion mechanisms later. This is needed when the user has several packages that try to use this completion extension mechanism. With the current code, the last loaded package overwrites the rest.

Regards

### DIFF
--- a/etc/git-extras-completion.zsh
+++ b/etc/git-extras-completion.zsh
@@ -427,7 +427,9 @@ _git-undo(){
         '(--hard -h)'{--hard,-h}'[wipes your commit(s)]'
 }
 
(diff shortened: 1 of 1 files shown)


spacewander:
@lhernanz 
Could you cherry-pick your change and `push -f`, to avoid the trivial merge commit in the pull request?

lhernanz:
@spacewander 
Done. You should only see a commit now. I should have created a topic branch but the change was so small...

spacewander:
@lhernanz 
Merged. Thank you!

PR #690, opened 2018-01-23
feat(git-release): push tags first to help ci tools not miss them

dominicbarnes:
If you are buliding on CircleCI for example, pushing the commits first can lead to a race where CI env will not see the tags since the pull happened before the tags were finished pushing.

It should be fully safe (and backwards-compatible) to change the order of pushing to the remote. This should make integrating with CI tools less of a race.

### DIFF
--- a/bin/git-release
+++ b/bin/git-release
@@ -91,8 +91,8 @@ if test $# -gt 0; then
   git commit -a -m "$msg"
   # shellcheck disable=SC2086
   git tag $version -a -m "$msg" \
-    && git push $remote \
     && git push $remote --tags \
+    && git push $remote \
     && hook post-release $hook_args \
     && echo "... complete"
 else


bin/git-release
```diff
@@ -91,8 +91,8 @@ if test $# -gt 0; then
   git commit -a -m "$msg"
   # shellcheck disable=SC2086
   git tag $version -a -m "$msg" \
-    && git push $remote \
     && git push $remote --tags \
+    && git push $remote \
```
spacewander: IMHO, we could omit this line because `git push $remote --tags` already pushes the commit to remote branch.

dominicbarnes: I don't know if that's the case, I'll run a quick test.

dominicbarnes: I just did a test. The tag will be pushed, and you can find the commit if you reference it by SHA. However, the commit will **not** be pushed to the branch, so we still need both commands.

spacewander: @dominicbarnes 
My mistake. Push commit in this way doesn't push it to proper branch.

spacewander:
Merged. Thank you!

PR #691, opened 2018-02-04
Remove trailing dot of German Windows

koppor:
This fixed `install.cmd` for German Windows distributions.

When executing `CP`, the output is

    Aktive Codepage: 850.

Thus, the content of the variable `CP` is

    850.

When passing this to `chcp`, the output is

    Parameterformat falsch - 850.

Thus, the final lines of the script have no effect.

With this patch, the restore of the old code page works.


spacewander:
@koppor 
Merged. Thank you!

PR #697, opened 2018-02-07
git-release: create an empty release commit if there is nothing to commit

### LINKED ISSUE #696: git-release: does not work for me
Here is what it says:

```
$ git release 0.3.1
... releasing 0.3.1
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
```

No tag is generated, no changes are pushed. Any ideas on how to debug this?

Some more details:

```
$ git tag -l
0.1.0
0.1.1
0.2.0
0.3.0
$ git --version
git version 2.15.1
$ uname -a
Darwin ecvaawplun6.local 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro14,3 Darwin
```

spacewander:
May fix #696

### DIFF
--- a/bin/git-release
+++ b/bin/git-release
@@ -88,7 +88,7 @@ if test $# -gt 0; then
   if [ "$need_changelog" = true ]; then
     git-changelog -t "$version"
   fi
-  git commit -a -m "$msg"
+  git commit -a -m "$msg" --allow-empty
   # shellcheck disable=SC2086
   git tag $version -a -m "$msg" \
     && git push $remote --tags \
--- a/bin/git-release
+++ b/bin/git-release
@@ -35,6 +35,7 @@ if test $# -gt 0; then
     -r) remote=$2; shift ;;
     -m) msg=$2; shift ;;
     --semver) semver=$2; shift ;;
+    --no-empty-commit) no_empty_commit=true;;
     --) shift; hook_args="$hook_args $*"; break;;
     *) test -z "$version" && version=$1 ;;
     esac
@@ -88,7 +89,13 @@ if test $# -gt 0; then
   if [ "$need_changelog" = true ]; then
     git-changelog -t "$version"
   fi
-  git commit -a -m "$msg" --allow-empty
+
+  if [ "$no_empty_commit" = true ]; then
+    git commit -a -m "$msg" || true
+  else
+    git commit -a -m "$msg" --allow-empty
+  fi
+
   # shellcheck disable=SC2086
   git tag $version -a -m "$msg" \
     && git push $remote --tags \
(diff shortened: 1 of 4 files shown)


bin/git-release
```diff
@@ -88,7 +88,7 @@ if test $# -gt 0; then
   if [ "$need_changelog" = true ]; then
     git-changelog -t "$version"
   fi
-  git commit -a -m "$msg"
+  git commit -a -m "$msg" --allow-empty
```
perlun: I wonder if this really a reasonable fix to this problem. I mean, do we _really_ want an empty commit in these cases? Wouldn't it make more sense to just ignore committing?

spacewander: @perlun 
IMHO, if someone keeps using the `git release` to make new release, she will expect to see each release has its corresponding commit, with special commit message. If a commit could be ignored, this assumption could not work any more. Especially, when the user specifies her custom commit message with `git release`, she might consider it as a bug if the message is missing in `git log` because the committing was ignored.

perlun: That's a valid point. But still, it feels odd to have all-empty commits in a git tree.

My typical use case: I use `git release` for a shorthand for `git tag foo && git push origin foo`. Sometimes, there are modified files (like for Ruby gems where I typically bump the version in `version.rb`), but there are also cases when no files need to be modified just for the sake of making the release.

spacewander: Well, an empty commit is not quite odd.

1. If you use `merge --no-ff` to merge a fast-forward branch, you will get an empty merge commit at the top of git tree. Some teams require all merge operations should be done with `--no-ff` option. (I have been worked in one of these teams)
2. Some people suggest to use an empty commit as the initial commit of a git repo. For example, see https://bit-booster.com/doing-git-wrong/2017/01/02/git-init-empty/. At least, it proves that making an empty commit is not a taboo.

perlun: > If you use merge --no-ff to merge a fast-forward branch, you will get an empty merge commit at the top of git tree. Some teams require all merge operations should be done with --no-ff option. (I have been worked in one of these teams)

Yeah, I understand - I personally hate these commits, but that's a different story. 😉 I know that this is the policy at certain places, and I respect that people have different opinions on the matter.

> Some people suggest to use an empty commit as the initial commit of a git repo. For example, see https://bit-booster.com/doing-git-wrong/2017/01/02/git-init-empty/.

That's an interesting point actually - it makes a certain bit of sense (haven't read the link.)

spacewander: @perlun 
What about adding an extra option called `--no-empty-commit`? So that you could use this option to avoid creating empty commit.

spacewander:
@perlun 
The `--no-empty-commit` option is added in my latest commit.

bin/git-release
perlun: Makes sense, thanks for adding!

PR #703, opened 2018-03-01
Installation.md: Zsh-plugin method which doesn't require root access

psprint:
A short section to `Installation.md`. It describes how to install as Zsh plugin. Benefits of such method:

 - no administrator access required (easy install on e.g. digital ocean, etc.),
 - declarative method of installing – 2 lines in .zshrc lead to git-extras on each account.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -112,6 +112,18 @@ One-liner:
 curl -sSL http://git.io/git-extras-setup | sudo bash /dev/stdin
 ```
 
+## Installing as Zsh plugin
+
+[Zplugin](https://github.com/zdharma/zplugin) can install git-extras by using:
+
+```zsh
+zplugin ice as"program" pick"$ZPFX/bin/git-*" make"PREFIX=$ZPFX"
+zplugin light tj/git-extras
+```
+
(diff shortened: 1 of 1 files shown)


spacewander:
@psprint 
So the `zplugin` will take over the upgrade of `git-extras`?
Should we also apply a patch as what we have done with Homebrew?
https://github.com/tj/git-extras/blob/master/brew-release.patch
https://github.com/tj/git-extras/blob/master/.travis.yml#L9

spacewander:
Now I think we should have the same patch instead of `xxx-release.patch`. We could use some unix command to update the patch according to the package manager specified via Makefile arguments.

psprint:
Yes Zplugin will update. [Here](https://github.com/zdharma/zplugin#hint-extending-git) is more detailed description. I've tested `git-extras update` and it works normally, because files are in `$HOME`, writable. Issuing `zplugin update tj/git-extras` will update too, not to latest tag but to HEAD, but that's not a problem. So I think the patch isn't needed.

spacewander:
@psprint 
Good news to know. Already merged, thanks!

Issue #705, opened 2018-03-23
git-delete-branch not working for branches without upstream


MontakOleg:
Hello!

# Problem:

```
git init
touch readme
git add readme
git commit

# create branch with upstream
git checkout -b test --track
git checkout master

# creare branch without upstream
git checkout -b test2
git checkout master

git branch
->
  master
  test
  test2

# this successfully delete local branch test
git delete-branch test

# this quietly do nothing <---- PROBLEM
git delete-branch test2

```

# Reason:

`set -e` command added in 9744830d4aa48b45b743cf7f77bfffb559c153dc

# Explanation:

Command
```
git config branch.$branch.remote
```
exited with non-zero code if branch has no remote. In example above:

```
git config branch.test.remote
echo $?
->
 1   # <------- 1 exit code
```

# Solution:

Are there any reasons for `set -e` ? I'm thinking of deleting this line.

spacewander:
Closed as #706 is merged.

PR #706, opened 2018-03-23
Fix deleting branches without upstream

MontakOleg:
See https://github.com/tj/git-extras/issues/705 for details

### DIFF
--- a/bin/git-delete-branch
+++ b/bin/git-delete-branch
@@ -6,14 +6,12 @@ test -z $1 && echo "branch required." 1>&2 && exit 1
 
 for branch in "$@"
 do
-  remote=$(git config branch.$branch.remote)
-  test -z $remote && remote="origin"
-  ref=$(git config branch.$branch.merge)
-  test -z $ref && ref="refs/heads/$branch"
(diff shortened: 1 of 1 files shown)


spacewander:
@MontakOleg 
Maybe we could use `git config branch.$branch.remote || echo 'origin'` and `git branch -D ... || continue` to achieve the same result, without removing the `set -e`?

MontakOleg:
I will try

MontakOleg:
`git branch -D ... || continue` will skip deleting remote branch, if local branch does not exists. This is expected behavior?

And what we want to do if `git branch -d -r $remote/$branch` fails?


I would prefer to try delete remote branch even if there no local branch.

spacewander:
@MontakOleg 
Well, we could use `git branch -D ... || true` instead?
In my opinion, we could use  `|| continue` with `git branch -d -r $remote/$branch`. If we failed to delete the remote tracking branch locally, we don't need to push the change to the remote.

MontakOleg:
@spacewander
Look at this, please.

spacewander:
@MontakOleg 
Merged. Thank you!

PR #709, opened 2018-04-20
Improve the git-sed command

pabs3:
Discover a separator when the / character is used in arguments.
Prevent running sed when the search pattern does not match any files.
Allow the flags to be passed as a third argument.

### DIFF
--- a/bin/git-sed
+++ b/bin/git-sed
@@ -66,7 +66,22 @@ actual command:
     shift
 done
 
+all="$search$replacement$flags"
+case "$all" in
+    */*)
+        ascii="$(for((i=32;i<=127;i++)) do printf '%b' "\\$(printf '%03o' "$i")"; done)"
+        sep="$(printf '%s' "$ascii" | tr -d "$all")"
+        sep="$(printf %.1s "$sep")"
+        if [ "X$sep" = "X" ] ; then
+            echo 'could not find an unused character for sed separator character'
+            exit 1
+        fi
+        ;;
+    *)
+        sep=/
+        ;;
+esac
 r=$(xargs -r false < /dev/null > /dev/null 2>&1 && echo r)
-command="git grep -lz '$search' | xargs -0$r sed -i 's/$search/$replacement/$flags'"
-git grep -lz "$search" | xargs -0"$r" sed -i "s/$search/$replacement/$flags"
+command="git grep -lz '$search' | xargs -0$r sed -i 's$sep$search$sep$replacement$sep$flags'"
+git grep -lz "$search" | xargs -0"$r" sed -i "s$sep$search$sep$replacement$sep$flags"
 do_commit
--- a/bin/git-sed
+++ b/bin/git-sed
@@ -2,10 +2,10 @@
 
 usage() {
     cat <<EOF
-usage: git sed [ -c ] [ -f <flags> ] <search> <replacement>
+usage: git sed [ -c ] [ -f <flags> ] <search> <replacement> [ <flags> ]
 
 Run git grep and then send results to sed for replacement with the
-given flags, if -f is provided.
+given flags, if they are provided via -f or as the third argument.
 
 Also runs git commit if -c is provided.
 EOF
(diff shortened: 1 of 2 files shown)


spacewander:
A little question:
What's the reason to allow the flag to be passed as the third option?

spacewander:
@pabs3 
I see.

bin/git-sed
```diff
@@ -64,6 +66,21 @@ actual command:
...
+        if [ "X$sep" = "X" ] ; then
+            echo 'could not find an unused character for sed separator character'
+            exit 1
+        fi
+        ;;
+    *)
+        sep=/
+        ;;
+esac
+command="git grep -lz '$search' | xargs -0r sed -i 's$sep$search$sep$replacement$sep$flags'"
```
spacewander: `-r` option is a GNU extension. The `xargs` in OS X doesn't support it.
We have to remove it for the compability.

man/git-sed.md
```diff
@@ -3,12 +3,12 @@ git-sed(1) -- replace patterns in git-controlled files
 
 ## SYNOPSIS
 
-`git-sed` [ -c ] [ -f <flags> ] <search> <replacement>
+`git-sed` [ -c ] [ -f <flags> ] <search> <replacement> [ <flags> ]
```
spacewander: Also need to update the html and man page.
See https://github.com/tj/git-extras/blob/master/man/Readme.md

pabs3:
I've updated my branch to add a mechanism to check if the xargs -r
option exists and only use it when it does exist.

For the manual pages:

 * Personally I don't think it is appropriate to include automatically
   generated files in a git repository containing the source. I would
   put generated files into either the website git or into packages for
   end-users but not into the source git repository.
 * There are two different build mechanisms that produce two different
   results for manual pages. Both of them produce different results
   than the files that are already in the git repository.
 * Most of the files need updating at minimum for the dates.
 * The ronn tool does not seem to perform deterministic builds,      
   it seems to insert random characters into email addresses.          
   https://reproducible-builds.org/

So I prefer to do the documentation changes in another pull request.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

spacewander:
> I've updated my branch to add a mechanism to check if the xargs -r option exists and only use it when it does exist. 

Does it cause different behaviors between different platforms?
Maybe we'd better check the output of `git grep` by ourselves?

pabs3:
Yes, it causes the sed error to be hidden on GNU and shown on macOS.

I also don't think it is useful to workaround macOS xargs missing -r
since it would require running `git grep` twice instead of once and
the consequences of -r being missing are very much minor.
 
If macOS folks want the GNU/Linux behaviour they can install GNU xargs.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

spacewander:
> running `git grep` twice instead of once

Maybe we could store the result in a bash variable?

pabs3:
The results of `git grep -z` include NUL bytes and as far as I can tell
NUL bytes cannot be stored in bash variables, they are ignored:

   $ foo="$(printf '%s\0' foo)" ; printf '%s' "$foo" | hexdump 
   bash: warning: command substitution: ignored null byte in input
   00000000  66 6f 6f                                          |foo|
   00000003

spacewander:
@pabs3 
OK.
Where is your latest change? I could not find it in this pull request.

pabs3:
I sent it to the git-sed branch on my git-extras fork, but I guess
github doesn't know how to associate that with the pull request.

pabs3:
Ah, I pushed to the wrong branch on my fork, fixed now.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

bin/git-sed
```diff
@@ -64,6 +66,22 @@ actual command:
...
+        if [ "X$sep" = "X" ] ; then
+            echo 'could not find an unused character for sed separator character'
+            exit 1
+        fi
+        ;;
+    *)
+        sep=/
+        ;;
+esac
+r=$(xargs -0r false < /dev/null > /dev/null 2>&1 && echo r)
```
spacewander: Maybe we could use `-r` instead of `-0r` here, since we only want to test for `-r`?

spacewander:
@pabs3 
Merged. Thank you!

spacewander:
@pabs3 
Will you submit another pull request to update the documentation?

PR #713, opened 2018-05-05
Updates to the documentation building workflow and docs

pabs3:
* Remove one of the documentation building tools
* Rebuild all generated documentation files


spacewander:
@pabs3 
Err... I think just update the documentation of `git-sed` is enough. Each documentation should have its separate update. It would a nightmare of version control if we update the timestamp of all the documentation after one of them changed.

pabs3:
The dates aren't being changed to the current date so I got the
impression the dates were changing for all files because someone had
forgotten to update the generated files after modifying the source
files. This is why I like to remove generated files from VCS and
tarballs so the build process is well tested and run often. 

Almost every file is missing the "Git Extras" header that the Makefile
adds by default, but the manning-up.sh script does not, so they all
need to be updated anyway.

I can do one commit per file if you prefer that to one big update.

I can drop ronn's email modification weirdness if you want.

spacewander:
@pabs3 
I see.

BTW, after I run the `make -C .. docs` on your branch, `git-extras.html` is changed:
```
$ git diff
diff --git a/man/git-extras.html b/man/git-extras.html
index 9f7b700..c31fb5b 100644
--- a/man/git-extras.html
+++ b/man/git-extras.html
@@ -164,7 +164,7 @@
 
 <h2 id="AUTHOR">AUTHOR</h2>
 
-<p>Written by Tj Holowaychuk &lt;<a href="&#109;&#97;&#x69;&#x6c;&#x74;&#x6f;&#58;&#x74;&#106;&#64;&#x76;&#x69;&#115;&#x69;&#x6f;&#x6e;&#45;&#109;&#x65;&#100;&#x69;&#x61;&#46;&#99;&#x61;" data-bare-link="true">&#x74;&#106;&#x40;&#x76;&#105;&#115;&#x69;&#111;&#x6e;&#45;&#x6d;&#101;&#x64;&#105;&#97;&#x2e;&#x63;&#97;</a>&gt;</p>
+<p>Written by Tj Holowaychuk &lt;<a href="&#109;&#97;&#105;&#x6c;&#116;&#111;&#x3a;&#x74;&#106;&#64;&#x76;&#105;&#x73;&#105;&#111;&#x6e;&#x2d;&#109;&#x65;&#x64;&#x69;&#x61;&#46;&#99;&#97;" data-bare-link="true">&#x74;&#106;&#64;&#x76;&#x69;&#115;&#x69;&#x6f;&#110;&#45;&#x6d;&#101;&#x64;&#105;&#x61;&#x2e;&#x63;&#x61;</a>&gt;</p>
 
 <h2 id="REPORTING-BUGS">REPORTING BUGS</h2>
```
Could you delete the html files and rebuild them?

And I hope you could squash your changes into three commits:
1. Change to documentation relative tools
2. Change to documentation (without `git-sed`)
3. Change to documentation of `git-sed`

pabs3:
The changes to the email addresses in the HTML are because ronn inserts random characters to prevent spammers from being able to harvest them. 

I've made some changes and pushed to my docs branch.

I split out the rebuild of the documentation into several commits:

* git-sed flags argument
* git-clear whitespace removal
* git-extras sorting change
* index.txt sorting and whitespace change
* headers and dates, I combined these because the manual pages include both on the same line.

I've removed the HTML email address randomisation from my commits, since it is just commit diff bloat for no useful reason.

I've added $(MANS) to the dependencies of the index.txt make target since it references all of the manual pages, including git-extras(1).

I've switched to $(MAN_BINS) for the name of the variable that is a dependency of the git-extras.md make target. That seems more self-explanatory since it is about manual pages for the $(BINS) files.

I've moved the $(MAN_BINS) calculation next to $(MANS).

I've added index.txt to the dependencies of the HTML make target.

man/Readme.md
```diff
@@ -7,12 +7,12 @@ To generate documentation:
...
 
-2) Then use a program ronn. [Get ronn from github.](https://github.com/rtomayko/ronn)
+2) Then install a program ronn. [Get ronn from github.](https://github.com/rtomayko/ronn)
 
-3) Run ronn:
+3) Run make:
 
 ```
-$ ronn <filename>.md
+$ make -C .. docs
```
spacewander: @pabs3 
Since the build result is not reproducible, should we use `make man/git-<command>.{1,htmll}` instead of fully updating?

pabs3:
`make man/git-<command>.{1,html}` will not update index.txt and
git-extras.md so I prefer docs. Alternatively we could add an explicit
command for updating index.txt and git-extras.md.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

spacewander:
@pabs3 
`man/%.html` depends on `man/index.txt`, and `man/index.txt` depends on all `*.md` including the markdown page user just changed. If I am correct, `man/index.txt` would be triggered to update itself.

spacewander:
@pabs3 
Merged. Thank you for your contributions!

PR #715, opened 2018-05-07
Fix bugs with git-release --semver

tjaartvdwalt:
I found 2 bugs when using `git-release --semver`

1. The version regex greedily matches only the first digit for version numbers that contain multiple digits. Matching the multi digit case first solves this problem. 

2. When doing a `major` release, the `minor` and `patch` version numbers should be reset to `0`, and when doing a `minor` release, the `patch` number should be reset to `0`.


spacewander:
@tjaartvdwalt 
Merged. Thanks!

PR #716, opened 2018-05-20
Add installation guide for Arch Linux

francoism90:
Link to packages, since Arch Linux users may use an AUR helper.


spacewander:
@francoism90 
Merged. Thank you!

PR #717, opened 2018-05-23
add: git-ignore -p (private to repo)

flatcap:
When working in a repo, it's common to create temporary files -- files that are specific to what you're doing at that very time.

Git has a private ignore file `.git/info/exclude` which isn't replicated.
This is a perfect place to ignore those one-off temporary files.

I've added the option to `git-ignore`, updated the docs and the completion scripts.

I hope you find this useful.

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -35,6 +35,16 @@ function add_local {
   add_patterns .gitignore "$@"
 }
 
+function show_private {
+  cd "$(git root)"
+  show_contents Private .git/info/exclude
+}
+
+function add_private {
+  cd "$(git root)"
+  add_patterns .git/info/exclude "$@"
+}
+
 function add_patterns {
   echo "Adding pattern(s) to: $1"
   local file="${1/#~/$HOME}"
@@ -48,6 +58,8 @@ if test $# -eq 0; then
    show_global
    echo "---------------------------------"
    show_local
+   echo "---------------------------------"
+   show_private
 else
   case "$1" in
     -l|--local)
@@ -57,6 +69,10 @@ else
     -g|--global)
       test $# -gt 1 && add_global "${@:2}" && echo
       show_global
+      ;;
+    -p|--private)
+      test $# -gt 1 && add_private "${@:2}" && echo
+      show_private
       ;;
     *)
       add_local "$@"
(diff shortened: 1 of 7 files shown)


bin/git-ignore
```diff
@@ -35,6 +35,16 @@ function add_local {
...
 }
 
+function show_private {
+  cd "$(git root)"
+  show_contents Private .git/info/exclude
+}
+
+function add_private {
+  cd "$(git root)"
+  add_patterns .git/info/exclude "$@"
```
spacewander: @flatcap 
I think you need to create `.git/info` first if the directory doesn't exist.

flatcap: Hmm... I never would have guessed that the `info` dir was optional!
I've added a `mkdir -p` to be sure.

spacewander: @flatcap 
Better to check if the directory is existed before running the `mkdir` command?

flatcap: We _could_, but `mkdir --parents` will do that for us.
It quietly succeeds if the dir exists.

spacewander: @flatcap 
But `mkdir` only in need is serveral times faster than running `mkdir -p`.
Just wrote a simple benchmark:
```bash
#!/usr/bin/env bash

for (( i = 0; i < 1000; i++ ))
do
    # ./mkdir_p.sh  1.16s user 0.72s system 103% cpu 1.823 total
    mkdir -p .git/info
done
```

```bash
#!/usr/bin/env bash

for (( i = 0; i < 1000; i++ ))
do
    # ./mkdir_if_need.sh  0.03s user 0.00s system 98% cpu 0.031 total
    if [ ! -d ".git/info" ]
    then
        mkdir .git/info
    fi
done
```

Maybe it is trivial to care about performance in shell script. But I write C and Lua in my daily job, so occupation disease, you know.

flatcap: > Maybe it is trivial to care about performance in shell script

hehe, not at all, but you have to pick your battles.
In this case, there's no loop so the time saving is too small to be noticed (by the human running it)
However, it **is** more efficient to `test` first, so I've updated the PR :-)

spacewander:
@flatcap 
Merged. Thank you!

flatcap:
Thanks @spacewander

PR #743, opened 2019-02-02
git-pr supports pulling multiple pr at once

spacewander:
Via `git pr upstream:12345 upstream:12678 upstream:52345`
CC @ZhilkinSerg


hemanth:
RAD!

PR #751, opened 2019-04-26
git-rename-remote: rename a remote regardless of any existing remotes

timfeirg:
I'll add docs and any other relevent stuff if this function is approved

### DIFF
--- a/bin/git-rename-remote
+++ b/bin/git-rename-remote
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+old=$1
+new=$2
+
+test -z $old && echo "old remote name required." 1>&2 && exit 1
+test -z $new && echo "new remote name required." 1>&2 && exit 1
+
+if ! git config --get "remote.$old.fetch" > /dev/null; then
+  echo "remote $old doesn't exist"
+  exit 1
+fi
+
+if git config --get "remote.$new.fetch" > /dev/null; then
+  git remote remove $new
+fi
+git remote rename $old $new
+git remote -v
--- a/Commands.md
+++ b/Commands.md
@@ -48,6 +48,7 @@
  - [`git release`](#git-release)
  - [`git rename-branch`](#git-rename-branch)
  - [`git rename-tag`](#git-rename-tag)
+ - [`git rename-remote`](#git-rename-remote)
  - [`git repl`](#git-repl)
  - [`git reset-file`](#git-reset-file)
  - [`git root`](#git-root)
@@ -509,6 +510,19 @@ $ git tag
 test2
 ```
 
+## git rename-remote
+
+Rename a git remote regardless of name conflict, and then list current git remotes.
+```
+$ git remote -v
+origin	git@myserver.com:myuser/foo.git (fetch)
+origin	git@myserver.com:myuser/foo.git (push)
+upstream	git@myserver.com:myuser/bar.git (fetch)
+upstream	git@myserver.com:myuser/bar.git (push)
+$ git-rename-remote upstream origin
+origin	git@myserver.com:myuser/bar.git (fetch)
+origin	git@myserver.com:myuser/bar.git (push)
+```
 
 ## git reauthor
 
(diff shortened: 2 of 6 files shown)


bin/git-mv-remote
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+old=$1
+new=$2
+
+test -z $old && echo "old remote name required." 1>&2 && exit 1
+test -z $new && echo "new remote name required." 1>&2 && exit 1
+
+git remote remove $new || true
```
spacewander: Better to check if the `$new` remote exists before removing it, so that we won't see error message like `fatal: ....`.

man/git-mv-remote.md
```diff
@@ -0,0 +1,44 @@
+git-mv-remote(1) -- Rename a remote
```
spacewander: What about changing the name to `git-rename-remote`? Since we already have `git-rename-branch` and `git-rename-tag`.

timfeirg: done

bin/git-mv-remote
timfeirg: I just suppressed its output because I was feeling lazy, hope that's ok

spacewander: @timfeirg 
Hiding the error message is not a good idea...
Maybe you could use `git config --get 'remote.$name.fetch'` to check if a remote exists.

timfeirg: fixed as suggested

bin/git-rename-remote
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+old=$1
+new=$2
+
+test -z $old && echo "old remote name required." 1>&2 && exit 1
+test -z $new && echo "new remote name required." 1>&2 && exit 1
+
+if git config --get "remote.$new.fetch"; then
```
spacewander: To avoid the confusing and useless output like `+refs/heads/*:refs/remotes/origin/*`, we should redirect the stdout to `/dev/null`.

bin/git-rename-remote
```diff
@@ -0,0 +1,13 @@
...
+  git remote remove $new
+fi
+git remote rename $old $new
+git remote -v
```
spacewander: Maybe we should skip the `git remote -v` when `git remote rename` failed?

timfeirg: such good advice, I've `set -e` this script.

bin/git-rename-remote
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail
+
+old=$1
+new=$2
+
+test -z $old && echo "old remote name required." 1>&2 && exit 1
+test -z $new && echo "new remote name required." 1>&2 && exit 1
+
+if git config --get "remote.$new.fetch" > /dev/null; then
+  git remote remove $new
```
spacewander: One little thing: maybe we should also check if the `old` remote exists?
Consider the situation below:

0. There are two remotes: `own` and `upstream`.
1. Alice wants to rename `own` to `upstream`.
2. She run `git rename oen upstream`.
3. Because the typo, the operation fails.
4. However, `upstream` is removed by the this line: `git remote remove $new`.

man/git-extras.md
```diff
@@ -1,100 +0,0 @@
-git-extras(1) -- Awesome GIT utilities
```
spacewander: Why remove this file?

timfeirg:
I've built the docs but check_integrity.sh still won't pass, and I can't understand its output:

```
$ ./check_integrity.sh $(find bin | cut -b 5- | xargs)
Skip command rscp
Skip command line-summary
create man/git-extras.md
The command "./check_integrity.sh $(find bin | cut -b 5- | xargs)" exited with 1.
```

man/git-extras.md
timfeirg: I didn't figure out how `make docs` work so I messed with it a bit.
I believe I've fixed it now.

spacewander:
Great! I will merge it later.

timfeirg <notifications@github.com> 于2019年5月8日周三 下午2:09写道：

spacewander:
@timfeirg 
Thank you!

PR #755, opened 2019-06-02
Improved the 'git-info' command

bbugyi200:
### Added:
* colored output to git-info (can be disabled with the `--no-color` option)
* new `-c <N>` option which allows the user to specify the number of commits to show in the output

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -731,7 +731,7 @@ $ git info
 
 ```
 
-If you wish to omit the config section, you may use `--no-config`:
+If you wish to omit the config section, you may use `--no-config` (or `-n`):
 
 ```bash
 $ git info --no-config
--- a/Commands.md
+++ b/Commands.md
@@ -737,6 +737,12 @@ If you wish to omit the config section, you may use `--no-config` (or `-n`):
 $ git info --no-config
 ```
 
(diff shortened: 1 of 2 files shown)


bin/git-info
```diff
@@ -1,47 +1,64 @@
...
 
-echon "## Remote Branches:"
-echon "$(remote_branches)"
+header "Local Branches:"
+git_cmd "git branch ${color}"
 
-echon "## Local Branches:"
-echon "$(local_branches)"
+if [[ "${max_count}" -ne 1 ]]; then
+    header="Top ${max_count} Commits:"
```
spacewander: IMHO, `Top N Commits:` is vague. Actually we show the last N commits in the current branch.

Commands.md
```diff
@@ -737,6 +737,26 @@ If you wish to omit the config section, you may use `--no-config`:
 $ git info --no-config
 ```
 
+By default, the output is colored to make it easier on the eyes. When this is not desired (if piping the output of `git info` to another command, for example), the `--no-color` option may be used:
+
+```bash
+$ git info --no-color
+```
+
+You can select the number of previous commits to show by using the `-c <N>` option (where `<N>` is any integer). It is important to note, however, that `-c` **must be the FIRST option** you pass to `git info`. For example:
```
spacewander: We can use a while loop to get rid of this limitation, like: https://github.com/tj/git-extras/blob/master/bin/git-release#L31

PR #756, opened 2019-06-02
Move documentation to Wiki

haydennyyy:
I added pages to the wiki for Installation and Commands, so they can safely be removed from the folder.

I just need to sort out the links in the README, but that should be fine.


spacewander:
Is there a way to check if all commands are listed in the Commands wiki?
Currently we check it via `grep` the `Commands.md`: https://github.com/tj/git-extras/blob/master/check_integrity.sh#L32

spacewander:
I thought about it tonight. Currently we can track all the changes (including the documentation changes) via git, without any platform locking. It's good enough. If we put the documentation in the wiki, we can't review the change like every pull request.

I will suspend this pull request until there are more evidences suggest that putting the documentation in the wiki is a better idea.

PR #761, opened 2019-06-15
`create-branch` allows for preference for `remote` option

btmurrell:


bin/git-create-branch
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+# preference takes lowest priority; look for remote from prefs first
+REMOTE_PREF=`git config git-extras.create-branch.remote`
+if [ "$REMOTE_PREF" == "true" ]; then
+	REMOTE=origin
```
spacewander: A little question: Why we need to use `git config git-extras.create-branch.remote true` to indicate the `origin` remote? IMHO, using `git config git-extras.create-branch.remote origin` is enough, which is simpler.

man/git-create-branch.md
```diff
@@ -7,17 +7,27 @@ git-create-branch(1) -- Create branches
...
+
+## PREFERENCES
+
+You may save preferences for the `remote` option above by using `git config` with the key `git-extras.create-branch.remote` whose value works like the corresponding option. A value of `true` will use the default remote `origin`. Any other value will represent a remote name.
+
+    $ git config git-extras.create-branch.remote true
+
+    $ git config git-extras.create-branch.remote upstream
+
+With this preference set, `git-create-branch` will behave as if the `-r|--remote` option were given.
```
spacewander: We should mention that `-r|--remote` will override the preference.

man/git-create-branch.md
```diff
@@ -29,11 +39,12 @@ git-create-branch(1) -- Create branches
 
 ## NOTES
 
-As of 4.4.0, the default behavior has changed. `git-create-branch` will no longer automatically setup a remote tracking branch unless the `-r|-remote` option is specified.
+As of 4.4.0, the default behavior has changed. `git-create-branch` will no longer automatically setup a remote tracking branch unless the `-r|-remote` option is specified, however this behavior may be obtained by `remote` preference as described in [Preferences](#PREFERENCES) section.
```
spacewander: Need to point out that the preference support is added in `4.8.0-dev` (we will update the version number when we create the next release).

bin/git-create-branch
```diff
@@ -27,13 +35,18 @@ fi
 test -z $BRANCH && echo "branch required." 1>&2 && exit 1
 if [[ -n $REMOTE ]]
 then
-	git ls-remote --exit-code $REMOTE &>/dev/null
+	git ls-remote --exit-code $REMOTE 1>/dev/null 2>/tmp/ls-remote-error
```
spacewander: The `ls-remote-error` file will be created even there is not output to the stderr. We should remove it in all the paths.

bin/git-create-branch
```diff
@@ -27,13 +35,18 @@ fi
...
 	if [ $? = 0 ]
 	then
 		git push $REMOTE HEAD:refs/heads/$BRANCH
 		git fetch $REMOTE
 		git checkout --track -b $BRANCH $REMOTE/$BRANCH
 		exit $?
+	else
+		echo
+		echo "    Error connecting to remote '$REMOTE': `cat /tmp/ls-remote-error`"
```
spacewander: Better to use $(< /tmp/ls-remote-error) instead. It can save a `cat`.

bin/git-create-branch
btmurrell: @spacewander , fair point. My intention was to make the pref in direct parity with the argument of the command.  Presence of the `[-r|--remote]` option with no argument acts as a boolean. Perhaps parity adds awkwardness in this case. I'm happy to take your suggestion.  Knowing my intent, let me know what your thought is on that.

man/git-create-branch.md
btmurrell: :+1:

man/git-create-branch.md
btmurrell: Thanks. I was not sure how to reference version.

man/git-create-branch.md
```diff
@@ -7,17 +7,27 @@ git-create-branch(1) -- Create branches
...
+
+## PREFERENCES
+
+You may save preferences for the `remote` option above by using `git config` with the key `git-extras.create-branch.remote` whose value works like the corresponding option. A value of `true` will use the default remote `origin`. Any other value will represent a remote name.
+
+    $ git config git-extras.create-branch.remote true
+
+    $ git config git-extras.create-branch.remote upstream
+
+With this preference set, `git-create-branch` will behave as if the `-r|--remote` option were given. The command line option `-r|--remote` will override this preference.
```
spacewander: Need an example of the overriding, for example, run `git create-branch xxx -r` will use `origin` as the remote. And also another example to show that with the preference, the `git create-branch xxx` will work like `git create-branch xxx -r preference`.

bin/git-create-branch
spacewander: @btmurrell 
IMHO, if people want to set their preference, they will not want to use the default remote defined by the tool author. So I think we don't need to preserve a default option for the preference. What about your idea?

bin/git-create-branch
```diff
@@ -18,7 +26,7 @@ do
 	shift
 done
 
-if [[ -z $BRANCH ]] && [[ -n $REMOTE ]] 
+if [[ -z $BRANCH ]] && [[ -n $REMOTE ]]
```
spacewander: I just found an edge case. If the user runs `git create-branch` by mistake. he/she will get the preference remote created as the branch name.

bin/git-create-branch
btmurrell: you're right. that's a more common sense way to use it.

spacewander:
@btmurrell 
Will you address the two remain issues - fixing the edge case and adding two more example? Thanks!

man/git-create-branch.md
btmurrell: thanks; yes, needed.

bin/git-create-branch
btmurrell: great catch. now fixed

bin/git-create-branch
```diff
@@ -18,23 +26,29 @@ do
...
 		git push $REMOTE HEAD:refs/heads/$BRANCH
 		git fetch $REMOTE
 		git checkout --track -b $BRANCH $REMOTE/$BRANCH
 		exit $?
+	else
+		echo
+		echo "    Error connecting to remote '$REMOTE': $(</tmp/ls-remote-error)"
+		echo
 	fi
+	rm -f /tmp/ls-remote-error
```
spacewander: We need to preserve the exit code of `git checkout` or `git ls-remote`, and `exit $rc` here. Otherwise, the `$? = 0` branch will not remove the temp file and `$? != 0` branch will not exit correctly.

btmurrell: good catch! thanks for the reviews. handled now.

btmurrell:
@spacewander -- now that i *_think_* this PR is complete, i just noticed in `git-changelog`, this command is using `git config changelog.OPTION` for preferences.  that is, there is a precedent for preferences in the git-extras commands... and it does not prefix with `git-extras`.  perhaps the `git-extras` namespacing is superfluous?  i think continuity among all commands would be best.... i'm happy to remove the namespace.

spacewander:
@btmurrell 
Thanks for your report. I think adding `git-extras` prefix is better. But for backward compatibility, if nobody complains, we won't change it.

bin/git-create-branch
```diff
@@ -18,23 +26,34 @@ do
...
+	else
+		echo
+		echo "    Error connecting to remote '$REMOTE': $REMOTE_ERROR"
+		echo
+		exit $REMOTE_EXIT
 	fi
 fi
 
 git checkout -b $BRANCH
+exit $?
```
spacewander: Is this line useful? The shell script always exits with the exit code of the last command.

btmurrell:
ok, @spacewander.  i will squash now.

bin/git-create-branch
btmurrell: :blush:

spacewander:
@btmurrell 
Will you squash the commit so that I can merge it? Thanks!

btmurrell:
Sorry @spacewander -- i tried to squash earlier, but did the push wrong. did it this time, but got it to two commits. botched, but hopefully clean enough.

spacewander:
@btmurrell 
Merged. Thank you!

PR #768, opened 2019-07-18
git-guilt: protect against encoding issues with LC_ALL=C

bric3:
This does the same fix as tj/git-extras#460.


spacewander:
@bric3 
Merged. Thank you!

bric3:
You're welcome thanks for maintaining the project !

PR #772, opened 2019-08-17
Add git coauthor command

jacobherrington:
This adds a feature to add additional authors to the previous commit,
this feature is supported by GitHub and GitLab and based on the format
should be self-evident to anyone reading the commit message.

This is my first PR to this repo, let me know if I've done something wrong or need to improve the quality of this PR for any reason. Thanks! 🤠

### DIFF
--- a/bin/git-coauthor
+++ b/bin/git-coauthor
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+old_message="$(git log --format=%B -n1)"
+git commit --amend -m "$old_message" -m "Co-authored-by: $1 <$2>"
--- a/bin/git-coauthor
+++ b/bin/git-coauthor
@@ -1,4 +1,10 @@
 #!/usr/bin/env bash
 
+coauthor=$1
+email=$2
+
+test -z $coauthor && echo "co-author required." 1>&2 && exit 1
+test -z $email && echo "co-author email required." 1>&2 && exit 1
+
 old_message="$(git log --format=%B -n1)"
-git commit --amend -m "$old_message" -m "Co-authored-by: $1 <$2>"
+git commit --amend -m "$old_message" -m "Co-authored-by: $coauthor <$email>"
--- a/bin/git-coauthor
+++ b/bin/git-coauthor
@@ -1,10 +1,17 @@
 #!/usr/bin/env bash
 
+set -e -o pipefail
+
 coauthor=$1
 email=$2
 
-test -z $coauthor && echo "co-author required." 1>&2 && exit 1
-test -z $email && echo "co-author email required." 1>&2 && exit 1
+test -z "$coauthor" && echo "co-author required." 1>&2 && exit 1
+test -z "$email" && echo "co-author email required." 1>&2 && exit 1
 
 old_message="$(git log --format=%B -n1)"
-git commit --amend -m "$old_message" -m "Co-authored-by: $coauthor <$email>"
+if [[ $old_message == *"Co-authored-by:"* ]]; then
+  git commit --amend -m "$old_message
+Co-authored-by: $coauthor <$email>"
+else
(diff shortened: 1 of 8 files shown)


man/git-coauthor.md
```diff
@@ -0,0 +1,36 @@
+<NAME>(1) -- <short description>
```
spacewander: Need to update the template to use the real `Name` and `description`.

man/git-extras.md
```diff
@@ -31,6 +31,8 @@ git-extras(1) -- Awesome GIT utilities
    - **git-chore(1)** Create chore branch
    - **git-clear-soft(1)** Soft clean up a repository
    - **git-clear(1)** Rigorously clean up a repository
+   - **<NAME>(1)** <short description>
+   - **git-coauthor(1)** Add a co-author to the last commit
```
spacewander: Now I know why the `git-coauthor` is not added automatically.
The `**<NAME>(1)** <short description>` is added instead because the placeholder is not updated.

bin/git-coauthor
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
```
spacewander: We should check the number of the arguments.

man/git-extras.md
jacobherrington: Ahhhh. Thanks, I will make a fix 👍

bin/git-coauthor
jacobherrington: I agree, thanks 👍

jacobherrington:
@spacewander I believe this is ready for another review. Thanks for your help!

bin/git-coauthor
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
spacewander: I think we should add `set -eu -o pipefail` at the top of the script. Better to be strict.

bin/git-coauthor
```diff
@@ -0,0 +1,10 @@
...
+
+coauthor=$1
+email=$2
+
+test -z $coauthor && echo "co-author required." 1>&2 && exit 1
```
spacewander: We should quote the "$coauthor". Otherwise we will be in trouble when the input contains whitespace. Like this one: `git coauthor "John Doe" john@doe`.

bin/git-coauthor
```diff
@@ -0,0 +1,10 @@
...
+test -z $email && echo "co-author email required." 1>&2 && exit 1
+
+old_message="$(git log --format=%B -n1)"
```
spacewander: After running this command twice on the same commit, I found there is a newline between two `Co-authored-by`.
```
commit 698b33b589fdd3d05dc5f0d9912228a78d9a9537 (HEAD -> master)
Author: spacewander <spacewanderlzx@gmail.com>
Date:   Sat Aug 24 11:03:18 2019 +0800

    init
    
    Co-authored-by: John Doe <john@doe>
    
    Co-authored-by: John <john@doe>
```

It's caused by a trailing newline in the output of `git log --format=%B -n1`.
Better to find a way to remove the annoying newline.

spacewander: Maybe we can remove the trailing newline when the original commit message already contains `Co-auhored-by`.

jacobherrington:
@spacewander I made the changes you suggested. 👍

bin/git-coauthor
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
```
spacewander: My bad. The `-u` option raises out`$1: unbound variable` error when `coauthor` argument is missing. We have already checked the argument explicitly and print appropriate error message, so the error is not useful.
Could you remove the `-u` option? Thanks!

spacewander:
@jacobherrington 
Merged. Thank you!

PR #777, opened 2019-09-01
[Doc]Install completion with zplugin

fengkx:
This pr change the ice modifier to install completions using zplugin.


spacewander:
@fengkx 
Merged. Thank you!

PR #779, opened 2019-09-07
Allow git-* scripts to be POSIX shell scripts

pabs3:
Not all of the scripts need bash features.

There are more POSIX shells potentially available on more platforms
than the one implementation of bash available on some platforms.

Some of the POSIX shell implementations are faster than bash.


spacewander:
No, thanks. Only bash is supported by git-extras. We care about the consistency more than performance.

pabs3:
Seems weird, but OK. Please close the PR.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #781, opened 2019-09-11
Check scripts with shellcheck before accepting new commands

pabs3:


spacewander:
shellcheck is great. But to make the shellcheck happy with the existing scripts, we have to do a lot of things.

If we can not distinguish the new change and the existing code in a script, the contributor have to upgrade the existing code to pass the CI. This burden makes the contribution harder, especially when the contributor only wants to add a simple option or correct typo.

As the gatekeeper, I have already used shellcheck to verify each change, and ignored some warning manually. Sometimes `SC2086: Double quote to prevent globbing and word splitting.` is too strict.

pabs3:
My intention was to only check new or changed scripts and I had a
couple of solutions for that, but some missing features in xargs/grep
on macOS and the age of bash on both the test platforms blocked me.
I'll try to work on this more to find a solution that works in travis.

shellcheck has a way to disable particular checks on particular lines:

https://github.com/koalaman/shellcheck/wiki/Directive

In general I've found that SC2086 (quoting) usually either is correct
or is a sign that the script need to be changed to use bash arrays.
Do you have an example of where it was too strict?

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

pabs3:
I've come up with a new approach that checks all bash scripts changed
or added in a commit.

I had to drop testing the scripts on macOS because homebrew takes ages
to install shellcheck and because of some incompatibilities. I figure
it is fine to only check the scripts once as shellcheck probably
behaves the same on both platforms anyway.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #784, opened 2019-09-23
git-effort: adjust column limit according to the paths

### LINKED ISSUE #780: git-effort: adjust column limit if filtering by path
The column size for the output is currently chosen by looking at the entire repo:

https://github.com/tj/git-extras/blob/ff0b1cfb0bc2fe5eff47b8bc2bace379a0085b9f/bin/git-effort#L180-L181

When the user filters by path (ex. `git-effort path/to/folder/*`) it would be better to pick the column based on the files matching the filter, since those names might be significantly shorter than the largest filename in the git repo. Basing it on the entire git repo is safe, but can lead to less-than-ideal output:

```sh
$ git effort [PATH REDACTED]/*

  path                                                                                                                                                                                                                                                                                                                                                                                              commits    active days

  [PATH REDACTED]/36/char/filename.ext............................................................................................................................................................................................................................................................................................................................................................. X           X
  [PATH REDACTED]/41/character/filename.ext........................................................................................................................................................................................................................................................................................................................................................ X           X
```

^ These fake filenames are representative of the actual files. No filename in this particular folder is anywhere close to > 300 characters long.

spacewander:
Fix #780.

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -177,10 +177,6 @@ fi
 args_to_git_log="${args_to_git_log#\ \'\'}"
 export args_to_git_log
 
-# set column width to match longest filename
-columns=$(( $(git ls-files | awk '{ print length }' | sort -rn | head -1 ) + 5 ))
-export columns
-
 # [path ...]
 
 if test "${#paths}" -eq 0; then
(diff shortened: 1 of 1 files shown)


spacewander:
@nfischer 
Could you try this branch to verify if the problem is fixed?

bin/git-effort
```diff
@@ -191,6 +187,17 @@ if test "${#paths}" -eq 0; then
...
 fi
 
+max=0
+for path in "${paths[@]}"; do
+    cur=${#path}
+    if [[ $max -lt $cur ]]; then
+        max=$cur
+    fi
+done
+# set column width to match longest filename
```
nfischer: IMO this comment makes more sense above `max=0`, since that's the logic for finding the longest filename.

spacewander: @nfischer 
Tweaked in the new commit.

PR #785, opened 2019-09-25
Safer protected branches + prevent warning when nothing to clean-up

bobjackman:


bin/git-delete-merged-branches
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-branches=$(git branch --no-color --merged | grep -v "\*" | grep -v master | grep -v svn)
+branches=$(git branch --no-color --merged | grep -v "\*" | grep -vE "\b(master|dev|develop)\b" | grep -v svn)
 if [ -n "$branches" ]
 then
-    echo "$branches" | xargs -n 1 git branch -d
+    echo "$branches" | xargs -rn 1 git branch -d
```
KOGI: prevent error `fatal: branch name required` when there are no merged branches found

bin/git-delete-merged-branches
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-branches=$(git branch --no-color --merged | grep -v "\*" | grep -v master | grep -v svn)
+branches=$(git branch --no-color --merged | grep -v "\*" | grep -vE "\b(master|dev|develop)\b" | grep -v svn)
```
KOGI: protect `develop` branch as well as `master` and match only on whole-words so that branches like `feature/IAmTheGateKeeperAreYouTheKeymaster` aren't inadvertently spared from being cleaned up.

KOGI: * protect `develop` branch as well as `master`
* match only on whole-words so that branches like `feature/IAmTheGateKeeperAreYouTheKeymaster` aren't inadvertently spared from being cleaned up.

KOGI: (the following `grep -v svn` _could_ also be lumped in to this new pattern as `grep -vE "\b(master|dev|develop|svn)\b"`, but I'll defer that decision to you 🙂)

spacewander: @KOGI 
I vote for merging `svn` into the other patterns.

bin/git-delete-merged-branches
spacewander: @KOGI 
It is strange, we have already checked `if [ -n "$branches" ]` before. The `$branches` should not be empty there.

spacewander:
BTW, the documentation of git-delete-merge-branches should be updated too.
https://github.com/tj/git-extras/blob/master/man/git-delete-merged-branches.md#description
https://github.com/tj/git-extras/blob/master/man/Readme.md

bin/git-delete-merged-branches
KOGI: Ah, you're right. I must've had an old installation that I copied my changes from.

bobjackman:
closing PR due to extreme staleness

PR #787, opened 2019-10-02
Use GIT_DIR environment variable to to set .git directory

rvbuelow:
The GIT_DIR environment variable can be used to configure an alternative repository location within GiT. This is very useful i.e. for the management of dotfiles:
```
alias config='GIT_DIR=~/.local/share/dotfiles GIT_WORK_TREE=~ git'
```
This patch replaces the hardcoded `.git` paths in order to make `git ignore -p` command work correctly if GIT_DIR has been set to another location.

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+: ${GIT_DIR:=.git}
+
 function show_contents {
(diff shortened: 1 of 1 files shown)


spacewander:
@rvbuelow 
Merged. Thank you!

PR #789, opened 2019-10-20
Delete empty gitmodules

go2null:
- Handle case where this is the first submodule, so it leaves the git repository as it was before.
- Handle case where this is the last submodule.
- Check before declaring victory.
- Code reformatting and re-organization. 
- Added option to continue on error to allow cleaning up failed deletions.

NOTE: Not validated on Mac OS X and BSD.

### DIFF
--- a/bin/git-delete-submodule
+++ b/bin/git-delete-submodule
@@ -1,27 +1,31 @@
 #!/usr/bin/env bash
 
-test -z "$1" && echo "submodule required" 1>&2 && exit 1
-cd "$(git root)"
-test ! -f .gitmodules && echo ".gitmodules file not found" 1>&2 && exit 2
+abort() {
+	error="$1" && shift
+	echo "FATAL: $*" 1>&2 && exit "$error"
+}
 
-NAME="$(echo "$1" | sed 's/\/$//g')"
-test -z \
-	"$(   git config --file=.gitmodules submodule."$NAME".url)" \
-	   && echo "submodule not found" 1>&2 && exit 3
+test -z "$1"            && abort 1 'Submodule required'
+cd "$(git root)"        || abort 5 'Cannot change to repository root'
+test ! -f '.gitmodules' && abort 2 '.gitmodules file not found'
+
+NAME="${1%/}"
+test -z "$(git config --file='.gitmodules' "submodule.$NAME.url")" \
+	   && abort 3 'Submodule not found'
 
 # 1. Handle the .git directory
 # 1.a. Delete the relevant section from .git/config
-git submodule deinit -f "$NAME" || exit 4
+git submodule deinit -f "$NAME" || abort 4 "Failed to deinitialize $NAME"
 # 1.b. Delete the submodule .git directory
-rm -rf .git/modules/"$NAME"
+rm -rf ".git/modules/$NAME"
 # 1.c. Delete empty submodule directory
 git rm -f "$NAME"
 
 # 2. Handle .gitignore file
 # 2.a. Delete the relevant line from .gitmodules
-git config --file=.gitmodules --remove-section submodule."$NAME"
+git config --file='.gitmodules' --remove-section "submodule.$NAME"
 # 2.b and stage changes
-git add .gitmodules
+git add '.gitmodules'
 # 2.c. Delete empty .gitmodules
 [  "$(wc -l '.gitmodules' | cut -d' ' -f1)" = '0' ] && git rm -f '.gitmodules'
 
@@ -31,7 +35,7 @@ if git submodule status >/dev/null 2>&1 \
 	&& ! git submodule status | grep "$NAME"; then
 	echo "Successfully deleted $NAME."
 else
-	echo "Failed to delete $NAME." 1>&2 && exit 6
+	abort 6 "Failed to delete $NAME."
 fi
 echo
 git submodule status
--- a/bin/git-delete-submodule
+++ b/bin/git-delete-submodule
@@ -33,4 +33,8 @@ if git submodule status >/dev/null 2>&1 \
 else
 	echo "Failed to delete $NAME." 1>&2 && exit 6
 fi
-echo 'Confirm with `git submodule status` and commit the changes for yourself.'
+echo
+git submodule status
+echo
+echo 'Confirm the output of `git submodule status` above' \
+	' and commit the changes for yourself.'
--- a/bin/git-delete-submodule
+++ b/bin/git-delete-submodule
@@ -6,7 +6,7 @@ abort() {
 	test -z "$FORCE" && exit "$error"
 }
 
-# Don't abort on failures.  This allows to cleanup after a rpevious failure.
+# Don't abort on failures.  This allows to cleanup after a previous failure.
 [ "$1" = '--force' ] && FORCE=1 && shift
 
 test -z "$1"            && abort 1 'Submodule required'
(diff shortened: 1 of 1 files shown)


bin/git-delete-submodule
```diff
@@ -1,24 +1,46 @@
...
+	error="$1" && shift
+	echo "ERROR: $*" 1>&2
+	test -z "$FORCE" && exit "$error"
+}
 
-NAME="$(echo "$1" | sed 's/\/$//g')"
-test -z \
-    "$(git config --file=.gitmodules submodule."$NAME".url)" \
-    && echo "submodule not found" 1>&2 && exit 3
+# Don't abort on failures.  This allows to cleanup after a rpevious failure.
```
spacewander: s/rpevious/previous

bin/git-delete-submodule
```diff
@@ -1,24 +1,46 @@
...
+# 2.a. Delete the relevant line from .gitmodules
+git config --file='.gitmodules' --remove-section "submodule.$NAME"
+# 2.b and stage changes
+git add '.gitmodules'
+# 2.c. Delete empty .gitmodules
+[  "$(wc -l '.gitmodules' | cut -d' ' -f1)" = '0' ] && git rm -f '.gitmodules'
+
+# 3. Need to confirm and commit the changes for yourself
+if git submodule status >/dev/null 2>&1 \
+	&& ! git submodule status | grep "$NAME"; then
```
spacewander: Better to cache the ouput of `git submodule status` to avoid running the same command repeatedly

bin/git-delete-submodule
```diff
@@ -1,24 +1,46 @@
...
+	echo "Successfully deleted $NAME."
+else
+	abort 6 "Failed to delete $NAME."
+fi
 echo
-echo "Now submodule $NAME is deleted."
-echo 'Confirm with `git submodule status` and commit the changes for yourself.'
+git submodule status
+echo 'Confirm the output of `git submodule status` above' \
```
spacewander: Maybe we could move the `git submodule status` under the prompt, and use `git submodule status ... below` instead?

bin/git-delete-submodule
```diff
@@ -34,13 +34,16 @@ git add '.gitmodules'
 [  "$(wc -l '.gitmodules' | cut -d' ' -f1)" = '0' ] && git rm -f '.gitmodules'
 
 # 3. Need to confirm and commit the changes for yourself
-if git submodule status >/dev/null 2>&1 \
-	&& ! git submodule status | grep "$NAME"; then
+git_status_text="$(git submodule status >/dev/null 2>&1)"
```
spacewander: Err, now the stdout of `git submodule status` goes to `/dev/null` too.

bin/git-delete-submodule
```diff
@@ -34,13 +34,16 @@ git add '.gitmodules'
...
+git_status_exit=$?
+if [ "$git_status_exit" -eq 0 ] \
+	&& printf '%s' "$git_status_text" | grep "$NAME"; then
```
spacewander: Maybe you miss reverting the exit code in this commit?

bin/git-delete-submodule
```diff
@@ -1,24 +1,49 @@
...
+# 1.a. Delete the relevant section from .git/config
+git submodule deinit -f "$NAME" || abort 4 "Failed to deinitialize $NAME"
+# 1.b. Delete the submodule .git directory
+rm -rf ".git/modules/$NAME"
+# 1.c. Delete empty submodule directory
+git rm -f "$NAME"
+
+# 2. Handle .gitignore file
+# 2.a. Delete the relevant line from .gitmodules
+git config --file='.gitmodules' --remove-section "submodule.$NAME"
```
spacewander: It looks like the `submodule.$NAME` section is already removed by the `git rm -f` above. Running this command gets `fatal: No such section!` error. And the `.gitmodules` is added to the index too.
Maybe we just need to remove the empty `.gitmodules`?

bin/git-delete-submodule
```diff
@@ -1,24 +1,49 @@
...
+
+# 1. Handle the .git directory
+# 1.a. Delete the relevant section from .git/config
+git submodule deinit -f "$NAME" || abort 4 "Failed to deinitialize $NAME"
+# 1.b. Delete the submodule .git directory
+rm -rf ".git/modules/$NAME"
+# 1.c. Delete empty submodule directory
+git rm -f "$NAME"
+
+# 2. Handle .gitignore file
```
spacewander: Should be `.gitmodules`

bin/git-delete-submodule
go2null: fixed

bin/git-delete-submodule
go2null: not sure what you mean, could you elaborate?

bin/git-delete-submodule
go2null: maybe this change is for later git versions?
I've ignored the error, but can change the code to remove it.
Please let me know.

bin/git-delete-submodule
spacewander: @go2null 
If the operation is success, the `$NAME` should not be found in the `git submodule status` and `grep "$NAME"` should be failed.
If I am wrong, please correct me.

bin/git-delete-submodule
```diff
@@ -1,24 +1,50 @@
...
+# 2.b and stage changes
+git add '.gitmodules'
+# 2.c. Delete empty .gitmodules
+[  "$(wc -l '.gitmodules' | cut -d' ' -f1)" = '0' ] && git rm -f '.gitmodules'
+
+# 3. Need to confirm and commit the changes for yourself
+git_status_text="$(git submodule status 2>&1)"
+git_status_exit=$?
+if [ "$git_status_exit" -eq 0 ] \
+	&& printf '%s' "DUMMY$git_status_text" | grep -v "$NAME"; then
```
spacewander: @go2null 
Is the `DUMMY` prefix necessary?

bin/git-delete-submodule
spacewander: @go2null 
You are right. I tested it with an old version of git (1.8.3.1), and `git rm -f` didn't remove the section. A newer version of git (2.17) does.

go2null: ok, so ignoring the error is good

bin/git-delete-submodule
go2null: you are correct, I changed to `grep -v`.

bin/git-delete-submodule
go2null: yes, because `grep` fails when piping in an empty string.
(Note that there is no `\n` as I'm using `printf`, not `echo`.)

go2null:
BTW, here is the script I used to test.  Not sure where to save it, attaching it here.
[test-git-delete-submodule.txt](https://github.com/tj/git-extras/files/3768593/test-git-delete-submodule.txt)

bin/git-delete-submodule
```diff
@@ -1,24 +1,50 @@
...
+	echo "Successfully deleted $NAME."
+else
+	abort 6 "Failed to delete $NAME with error:\n$git_status_text"
+fi
 echo
-echo "Now submodule $NAME is deleted."
-echo 'Confirm with `git submodule status` and commit the changes for yourself.'
+# shellcheck disable=SC2016
+echo 'Confirm the output of `git submodule status` below' \
+	' and then commit the changes.'
```
spacewander: @go2null 
Sorry, I change my mind now. Now I think it would be better if we put the `git submodule status` above the prompt.
The output of the `git submodule status` can be too long and make the prompt unnoticeable.

BTW, you don't need to add space before the line next to `\`. A space is already added by bash.
For instance:
```
$ echo "aa" \
> a
aa a
```

go2null: done

spacewander:
Merged. Thanks!

PR #798, opened 2019-11-24
Signed release

wolviecb:
This PR adds support for gpg signed tags for the `git-release` command

```git release -s 1```

Will generate a tag `1` signed by the default gpg key

```git release -u key-id 1```

Will generate a tag `1` signed by `key-id`

### DIFF
--- a/bin/git-release
+++ b/bin/git-release
@@ -34,6 +34,17 @@ if test $# -gt 0; then
     -c) need_changelog=true;;
     -r) remote=$2; shift ;;
     -m) msg=$2; shift ;;
+    -s)
+      test -n "$keyid" &&
+          exit_with_msg "Please use '-s' OR '-u'"
+      sign=true
+      ;;
+    -u)
+      test -n "$sign" &&
+          exit_with_msg "Please use '-s' OR '-u'"
+      keyid=$2
+      shift
+      ;;
     --semver)
       test -z "$2" &&
           exit_with_msg "major/minor/patch required for --semver option"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: bin/git-release [hunk 1]
```diff
@@ -101,8 +103,17 @@ if test $# -gt 0; then
     git commit -a -m "$msg" --allow-empty
   fi
 
+  declare -a sign_args
```
spacewander: Since either `-s` or `-u` is allowed, maybe we should add an argument check to warn users when both `-s` and `-u` are provided?

spacewander:
@wolviecb
Please squash the commits into one before we can merge it, thanks!

wolviecb:
Changes made, commits squashed, thx for the feedback :)

bin/git-release
```diff
@@ -34,6 +34,17 @@ if test $# -gt 0; then
     -c) need_changelog=true;;
     -r) remote=$2; shift ;;
     -m) msg=$2; shift ;;
+    -s)
+    test -n "$keyid" &&
```
spacewander: A little cavil: please indent the `-s)` block like the `--semver)` one below.

bin/git-release
```diff
@@ -34,6 +34,17 @@ if test $# -gt 0; then
...
+      exit_with_msg "Please use '-s' OR '-u'"
+      sign=true
+      ;;
+    -u)
+    test -n "$sign" &&
```
spacewander: Ditto

bin/git-release
wolviecb: done

bin/git-release
wolviecb: done

spacewander:
@wolviecb 
Merged. Thanks!

PR #800, opened 2019-11-28
Separate bulk projects

roxchgt:
`git-bulk` as it is now, does not support working on more than one project with multiple repositories simultaneously. Purging and adding all workspaces (repositories) has to be done whenever you'd need to switch between two projects, since all workspaces are added to global git config under the same section.

With the following changes each repository can keep a project name in its own local git config and based on that name there will be a new bulkws-section in the global config with which the list of repositories can be filtered at the time of bulk operations. Adding project name can be done directly via new function (in the git work tree) or interactively while adding workspace if not defined. Purging does also delete the local project name.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -22,7 +22,7 @@ usage() {
 }
 
 # add project name to current repository
-function addproject { git config --local --add project.name "$projectname"; }
+function addproject { git config --local --replace-all project.name "$projectname"; }
 
 # add another workspace to global git config
 function addworkspace { git config --global bulkworkspaces."$wsname" "$wsdir"; }
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -58,13 +58,13 @@ function addworkspace {
 function addcurrent { wsdir=$(git root); addworkspace; }
 
 # remove workspace from global git config
-function removeworkspace { checkWSName && git config --global --unset bulkworkspaces."$wsname"; }
+function removeworkspace { checkWSName && git config --global --unset bulkws-$(getproject)."$wsname"; }
 
 # remove workspace from global git config
-function purge { git config --global --remove-section bulkworkspaces; }
+function purge { git config --global --remove-section bulkws-$(getproject); }
 
 # list all current workspace locations defined
-function listall { git config --global --get-regexp ${getproject}; }
+function listall { git config --global --get-regexp bulkws-$(getproject); }
 
 # guarded execution of a git command in one specific repository
 function guardedExecution () {
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -20,6 +20,7 @@ usage() {
   echo 1>&2 "       git bulk --addcurrent <ws-name>"
   echo 1>&2 "       git bulk --purge"
   echo 1>&2 "       git bulk --listall"
+  echo 1>&2 "       git bulk --listcurrent"
 }
 
 function to_lowercase { echo $1 | tr [:upper:] [:lower:]; }
@@ -32,9 +33,11 @@ function addproject { git config --local --replace-all project.name $(to_lowerca
 
 # add another workspace to global git config
 function addworkspace {
+	current=$(pwd)
+	cd $wsdir
 	project=$(getproject)
 	if [ -z $project ]; then
-		echo -n "no project name is defined, "
+		echo -n "no project name is defined in this path, "
 		local project=$(to_lowercase $(basename $(dirname $(git root))))
 		local prompt="\e[1m\e[4m$project\e[0m"
 		while true; do
@@ -42,7 +45,8 @@ function addworkspace {
 			case $answer in
 				[Yy])
 					addproject $project
-					sleep 0.3; echo -e "project is called \e[1m$project\e[0m now."
+					cd "$current"
+					sleep 0.1; echo -e "project is called \e[1m$project\e[0m now."
 					break;;
 				[Nn])
 					echo -e "\e[1m\e[31m!\e[0m add a project name first."; usage; exit;;
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -13,7 +13,7 @@ allwsmode=false
 # print usage message
 #
 usage() {
-  echo 1>&2 "usage: git bulk [-g] ([-a]|[-w <ws-name>]) <git command>"
+  echo 1>&2 "usage: git bulk [-g] ([-a]|[-p <project-name>]|[-w <ws-name>]) <git command>"
   echo 1>&2 "       git bulk --addproject <project-name>"
   echo 1>&2 "       git bulk --addworkspace <ws-name> <ws-root-directory>"
   echo 1>&2 "       git bulk --removeworkspace <ws-name>"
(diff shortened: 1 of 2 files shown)


roxchgt:
WIP: `Commands.md`

spacewander:
Look like this PR changes the behavior of `listall` subcommand.
Previously, this command can be run outside a git repo. But now I get:
```bash
$  ~/git/git-extras/bin/git-bulk --listall
fatal: --local can only be used inside a git repository
```

Is it possible to add a `project` argument to `listall` and `purge` subcommands so that we can avoid API breaks?

bin/git-bulk
```diff
@@ -14,27 +14,66 @@ allwsmode=false
...
   echo 1>&2 "       git bulk --removeworkspace <ws-name>"
   echo 1>&2 "       git bulk --addcurrent <ws-name>"
   echo 1>&2 "       git bulk --purge"
   echo 1>&2 "       git bulk --listall"
 }
 
+function to_lowercase { echo $1 | tr [:upper:] [:lower:]; }
+
+# helper to get project name
+function getproject { git config --local --get project.name; }
```
spacewander: Is this function expected to be run under non-git repo? A workspace is not always a git repo. Maybe we need to add `2> /dev/null || echo ''` after it?

spacewander:
> git-bulk as it is now, does not support working on more than one project with multiple repositories simultaneously. Purging and adding all workspaces (repositories) has to be done whenever you'd need to switch between two projects, since all workspaces are added to global git config under the same section.

Could you provide more context?
If I understand your idea correctly, you want to run git command across workspace A and worksapce B (in the project α), or across workspace B and workspace C (in the project β).

To me, it is simpler to solve the problem by allowing multiple -w arguments, so to run command in project α, you can just specify `-w A -w B`. Add a new concept is harder to understand and implement correctly.

bin/git-bulk
roxchgt: well I have been always thinking in the context of git repositories, since it's all about git and git-bulk in a directory without `.git` in it is assumed out of scope by me. I'll take a more generic look at it later today!

roxchgt:
> > git-bulk as it is now, does not support working on more than one project with multiple repositories simultaneously. Purging and adding all workspaces (repositories) has to be done whenever you'd need to switch between two projects, since all workspaces are added to global git config under the same section.
> 
> Could you provide more context?
> If I understand your idea correctly, you want to run git command across workspace A and worksapce B (in the project α), or across workspace B and workspace C (in the project β).
> 
> To me, it is simpler to solve the problem by allowing multiple -w arguments, so to run command in project α, you can just specify `-w A -w B`. Add a new concept is harder to understand and implement correctly.
> 
> If you think `-w A -w B` is too long to type, you can use a little alias, like:
> 
> ```
> alias git-bulk-α='git bulk -w A -w B'
> ```

Well, literally you could all of git-extras stuff with aliases… it basically abstracts combinations of git and other commands in functions which makes it a handy tool out-of-the-box

To give an example or use case:
Assume I'm working on multi-repo project A and separately also on multi-repo B. Each of them consist of *n* and *m* git repository respectively. If I want to use git bulk on A and B independently, I'll have a problem: Since I can add *n* workspaces of A to git bulk and then the *m* workspaces of B likewise. Now they're all listed under [bulkworkspaces] section together like this:
```
[bulkworkspaces]
        projectA-workspace1 = /path/to/A/workspace1
        projectA-workspace2 = /path/to/A/workspace2
        projectA-workspace3 = /path/to/A/workspace3
        projectB-workspace1 = /path/to/B/workspace1
        projectB-workspace2 = /path/to/B/workspace2
        projectB-workspace3 = /path/to/B/workspace3
        projectB-workspace4 = /path/to/B/workspace4
```
what if i want to checkout branch `branchA` in all repos of project A and `branchB` in repos of Project B? It's not possible (maybe using guarded mode or giving the name of all workspaces of A in the first round with `-w` flag and for B likewise… not so ideal (I'm on a real world microservice project with more than 12 repositrories and another private one with also couple of repositories…

With adding project name to each of those repos of A and B separately, i can tell git bulk to go through which section or in this context project. hope it made my intention clear

bin/git-bulk
roxchgt: @spacewander I commited a small change to `--listall` which fixes this bug and the function returns all registered workspaces. So the API is not broken.
Now I should probably extend the function to return only workspaces of a given project / or current

roxchgt:
This needs still some work… will be reopened soon

spacewander:
@roxchgt 
The complexity and the API break still give me a headache...
What about adding a non-git repo as a workspace? Store the project info in the local git configure doesn't handle this case.
Allow multiple `-w` is simpler. 

> what if i want to checkout branch branchA in all repos of project A and branchB in repos of Project B? It's not possible (maybe using guarded mode or giving the name of all workspaces of A in the first round with -w flag and for B likewise… not so ideal (I'm on a real world microservice project with more than 12 repositrories and another private one with also couple of repositories…

You don't need to type all the workspace each time you run a git command, just writing them down in the alias section of `~/.gitconfig` is enough. Writing more then 12 repos in the alias is almost the same as adding more than 12 workspaces to a project.

roxchgt:
> @roxchgt
> The complexity and the API break still give me a headache...
Yeah! I can understand that… especially for those who are already using it (probably on smaller projects) and wouldn't wish any changes. I was just thinking to make this cool tool also suitable for enterprise projects, since people are not willing to use sth which comes to its limits in a complicated architecture. Anyway the API is not broken anymore after my last couple of commits (yeah it was as i pushed first and it was my bad). You can use it like before and adding a project is totally optional and you can leave it empty…Have you tested it since? or been too busy!?

> What about adding a non-git repo as a workspace? Store the project info in the local git configure doesn't handle this case.
I don't get it! Why you would add a non-git repo to `git-bulk`?? Would you need to execute git operations on non-git repos afterwards? Doesn't make really sense.

> Allow multiple `-w` is simpler.
> ... just writing them down in the alias section of `~/.gitconfig` is enough. Writing more then 12 repos in the alias is almost the same as adding more than 12 workspaces to a project.

well at the end of the day I could go on and write:
```bash
~/myprojectroot/
for dir in */
do
    cd $dir
    git pull  # for exmaple
    cd ../
done
```
Or simply write this in the a `git-all.sh` and run it at the base directory or more simple: alias…
But we want to have a out-of-the-box tool which has many capabilities, don't we?
So let's boil this feature down to an ideal solution, without over-engineering it!
I'll be looking forward for your revision

roxchgt:
@spacewander 💡Now i have a much simpler idea to handle this! The whole `git diff` will be less than this approach and nothing will be saved/added locally! Everything will refer tothe same section of `~/.gitconfig` and identifier will be added to `ws-name` and using the power of `--get--regexp` we can optionally filter the workspaces belonging to a group/project, :)
I'll close this request and work on a different branch from scratch and when finished open another pull req… Thanks for your tips 罗泽轩

spacewander:
> I don't get it! Why you would add a non-git repo to git-bulk?? Would you need to execute git operations on non-git repos afterwards? Doesn't make really sense.

The workspace can be a directory which contains multiple git repos. In this situation, a command applies to this workspace will be run in each git repo. It is useful when you have a `~/work` directory and want to run `git pull` in each git repo.

roxchgt:
@spacewander  
> …will be run in each git repo. It is useful when you have a `~/work` directory and ...
OOoh! I didn't notice that… wwhhhy I was using it wrong the whole time! I always added each path to git repo as a workspace and never thought it could be a dir containing them… 🤦‍♂️ _(although that's what the documentation says…)_

With that the issue is over! There was actually no issue but not using `git bulk` properly.

me now: https://media.giphy.com/media/jUwpNzg9IcyrK/giphy.gif

PR #802, opened 2019-12-01
fix bug in allowedargcount() to not execute command if wrong args count

spacewander:
This fix is taken from #800.
I did a little modification to resolve conflict.
Thanks @roxchgt for the fix.


spacewander:
@roxchgt 
Merged. Thanks!

PR #811, opened 2019-12-12
build: do not require the nonstandard and unpredictable 'which' utility

eli-schwartz:
This may not be installed on various systems, and it's difficult to test for the availability of the tool you need, if the check program itself does not exist.

The POSIX 2008 specification mandates the `command -v` builtin; bash is a POSIX 2008 compliant shell, and this builtin has worked since bash 1.x anyway.

A side benefit of using the POSIX portable option is that it requires neither an external disk executable, nor (because unlike "which", the exit code is reliable) a subshell fork. This therefore represents a mild speedup.

### DIFF
--- a/check_dependencies.sh
+++ b/check_dependencies.sh
@@ -5,6 +5,6 @@ err() {
     exit 1
 }
 
-if ! test "$(which column)"; then
+if ! command -v column >/dev/null 2>&1; then
     err "Need to install dependency 'column' before installation"
 fi


eli-schwartz:
Because I don't have `which` installed on my system, so `make install` failed.

eli-schwartz:
Totally as a side note, the script doesn't actually make use of bash-specific features, so it could be modified to run via /bin/sh; it's not the only script in that situation.

spacewander:
@eli-schwartz 
Merged. Thank you!

We only target to bash: https://github.com/tj/git-extras/blob/master/CONTRIBUTING.md#your-new-git-extra-command-should-support
Some scripts don't use bash-specific features but some do. Enforce `/usr/bin/env bash` keeps the consistence, so that we don't need to remember which script is bash-only while which is not.

eli-schwartz:
Yeah, probably makes sense I suppose. I'm just a persnickety /bin/sh script writer. :D

PR #815, opened 2019-12-22
Added search and replace for invalid characters in branch name. A '\'…

DarekDan:
…, '/', and ':', will be replaced with '_'.

### DIFF
--- a/bin/git-archive-file
+++ b/bin/git-archive-file
@@ -20,6 +20,8 @@ else
 	if [ "$BRANCH" = "master" ]; then
 		FILENAME=$ARCHIVE_NAME.$VERSION.zip
 	else
+	  # govern against invalid path characters
+	  BRANCH=$(echo "$BRANCH" | sed -r 's/[\/\\\:]+/_/g')
 		FILENAME=$ARCHIVE_NAME.$VERSION.$BRANCH.zip
 	fi
 fi


bin/git-archive-file
```diff
@@ -20,6 +20,8 @@ else
 	if [ "$BRANCH" = "master" ]; then
 		FILENAME=$ARCHIVE_NAME.$VERSION.zip
 	else
+	  # govern against invalid path characters
```
spacewander: Please follow the indentation of the other code.

bin/git-archive-file
```diff
@@ -20,6 +20,8 @@ else
...
+	  BRANCH=$(echo "$BRANCH" | sed -r 's/[\/\\\:]+/_/g')
```
spacewander: As `g` option is specific, is it necessary to use `+` in the regex?
If `+` can be omitted, `-r` can be omitted too.

bin/git-archive-file
DarekDan: The indentation is exactly the same in my IDE, IntelliJ.
19 12 23 09 04 51-git-extras  C__Repos_git-extras  -  _bin_git-archive-file - IntelliJ IDEA (Adm

bin/git-archive-file
DarekDan: It does not hurt, and covers all future scenarios.

bin/git-archive-file
spacewander: @DarekDan
Look like your IDE uses 2 spaces to represent a tab, and converts a tab into two space. Would you change the tabs around into 4 spaces?

bin/git-archive-file
spacewander: @DarekDan
Less code, less bugs.
For example, bsd `sed` doesn't have the `-r` option. If the extended regex is required, we have to write more code to work around.

PR #823, opened 2020-03-15
Add git-browse command

### LINKED ISSUE #666: FEATURE REQUEST: command to view the repo online
It would be great if there was a command to view the repo online. It would help me save some clicking around on my computer (ie, open up browser, type URL, press enter; versus a single simple CLI command that I could even alias to something shorter).

This is essentially going to be the equivalent of [git town repo](https://github.com/Originate/git-town/blob/master/documentation/commands/repo.md).

I don't mind personally working on a PR to introduce this, I just want to check first if there is interest to adding this command. Thank you.

mapitman:
Handles git@server and https://server remotes.
Allows specification of the remote to open
Closes #666

Tested on Linux, WSL and macOS.

### DIFF
--- a/bin/git-browse
+++ b/bin/git-browse
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+if [[ $1 == "" ]]
+then
+    remote=$(git remote | head -n 1)
+else
+    remote=$1
+fi
+
+if [[ $remote == "" ]]
+then
+    echo "Remote not found"
+    exit 1
+fi
+
+remote_url=$(git remote get-url $remote)
+
+if [[ $? -ne 0 ]]
+then
+    exit $?
+fi
+
+if [[ $remote_url = git@* ]]
+then
+    url=$(echo $remote_url | sed -E -e 's/:/\//' -e 's/\.git$//' -e 's/\w*@(.*)/https:\/\/\1/')
+elif [[ $remote_url = https* ]]
+then
+    url=$(echo $remote_url | sed -E -e 's/\.git//')
+fi
+
+case "$OSTYPE" in
+    darwin*)
+        # MacOS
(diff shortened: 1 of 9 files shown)


bin/git-browse
```diff
@@ -0,0 +1,53 @@
...
+remote_url=$(git remote get-url $remote)
+
+if [[ $? -ne 0 ]]
+then
+    exit $?
+fi
+
+if [[ $remote_url = git@* ]]
+then
+    url=$(echo $remote_url | sed -E -e 's/:/\//' -e 's/\.git$//' -e 's/\w*@(.*)/https:\/\/\1/')
```
spacewander: Would it be better to use `http` instead of `https`? The `http` is always supported, but `https` might not. No `https` by default looks bad, however, if the web server supports `https`, it will redirect the browser to the `https` endpoint. So the end result is the same.

spacewander: Would it be better to use `http` instead of `https`? The `http` is always supported, but `https` might not. No `https` by default looks bad, however, if the web server supports `https`, it will redirect the browser to the `https` endpoint. So the end result is the same.

bin/git-browse
```diff
@@ -0,0 +1,53 @@
...
+elif [[ $remote_url = https* ]]
+then
+    url=$(echo $remote_url | sed -E -e 's/\.git//')
```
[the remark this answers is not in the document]
mapitman: Done in second commit

bin/git-browse
mapitman: Done in second commit

mapitman: Done in second commit

spacewander:
@mapitman 
Merged. Thanks!

PR #826, opened 2020-03-19
Add git-cp

### LINKED ISSUE #825: git cp — Duplicate a file and keep history on both files
This would help with splitting large files while other PR updates it.

Natim:
Fixes #825 

Early feedbacks welcome!

- [x] Add man pages

Refs https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history/44036771#44036771


bin/git-cp
```diff
@@ -0,0 +1,57 @@
...
+    # We come back to the previous state
+    git reset --hard HEAD^  # Revert that change
+
+    # We move the file to its new destination
+    git mv "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
+    git add "${DESTINATION_FILENAME}"
+    git commit -nm "Copy $CURRENT_FILENAME into $DESTINATION_FILENAME"
+
+    # We get back our copy
+    rm -f "${DESTINATION_FILENAME}"  # Make sure the file doesn't exists somehow
```
spacewander: Maybe we need to use `rm -rf`, because the destination might be a directory.

bin/git-cp
```diff
@@ -0,0 +1,57 @@
...
+    git merge "${SAVED}"
+    git commit -a -m "Duplicate ${CURRENT_FILENAME} history."  # Keeping both files
+
+    # We get back our original name
+    git mv "${INTERMEDIATE_FILENAME}" "${CURRENT_FILENAME}"
+    git commit -nm "Set back $CURRENT_FILENAME file."
```
spacewander: Is it possible to squash multiple commits into one? Or is it intended?

bin/git-cp
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
spacewander: Let's add `set -euo pipefail` protection for this complex script...

bin/git-cp
```diff
@@ -0,0 +1,57 @@
...
+PROGRAM=$0
+CURRENT_FILENAME=""
+DESTINATION_FILENAME=""
+
+function usage()
+{
+    echo "USAGE: ${PROGRAM} CURRENT_FILENAME DESTINATION_FILENAME"
```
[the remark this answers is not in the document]
Natim: Do you know of another command in /bin that is doing that? Most of the one I opened are using echo or cat without stderr redirections.

bin/git-cp
Natim: This might be a bit dangerous in my opinion. Even removing the file itself is a bit bold. Maybe we should error before starting the copy?

bin/git-cp
Natim: Yes it is intended for the history to be kept.

bin/git-cp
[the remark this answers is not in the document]
Natim: I found git-bulk

bin/git-cp
spacewander: @Natim 
Maybe we can just leave one commit as the history to keep? Create four new commits per operation is too scary to me.

bin/git-cp
```diff
@@ -0,0 +1,64 @@
...
+    git add "${DESTINATION_FILENAME}"
+    git commit -nm "Copy $CURRENT_FILENAME into $DESTINATION_FILENAME"
+
+    # We get back our copy
+    git merge "${SAVED}"
+    git commit -a -m "Duplicate ${CURRENT_FILENAME} history."  # Keeping both files
+
+    # We get back our original name
+    git mv "${INTERMEDIATE_FILENAME}" "${CURRENT_FILENAME}"
+    git commit -nm "Set back $CURRENT_FILENAME file."
```
spacewander: @Natim
Maybe we can just leave one commit as the history to keep? Create four new commits per operation is too scary to me.

bin/git-cp
Natim: I invite you to read : https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history/44036771#44036771

It is because it cannot be done with one commit that I had to create git-cp in the first place.
I wish there were a better way.

bin/git-cp
Natim: If you figure out a way to reduce the number of commit I would be interested indeed.

Natim: It might be possible using git plumbering

Natim: All we have to do is to create a new object with the same parent object.

Natim: > **osse on #git-devel IRC channel**:
> file objects as in blobs? they don't have parents

Natim: Refs https://stackoverflow.com/a/1043566/186202

bin/git-cp
```diff
@@ -0,0 +1,40 @@
...
+
+if [[ "$DESTINATION_FILENAME" == "" ]]; then
+    usage
+    exit 20  # Missing arguments CURRENT_FILENAME
+elif [[ "$CURRENT_FILENAME" == "" ]]; then
+    usage
+    exit 10  # Missing arguments CURRENT_FILENAME
+else
+    echo "Coping $CURRENT_FILENAME into $DESTINATION_FILENAME"
+	cp -r "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
```
spacewander: Please keep the indent consistent (use 4 spaces like other lines in this file).

bin/git-cp
```diff
@@ -0,0 +1,40 @@
...
+
+	git add "$DESTINATION_FILENAME"
```
spacewander: Ditto.

spacewander:
@Natim 
Merged. Thanks!

PR #827, opened 2020-03-19
Update man page export for git-psykorebase with ronn-ng 0.9.0

Natim:


spacewander:
@Natim 
LGTM. Merged. Thanks!

PR #832, opened 2020-04-20
added color to comments for easy review

victorhck:
added color to git-info comments to highlight them


spacewander:
@victorhck 
I think it would be better to provide a separate `--color` option instead.

Issue #834, opened 2020-05-12
git-feature prefix needs config setting


jldugger:
I commonly use a personal prefix for short lived branches, i.e. `jldugger/fix-bug-321`. git-feature supports a prefix option, but provides no means of setting a default. My first approach was to simply apply an alias, as one does in bash, to use the flag by default. However, this does not work, [by design](https://git-scm.com/docs/git-config#Documentation/git-config.txt-alias):

> To avoid confusion and troubles with script usage, aliases that hide existing Git commands are ignored. 

Can we add git config setting for this, so I can set a global default name for the prefix?

spacewander:
You can use git config git-extras.feature.xxx to store the default prefix. Pull request is welcome!

spacewander:
Solved.

PR #836, opened 2020-05-13
add a config setting for prefix

jldugger:
Per feedback on #834 , here is a one-liner change to fetch a default branch prefix from `git config.`

### DIFF
--- a/bin/git-feature
+++ b/bin/git-feature
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-branch_prefix=feature
+branch_prefix=$(git config --get git-extras.feature.prefix)
+
+if [ -z "$branch_prefix" ]; then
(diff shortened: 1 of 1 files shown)


bin/git-feature
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-branch_prefix=feature
+branch_prefix=$(git config --default feature --get git-extras.feature.prefix)
```
spacewander, line 3: Look like `--default` only exists in newer version of `git`. I gets `error: unknown option default` when running this command. My git version is `git version 2.17.1`.

jldugger: Okay, thanks for testing for me, will rework this.

spacewander:
@jldugger 
Merged. Thank you!

PR #844, opened 2020-05-30
docs: mention installation via FreeBSD pkg

ihoro:
I guess it's possible to do so for more than 5 years, but the package follows release versions only. Probably it will be useful for someone if it's mentioned explicitly.


spacewander:
@ihoro 
Merged. Thank you!

PR #845, opened 2020-05-30
Remove scripts that are mere alias wrappers for git-feature

### LINKED ISSUE #842: Please remove git-bug
This project has some very useful scripts, but it is way too aggressive about stealing subcommand namespaces. This makes it hard to use wholesale and I have to pick and choose.

A notable example is `git-bug`. In this project it's only functioning as a super light weight alias for `git feature -a bug`. This might be useful for some people, but I would much rather specify this as an alias in my own git config (which is easier to share between machines) and reserve scripts for things that actually **do work** that couldn't be done by a oneliner or alias.

The [real `git-bug`](https://github.com/MichaelMure/git-bug) is a quite successful open source project of it's own (4.5k stars on Github) and it makes a lot more sense to keep that namespace free so people can use it instead.

I believe the same goes for several other small scripts that are just aliases for other functions. Please remove these so the namespaces are clear for other current or future uses. Instead for these one-liners have a block of git aliases ready for copy/pasting that is easier for people to tweak and mix in with their systems.

alerque:
Closes #842.

Notably `git-bug` clobbers a very popular other project of the same name.
Having lightweight wrapper scripts clogging up the subcommand namespace
makes it awkward to make use of this project wholesale instead of
piecemeal. These function can be implemented by users with git aliases
very easily.

### DIFF
--- a/bin/git-bug
+++ b/bin/git-bug
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-
-ALIAS=bug
-git feature -a $ALIAS "$@"
--- a/bin/git-chore
+++ b/bin/git-chore
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-
-ALIAS=chore
-git feature -a $ALIAS "$@"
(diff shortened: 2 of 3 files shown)


spacewander:
@alerque 
Merged. Thanks!

PR #849, opened 2020-06-16
Correct undo documentation to match code order of parameters

vr8hub:
The undo documentation has commit count as the first parameter and --soft/hard as the second. The code has the order reversed. This corrects the documentation to match the code.


spacewander:
@vr8hub 
Merged. Thank you!

PR #850, opened 2020-06-18
Modify to work when only a single commit, add parameter checks

vr8hub:
Undo currently does not work when there is only a single commit. There are instances when doing a soft undo on a repository with only a single commit is desired.
In implementing that, I added a few parameter checks, including:
* Cannot specify an undo count greater than the commit count.
* Cannot hard undo all commits, e.g. if commit count is 3, can't hard undo all 3.
* Catch bad parameters, i.e. something other than -s/--soft/-h/--hard as the first parameter or a non-numeric as the second parameter.

Tested on OSX 10.14.6, which is all I have available. I don't believe I used anything that isn't portable, however. (Famous last words.)

### DIFF
--- a/bin/git-undo
+++ b/bin/git-undo
@@ -64,7 +64,7 @@ EOL
     ;;
 esac
 
-if [[ ! $parm2 =~ [1-9]*([0-9]) ]]; then
+if [[ ! $parm2 =~ ^[1-9][0-9]*$ ]]; then
     echo "Invalid undo count: $parm2"
     exit 1
 fi
--- a/bin/git-undo
+++ b/bin/git-undo
@@ -1,6 +1,28 @@
 #!/usr/bin/env bash
 
 back="^"
+cstr="commits"
+ccnt=$(git rev-list --count HEAD)
+if [ $ccnt -eq 1 ]; then cstr="commit"; fi
+parm3=""
+
+function _undo()
+{
+  type=${1:-soft}
+  undo_cnt=${2:-1}
+  reset=${3:-""}
+echo "type=$type, cnt=$undo_cnt, reset=$reset"
+  if [ $undo_cnt -gt $ccnt ]; then
+    echo "Only $ccnt $cstr, cannot undo $undo_cnt"
+  elif [ "$type" = "hard" -a $ccnt -eq $undo_cnt ]; then
+    echo "Cannot hard undo all commits"
+  elif [ "$type" = "soft" -a $ccnt -eq 1 ]; then
+    git update-ref -d HEAD
+  else
+    git reset --$type HEAD~$undo_cnt
+  fi
+  if [ "$reset" != "" ]; then git reset; fi
+}
 
 case "$1" in
   -h)
@@ -12,9 +34,8 @@ EOL
     read -r res
     case "${res}" in
       "Y" | "y")
-        test $2 -gt 1 > /dev/null 2>&1 && back="~$2"
-        git reset --hard HEAD$back
-        exit $?
+        parm1=hard
+        parm2=${2:-1}
         ;;
       * )
         exit 0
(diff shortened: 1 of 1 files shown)


bin/git-undo
```diff
@@ -1,6 +1,28 @@
 #!/usr/bin/env bash
 
 back="^"
```
spacewander, line 3: The `$back` is unused now. Please remove it too.

bin/git-undo
```diff
@@ -1,6 +1,28 @@
...
+ccnt=$(git rev-list --count HEAD)
+if [ $ccnt -eq 1 ]; then cstr="commit"; fi
+parm3=""
+
+function _undo()
+{
+  type=${1:-soft}
+  undo_cnt=${2:-1}
+  reset=${3:-""}
+echo "type=$type, cnt=$undo_cnt, reset=$reset"
```
spacewander, line 14: Please remove the debug log.

bin/git-undo
```diff
@@ -12,26 +34,40 @@ EOL
     read -r res
     case "${res}" in
       "Y" | "y")
-        test $2 -gt 1 > /dev/null 2>&1 && back="~$2"
-        git reset --hard HEAD$back
-        exit $?
+        parm1=hard
+        parm2=${2:-1}
```
spacewander, line 38: Better to use more reasonable argument name than `parmX`.

bin/git-undo
```diff
@@ -12,26 +34,40 @@ EOL
...
   *)
-    test $1 -gt 1 > /dev/null 2>&1 && back="~$1"
-    git reset --soft HEAD$back
-    git reset
+    parm1=soft
+    parm2="$1"
     ;;
 esac
+
+if [[ ! $parm2 =~ [0-9]*([0-9]) ]]; then
```
spacewander, line 68: Should it be `[1-9]*` instead?

bin/git-undo
```diff
@@ -1,6 +1,28 @@
...
+
+function _undo()
+{
+  type=${1:-soft}
+  undo_cnt=${2:-1}
+  reset=${3:-""}
+echo "type=$type, cnt=$undo_cnt, reset=$reset"
+  if [ $undo_cnt -gt $ccnt ]; then
+    echo "Only $ccnt $cstr, cannot undo $undo_cnt"
+  elif [ "$type" = "hard" -a $ccnt -eq $undo_cnt ]; then
```
spacewander, line 17: Better to use `[ "$type" = "hard" ] && [ $ccnt -eq $undo_cnt ]`, which is more clear.

bin/git-undo
```diff
@@ -1,6 +1,28 @@
...
+    echo "Cannot hard undo all commits"
+  elif [ "$type" = "soft" -a $ccnt -eq 1 ]; then
```
spacewander, line 19: Ditto.

bin/git-undo
vr8hub: Yep, should have caught that. Thanks.

bin/git-undo
vr8hub: Doh! I looked at that right before my last round of testing and said, "Don't forget to get rid of that!" and promptly didn't get rid of that.

bin/git-undo
vr8hub: I kept them generic on purpose. The names in the function are what matter. If you would like to name the parameters, I'll be glad to use them.

bin/git-undo
vr8hub: Well, I disagree that it's clearer, but it's not a hill worth dying on, so I'll change it.

bin/git-undo
vr8hub: Ditto.

vr8hub:
Fixed everything but the parmX names. Again, I named them that way on purpose, but if you have suggestions I'll be glad to use them instead.

bin/git-undo
```diff
@@ -12,26 +33,40 @@ EOL
...
   *)
-    test $1 -gt 1 > /dev/null 2>&1 && back="~$1"
-    git reset --soft HEAD$back
-    git reset
+    parm1=soft
+    parm2="$1"
     ;;
 esac
+
+if [[ ! $parm2 =~ [1-9]*([0-9]) ]]; then
```
spacewander, line 67: I have changed my mind. Since `0` is not a valid count, maybe the pattern should be `[1-9][0-9]*`?

vr8hub: That's not really changing your mind, that's what you brought up initially, which is why I changed the first one to be 1-9 instead of 0-9. What I have does exactly what you're suggesting, using [[ regex syntax.

spacewander: @vr8hub
Err... This regex pattern is better: `if [[ ! $parm2 =~ ^[1-9][0-9]*$ ]]`.
It can detect all three invalid cases:`0`, `1s`, `01`.

vr8hub: Done.

spacewander:
@vr8hub 
Merged. Thanks!

PR #853, opened 2020-06-22
feat(browse): select remote of current branch, fallback to origin

bethesque:


bin/git-browse
```diff
@@ -2,7 +2,11 @@
 
 if [[ $1 == "" ]]
 then
-    remote=$(git remote | head -n 1)
+    remote=$(git remote | grep origin | head -n 1)
```
spacewander, line 5: Better to use `git config branch.$branch.remote || echo "origin"`, while the `$branch` is the current branch? So that we can get the remote of the current branch, or fallback to `origin`.

bethesque: Nice. I'll update the code.

spacewander:
@bethesque 
Will you continue to work on this PR?

bethesque:
Yes! I just haven't had my usual "train trip" OSS time recently due to some pesky pandemic. I'll see if I can get to it.

spacewander:
@bethesque 
Merged. Thanks!

bethesque:
Thanks for maintaining this handy tool.

PR #854, opened 2020-06-22
fix(browse): update sed pattern to work on mac

bethesque:
The `\w` character class on Mac's version of sed (BSD) does not seem to be recognised as "all word characters".

```
echo "foo" | sed -E -e 's/\w*/_/'
_foo
```

I had to update my pattern to `.*` to get it to work correctly. I understand it's not as precise, but I couldn't find a character class that was acknowledged to work the same across all platforms.

### DIFF
--- a/bin/git-browse
+++ b/bin/git-browse
@@ -22,7 +22,7 @@ fi
 
 if [[ $remote_url = git@* ]]
 then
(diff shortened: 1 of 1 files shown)


spacewander:
Merged. Thanks!

PR #859, opened 2020-07-24
Suggestions for git-sed

mxmehl:
I use this great script stand-alone and added some features. Perhaps not all feasible for upstream:

* Add [REUSE Software](https://reuse.software) identifiers to ease code reuse
* Add a link to the manpage, useful if used stand-alone
* Replace git_extra_mktemp by built-in function
* Add extended regex to grep and sed


mxmehl:
Is there something missing to get this Pull Request merged?

PR #867, opened 2020-08-21
Improved the 'git-info' command (color, submodule, config)

yuravg:
Hi,
I see [#755](https://github.com/tj/git-extras/pull/755), but did something similar in another way

added:
- info about submodules
- colorized headers
- extend info commit and repo configuration

note:
- changes checked in MinGW, Debian Linux
- I do not  update git-extras-completion.zsh (I use bash only, and do not have zsh)

I hope this will be useful

### DIFF
--- a/bin/git-info
+++ b/bin/git-info
@@ -48,7 +48,6 @@ fi
 
 echon "## Most Recent Commit:"
 echon "$(most_recent_commit)"
-echon "Type 'git log' for more commits, or 'git show <commit id>' for full commit details."
 
 if test "$1" != "--no-config"; then
   echon "## Configuration (.git/config):"
--- a/bin/git-info
+++ b/bin/git-info
@@ -21,11 +21,9 @@ remote_urls() {
 }
 
 echon() {
-    echo "$@"
-    echo
+  echo "$@"
+  echo
 }
-
-# Show info similar to svn
 
 echo
 echon "## Remote URLs:"
--- a/bin/git-info
+++ b/bin/git-info
@@ -1,5 +1,19 @@
 #!/usr/bin/env bash
 
+GREEN="$(tput setaf 2)"
+NORMAL="$(tput sgr0)"
+if [ "$1" = "--color" ] || [ "$2" = "--color" ] || \
+     [ "$1" = "-c" ] || [ "$2" = "-c" ] ; then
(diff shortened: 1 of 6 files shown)


bin/git-info
```diff
@@ -1,11 +1,26 @@
 #!/usr/bin/env bash
 
 get_config() {
-  git config --list
+  cmd_get_config="$(git config --get-all extras.info.config-grep)"
```
spacewander, line 4: Now we prefer to use `git-extras` as the namespace, like `git config --get git-extras.pr.remote`.

bin/git-info
```diff
@@ -21,27 +36,39 @@ remote_urls() {
...
+  echo "$@"
+  echo
 }
 
-# Show info similar to svn
+GREEN="$(tput setaf 2)"
+NORMAL="$(tput sgr0)"
+if [ "$1" = "--color" ] || [ "$2" = "--color" ] || \
+     [ "$1" = "-c" ] || [ "$2" = "-c" ] ; then
+  COLOR_TITLE="$GREEN"
```
spacewander, line 47: Better to mention that `-c` only affects the title in the documentation.

bin/git-info
```diff
@@ -21,27 +36,39 @@ remote_urls() {
...
+  echon "$SUBMODULES_LOG"
+fi
+
+echon "${COLOR_TITLE}## Most Recent Commit:${NORMAL}"
 echon "$(most_recent_commit)"
-echon "Type 'git log' for more commits, or 'git show <commit id>' for full commit details."
 
-if test "$1" != "--no-config"; then
-  echon "## Configuration (.git/config):"
+if [ "$1" != "--no-config" ] && [ "$2" != "--no-config" ]; then
```
spacewander, line 71: Better to move the argument checks together at the top of the script.

spacewander (changes requested):

bin/git-info
yuravg: Fixed

bin/git-info
yuravg: Fixed

bin/git-info
yuravg: Fixed

spacewander (approved):

spacewander:
@yuravg 
Merged. Thanks!

PR #880, opened 2020-09-24
git-info: fix regex for bsd sed

spacewander:


SuperSandro2000:
The patch does not break gnu sed.

PR #889, opened 2020-11-07
feat(delete-tag): get default remote from git config if exists

### LINKED ISSUE #870: cmd `delete-tag` assumes "origin" as remote for tags
`delete-tag` the pushes the against the hard-coded remote 'origin' .

This failed for me b/c i renamed this usually to 'o' for laziness' sake ;)

`delete branch` figures the remote out from the branche's remote which doens't work for tags.
However, the current branch's remote could be used as a fall back which to me would make sense in most cases.

Another idea were to make the remote a CLI option or could be even put into the git config.
Note sure if this issue comes up in other extra commands as well in which case a consistent behavior across all commands would be appreciated

vanpipy:
fix #870

### DIFF
--- a/bin/git-delete-tag
+++ b/bin/git-delete-tag
@@ -3,6 +3,15 @@
 # Assert there is at least one tag provided
 test -z $1 && echo "tag required." 1>&2 && exit 1
 
+# Detect the defalt remote exists or not
+defualt_remote=$(git config git-extras.default-remote)
+
+if [[ -n "$defualt_remote" ]]; then
+    origin="$defualt_remote"
+else
+    origin=origin
+fi
+
 # Concatenate all the tag references
 local_tags=""
 origin_refs=""
@@ -14,4 +23,4 @@ done
 
 # Delete all the tags
 git tag -d $local_tags
-git push origin $origin_refs
+git push $origin $origin_refs


bin/git-delete-tag
```diff
@@ -3,6 +3,16 @@
...
 test -z $1 && echo "tag required." 1>&2 && exit 1
 
+# Set the default origin
+origin="origin"
+
+# Detect the defalt remote exist or not
+defualt_remote=$(git config git-extras.default-remote)
+
+if [[ ! -z "$defualt_remote" ]]; then
+    origin="$defualt_remote"
```
SuperSandro2000, line 13:
```suggestion
default_remote=$(git config git-extras.default-remote)

if [[ -n "$default_remote" ]]; then
    origin="$default_remote"
else
   origin=origin
fi
```

bin/git-delete-tag
```diff
@@ -3,6 +3,16 @@
 # Assert there is at least one tag provided
 test -z $1 && echo "tag required." 1>&2 && exit 1
 
+# Set the default origin
+origin="origin"
+
+# Detect the defalt remote exist or not
```
SuperSandro2000, line 9:
```suggestion
# Detect the defalt remote exists or not
```

bin/git-delete-tag
```diff
@@ -3,6 +3,16 @@
 # Assert there is at least one tag provided
 test -z $1 && echo "tag required." 1>&2 && exit 1
 
+# Set the default origin
+origin="origin"
+
```
SuperSandro2000, line 8:
```suggestion
```

bin/git-delete-tag
vanpipy: Update.

bin/git-delete-tag
vanpipy: Correct the typo and update

bin/git-delete-tag
vanpipy: Move the `origin` into the `else` and update.

PR #894, opened 2020-11-16
Merge-into: add stash before to checkout

alessandro308:
Sometimes, if you have dirty env, git checkout fails due to unstashed files.


spacewander:
@alessandro308 
Merged. Thanks!

PR #895, opened 2020-11-18
feat(#861): add optional parameter --from to set the start point

### LINKED ISSUE #861: Add start-point option to `git feature`
Currently, `git feature` creates a new branch based on the current active one. 

As `git flow` specifies that feature branches should be based on `develop`, it would be nice to be able to specify a `start-point`, so we don't need to switch branches first.

After finishing a feature, I currently need to run 3 commands to create my new feature branch:

```bash
git checkout develop
git pull
git feature my-feat
```

It would be awesome if we could simplify those 3 lines to a single command:

```bash
# git feature <branchname> [<start-point>]
  git feature my-feat develop 
```

Regardless of my active branch, `git feature my-feat develop` would:

- pull/fetch the commits from `origin`
- create a new branch with `develop` as `start-point`, 
- check out the new feature branch.

Does `git-extras` have global options? It would be awesome to be able to set the default `start-point` trough a global option

vanpipy:
Close #861

### DIFF
--- a/bin/git-create-branch
+++ b/bin/git-create-branch
@@ -20,6 +20,12 @@ do
 				REMOTE=origin
 			fi
 			;;
+		--from)
+			if [[ -n $2 ]]; then
+				START_POINT=$2
+				shift
+			fi
+			;;
 		*)
 			BRANCH=$1
 	esac
@@ -44,9 +50,15 @@ then
 	rm -f "$stderr"
 	if [ $REMOTE_EXIT -eq 0 ]
 	then
-		git push $REMOTE HEAD:refs/heads/$BRANCH
-		git fetch $REMOTE
-		git checkout --track -b $BRANCH $REMOTE/$BRANCH
+		if [[ -n $START_POINT ]]; then
+			git fetch $REMOTE
+			git checkout --track -b $BRANCH $START_POINT
+			git push $REMOTE HEAD:refs/heads/$BRANCH
+		else
+			git push $REMOTE HEAD:refs/heads/$BRANCH
+			git fetch $REMOTE
+			git checkout --track -b $BRANCH $REMOTE/$BRANCH
+		fi
 		exit $?
 	else
 		echo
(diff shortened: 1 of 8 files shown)


bin/git-create-branch
```diff
@@ -20,6 +20,13 @@ do
 				REMOTE=origin
 			fi
 			;;
+		--from)
+			if [[ -n $2 ]]
+			then
```
SuperSandro2000, line 25:
```suggestion
			if [[ -n $2 ]]; then
```

bin/git-create-branch
```diff
@@ -44,9 +51,16 @@ then
 	rm -f "$stderr"
 	if [ $REMOTE_EXIT -eq 0 ]
 	then
-		git push $REMOTE HEAD:refs/heads/$BRANCH
-		git fetch $REMOTE
-		git checkout --track -b $BRANCH $REMOTE/$BRANCH
+		if [[ -n $START_POINT ]]
+		then
```
SuperSandro2000, line 55:
```suggestion
		if [[ -n $START_POINT ]]; then
```

SuperSandro2000 (commented):
you get the gist.

bin/git-create-branch
vanpipy: Update.

bin/git-create-branch
vanpipy: Update.

And a question I want to ask is how to choose the type of using the style of the `if then`:

Style A:
```
if [[ ... ]]; then
   ...
fi
```

Style B:
```
if [[ ... ]]
then
    ...
fi
```

Both of them exist in the file. That confuse me a little.

Thanks :D.

spacewander:
@vanpipy 
Would you update the doc?

vanpipy:
> @vanpipy
> Would you update the doc?

Hi, @spacewander ,

The git-create-branch and git-feature docs will update at the next commit.

spacewander (approved):

PR #896, opened 2020-11-26
fix: zsh completion of git-coauthor command

Aloxaf:


spacewander:
@Aloxaf 
Merged. Thanks!

PR #929, opened 2021-06-04
Improve branch name handling in git-delete-merged-branches

nkiesel:
We have branches whose name contain either svn or the name or extra default branch (e.g. svnner).  Thus,
improving the filter to only drop exact matches.


bl-ue:
The proble with this patch is that `-P` is not supported on macOS's builtin `grep` (which is BSD) 🤔

PR #934, opened 2021-08-07
No merge summary

tfendin:
Added option --no-merges to git summary, which excludes merges when counting everything. I've only tested it on Linux, I'd appreciate if someone could help with testing on OSX and BSD.

### DIFF
--- a/man/git-summary.md
+++ b/man/git-summary.md
@@ -3,7 +3,8 @@ git-summary(1) -- Show repository summary
 
 ## SYNOPSIS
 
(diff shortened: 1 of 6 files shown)


man/git-summary.md
```diff
@@ -40,6 +40,10 @@ Shows a summary of the repository.
   Summarize with lines other than commits.
   When `--line` is specified, the last argument is treated as &lt;path&gt;.
 
+  --no-merges
+
+  Exclude merge commits.
```
spacewander, line 45: Should mention "This option can not be used together with --line"

spacewander (approved):

PR #936, opened 2021-08-11
Format fix of Synopsis in manuals

tfendin:
Fixed issues where the synopsis was split in to several paragraphs, or when several command variants was garbled into one line.

### DIFF
--- a/man/git-alias.md
+++ b/man/git-alias.md
@@ -3,9 +3,9 @@ git-alias(1) -- Define, search and show aliases
 
 ## SYNOPSIS
 
-`git-alias`
-`git-alias` &lt;search-pattern&gt;
-`git-alias` &lt;alias-name&gt; &lt;command&gt;
+`git-alias`  
+`git-alias` &lt;search-pattern&gt;  
+`git-alias` &lt;alias-name&gt; &lt;command&gt;  
 
 ## DESCRIPTION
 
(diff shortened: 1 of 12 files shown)


spacewander:
@tfendin 
Could you make a shorter title to avoid breaking in the middle?

tfendin:
@spacewander 
Is this title better?

man/git-alias.md
```diff
@@ -3,9 +3,9 @@ git-alias(1) -- Define, search and show aliases
 
 ## SYNOPSIS
 
-`git-alias`
-`git-alias` &lt;search-pattern&gt;
-`git-alias` &lt;alias-name&gt; &lt;command&gt;
+`git-alias`
```
spacewander, line 6: Is this trailing whitespace necessary?

tfendin: Yes, that's is the markdown syntax for a line break.

spacewander: Would it be better to use multiple newlines?

tfendin: No, that would create an empty line between the different command variants. Which is not the style used by other commands in git-extras (git bulk for example) or in git-core (git stash for example). I'd think it was good if the same style was used in git-extras as in git-core etc.

spacewander (approved):

PR #937, opened 2021-08-12
Compare email adresses case insensitive in --dedup-by-email

tfendin:
SMTP adresses are case insensitive, so --dedup-by-email should treat them as that.

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -87,11 +87,11 @@ dedup_by_email() {
     LC_ALL=C awk '
     {
         sum += $1
-        lastField = tolower($NF)
-        if (lastField in emails) {
-            emails[lastField] += $1
+        last_field = tolower($NF)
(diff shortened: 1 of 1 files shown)


bin/git-summary
```diff
@@ -80,17 +80,18 @@ file_count() {
...
     {
         sum += $1
-        if ($NF in emails) {
-            emails[$NF] += $1
+        lastField = tolower($NF)
+        if (lastField in emails) {
+            emails[lastField] += $1
         } else {
-            email = $NF
+            email = lastField
```
spacewander, line 94: Style: use underscore style variable

tfendin: Fixed. Perhaps the coding style could be mentioned in CONTRIBUTING.md.

spacewander (approved):

PR #944, opened 2021-10-29
git-release: Add prefix to semver

SimonTate:
To allow different release names other than specifically mm.vv.pp, this
adds the ability to use a prefix too.

For example, you can use:

```
$ git release --semver minor --prefix "prefix-"
```

This would allow the previous tags to be in the form `prefix-0.1.0`. A more obvious use case for this would be a single letter, such as `v` or `r` to use tags like: `r0.1.0` or `v0.1.0`.

### DIFF
--- a/bin/git-release
+++ b/bin/git-release
@@ -51,6 +51,12 @@ if test $# -gt 0; then
       semver=$2
       shift
       ;;
+    --prefix)
+      test -z "$2" &&
+          exit_with_msg "prefix string required for --prefix option"
+      prefix="$2"
+      shift
+      ;;
     --no-empty-commit) no_empty_commit=true;;
     --) shift; hook_args="$hook_args $*"; break;;
     *) test -z "$version" && version=$1 ;;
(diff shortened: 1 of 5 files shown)


man/git-release.1
```diff
@@ -1,13 +1,13 @@
...
 .
-.TH "GIT\-RELEASE" "1" "April 2018" "" "Git Extras"
+.TH "GIT\-RELEASE" "1" "October 2021" "" "Git Extras"
 .
 .SH "NAME"
 \fBgit\-release\fR \- Commit, tag and push changes to the repository
 .
 .SH "SYNOPSIS"
-\fBgit\-release\fR [<tagname> | \-\-semver <name>] [\-r <remote>] [\-m <commit info>] [\-\-no\-empty\-commit] [\-c] [\-s] [\-u <key-id>] [[\-\-] <hook arguments\.\.\.>]
+\fBgit\-release\fR [<tagname> | \-\-semver <name>] [\-r <remote>] [\-m <commit info>] [\-\-no\-empty\-commit] [\-c] [\-s] [\-u <key\-id>] [[\-\-] <hook arguments\.\.\.>]
```
spacewander, line 10: The change of markdown is not submitted?

SimonTate: should be corrected now @spacewander

bin/git-release
```diff
@@ -51,6 +51,12 @@ if test $# -gt 0; then
       semver=$2
       shift
       ;;
+    --prefix)
+      test -z "$2" &&
+          exit_with_msg "prefix string required for --prefix option"
+      prefix=$2
```
spacewander, line 57:
```suggestion
      prefix="$2"
```

SimonTate: Suggestion taken in latest commit

spacewander (approved):

Issue #476, opened 2015-10-20
ignore: need to ensure a linebreak


jankatins:
If `.gitignore` does not include a linebreak (because I forgot...), the `git ignore xxx/` will include the new ignore pattern at the end _without_ a leinreak inbetween:

This `.gitignore`:

```
pattern/
# no linebreak at the end of the file
```

```
git ignore build/
```

will result in:

```
pattern/build/
# linbreak
```

tfendin:
This was fixed in #949. Could be closed

spacewander:
Thanks for @tfendin!

PR #949, opened 2021-12-29
git-ignore: ensure new patterns comes after a newline

tfendin:
As per #476

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -57,6 +57,10 @@ function add_private {
 function add_patterns {
   echo "Adding pattern(s) to: $1"
   local file="${1/#~/$HOME}"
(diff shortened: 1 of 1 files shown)


bin/git-ignore
```diff
@@ -57,6 +57,10 @@ function add_private {
 function add_patterns {
   echo "Adding pattern(s) to: $1"
   local file="${1/#~/$HOME}"
+  if [ -s "$file" ]; then
+      # If $file doesn't end with a newline, add one
```
spacewander, line 61:
```suggestion
      # If the content of $file doesn't end with a newline, add one
```

spacewander (approved):

PR #951, opened 2022-01-03
Made local-commits work with all remotes

tfendin:
The previous version required that the remote branch was in the origin remote. This version uses the upstream branch instead, so it works with any remote. However, it fails if there is no upstream configured. If required, this could be fixed with some checks (i.e. revert to the previous behavior if no upstream is configured.)


spacewander (commented):
Let's update the doc too

spacewander (approved):

PR #959, opened 2022-03-20
feat: implement git-browse-ci

pbnj:
This is a proposal and an implementation to open repo's CI page (e.g. GitHub Actions, GitLab CI, Bitbucket Pipelines) in default browser.

Feedback is welcome, even if feedback is "thanks, but no thanks".

### DIFF
--- a/bin/git-browse-ci
+++ b/bin/git-browse-ci
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+
+if [[ $1 == "" ]]
+then
+    branch=$(git rev-parse --abbrev-ref HEAD 2&> /dev/null)
+    remote=$(git config branch."${branch}".remote || echo "origin")
+else
+    remote=$1
+fi
+
+if [[ $remote == "" ]]
+then
+    echo "Remote not found"
+    exit 1
+fi
+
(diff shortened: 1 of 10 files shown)


bin/git-browse-ci
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
```
spacewander, line 1: Let's add `set -e -o pipefail`

pbnj: Done.

bin/git-browse-ci
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e -o pipefail
```
spacewander, line 3: I get 128 exit code when no upstream is given after this modification...

pbnj: It seems to be the combination of `set -e` and `git rev-parse --abbrev-ref HEAD 2&> /dev/null` (specifically the `2&> /dev/null`). 

Changing it to the shorthand version `&> /dev/null` or the longhand version `>/dev/null 2>&1` fixes it for me.

spacewander (approved):

PR #969, opened 2022-05-01
`git-utimes`: add --newer flag and don't touch files in the working tree and index

wpwoodjr:
The PR adds a new flag, `-n` or `--newer`, to `git-utimes`.  This flag preserves the original modification date of files that originally were committed from the local repo, where the file may have been modified well before the commit date.  The `--newer` flag causes `git-utimes` to only touch files with a date that is newer than their commit date.

Additionally, this PR addresses a few issues:

1. `git-utimes` would change the date of files in the working tree and index to their commit date.  This version doesn't touch files in the working tree and index, on the assumption that they are being worked on and the commit date is not relevant.
2. `git-utimes` would attempt to change the modification date on symbolic links.  This was incorrectly changing the date on the linked file, not the symbolic link.  This version correctly updates the link itself.
3. `git-utimes` was using the processor count to determine how many files `xargs` would process on each command line (`-n` argument to `xargs`), which didn't seem to make much sense.  This version passes 24 files on each `xargs` command line.
4. `git-utimes` was always touching files even if the modification date was the same as the commit date.  After a pull, even files that were not modified by the pull would be touched.  This version only touches files whose modification date is not the same as the commit date.  This reports only updated files and makes it easier to understand which files have been changed by a pull.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -1520,12 +1520,14 @@ $ git browse-ci upstream
 
 ## git utimes
 
-Change files modification time to their last commit date.
+Change files modification time to their last commit date.  Does not touch files that are in the working tree or index.
+
+The `-n` or `--newer` flag preserves the original modification time of files that were committed from the local repo, by only touching files that are newer than their last commit date.
 
 ```bash
 git-extras$ ls -l bin | head
 total 308
--rwxr-xr-x 1 vt vt   489 Nov  8 13:56 git-alias
+-rwxr-xr-x 1 vt vt   489 Jul 28  2015 git-alias
 -rwxr-xr-x 1 vt vt  1043 Nov  8 13:56 git-archive-file
 -rwxr-xr-x 1 vt vt   970 Nov  8 13:56 git-authors
 -rwxr-xr-x 1 vt vt   267 Nov  8 13:56 git-back
@@ -1534,8 +1536,7 @@ total 308
 -rwxr-xr-x 1 vt vt  6282 Nov  8 13:56 git-bulk
 -rwxr-xr-x 1 vt vt 18561 Nov  8 13:56 git-changelog
 -rwxr-xr-x 1 vt vt   215 Nov  8 13:56 git-clear
-git-extras$ git utimes
-+ touch -d 2015-08-09T19:27:49+08:00 bin/git-alias
+git-extras$ git utimes --newer
 + touch -d 2020-05-22T10:40:29+08:00 bin/git-archive-file
 + touch -d 2017-05-05T16:02:09+08:00 bin/git-authors
 + touch -d 2020-02-23T11:41:54+08:00 bin/git-back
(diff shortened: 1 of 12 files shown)


Commands.md
```diff
@@ -1520,12 +1520,14 @@ $ git browse-ci upstream
 
 ## git utimes
 
-Change files modification time to their last commit date.
+Change files modification time to their last commit date.  Does not touch files that are in the working tree or index.
```
spacewander, line 1523:
```suggestion
Change files modification time to their last commit date. Does not touch files that are in the working tree or index.
```

Commands.md
```diff
@@ -1520,12 +1520,14 @@ $ git browse-ci upstream
...
+
+The `-n` or `--newer` flag preserves the original modification time of files that were committed from the local repo, by only touching files that are newer than their last commit date.
```
spacewander, line 1525: Personally speaking, we can only use `--newer`? The shorter version `-n` might be confusing.

bin/git-utimes
```diff
@@ -25,6 +33,12 @@ else
   # because all args will go into single worker.
   NPROC=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || getconf _NPROCESSORS_ONLN 2>/dev/null)
   # shellcheck disable=SC2086
-  git ls-tree -z -r -t --name-only @ \
-  | xargs -0 -P${NPROC:-1} -n${NPROC:-1} $opt_r git utimes --touch
+  # don't touch files that have been modified in the worktree or index
+  #   bsd doesn't have `-z` option for `comm` and `cut`, so use `tr` as work around
+  prefix="$(git rev-parse --show-prefix) "
+  comm -23 <(git ls-tree -z -r --name-only --full-name @ | tr '\0' '\n') \
```
spacewander, line 39: I get `comm: file 1 is not in sorted order` when running the command. Maybe something is missing?

wpwoodjr: OK, I had thought that git ls-tree was always sorted... Let me fix it and add your other suggestions and re-commit.

Commands.md
wpwoodjr: OK

wpwoodjr:
@spacewander  I have updated per your suggestions.  I also added a hint where to find the `column` dependency :)

bin/git-utimes
```diff
@@ -2,20 +2,28 @@
...
-    iso_t=$(git --no-pager log --no-renames --pretty=format:%cI -1 @ -- "$f" 2>/dev/null)
-    if [ -n "$iso_t" ]; then
-      bsd=$(date -j > /dev/null 2>&1 && echo 'true')
-      if [ "$bsd" == "true" ]; then
-        t=$(date -j -f %Y-%m-%dT%H:%M:%S "$iso_t" +%Y%m%d%H%M.%S)
-      else
-        t=$(date -d"$iso_t" +%Y%m%d%H%M.%S)
+    git_s=$(git --no-pager log --no-renames --pretty=format:%ct -1 @ -- "$f" 2>/dev/null)
+    mod_s=$(stat $stat_flags "$f" 2>/dev/null)
+    if [ -n "$git_s" ] && [ -n "$mod_s" ] && [ $mod_s -ne $git_s ]; then
```
spacewander, line 21: Would be better to double-quote every var

wpwoodjr: git_s and mod_s are integers but OK, done.

spacewander (approved):

PR #982, opened 2022-06-25
Add git-pseudo-version

chenlujjj:
It can be used to print pesudo-version of go module of current repository.


guenhter:
It seems like this is not exaclty what go defines as pseudo version:

https://go.dev/ref/mod#pseudo-versions

> Each pseudo-version has three parts ....

guenhter:
And shouldn't this be something which should go more into a go-module instead of the git-module?

PR #988, opened 2022-07-29
fix(git-cp): keep the history of the files

### LINKED ISSUE #985: `git-cp` does not preserve file history
The [man page](https://www.mankier.com/1/git-cp) states that using `git-cp` will "Copy a file keeping its history", however, the history is not preserved.

Repro steps:
1. Use git-cp to copy an existing file, e.g. `git cp 
2. Do `git diff HEAD^` - see that git considers the entire copied file to be completely new code authored by the current user

As you can see in the [source code](https://github.com/tj/git-extras/blob/master/bin/git-cp), rather than following the [process](https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history/44036771#44036771) referenced by the [man page](https://www.mankier.com/1/git-cp#See_Also), the current implementation just simply copies the file using `cp -r` and adds a commit without any steps to keep the git file history.

vanpipy:
fix #985 

The implement https://github.com/tj/git-extras/commit/91b3e47243b8c1ab549cb24a9271fda4a6fdd911 gives a great idea to show the way to reach the goal. I just renew the way to show it clearly.

### DIFF
--- a/bin/git-cp
+++ b/bin/git-cp
@@ -32,9 +32,36 @@ elif [[ "$CURRENT_FILENAME" == "" ]]; then
     usage
     exit 10  # Missing arguments CURRENT_FILENAME
 else
-    echo "Coping $CURRENT_FILENAME into $DESTINATION_FILENAME"
-    cp -r "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
+    if [ -e "$DESTINATION_FILENAME"  ]; then
+        echo 1>&2 "$DESTINATION_FILENAME already exists."
+        echo 1>&2 "Make sure to remove the destination first."
+        exit 40
+    fi
 
-    git add "$DESTINATION_FILENAME"
-    git commit -m "Duplicate ${CURRENT_FILENAME} history in ${DESTINATION_FILENAME} history." -n
+    echo "Coping $CURRENT_FILENAME into $DESTINATION_FILENAME"
+
+    INTERMEDIATE_FILENAME="${CURRENT_FILENAME//\//__}-move-to-${DESTINATION_FILENAME//\//__}"
+
+    # We keep the existing file on the side in a commit
+    git mv "${CURRENT_FILENAME}" "${INTERMEDIATE_FILENAME}"
+    git commit -nm "Keep $CURRENT_FILENAME"
+
+    # We come back to the previous state and revert that change
+    INTERMEDIATE_SAVED=$(git rev-parse HEAD)
+    git reset --hard HEAD^
+
+    # We move the file to its new destination
+    git mv "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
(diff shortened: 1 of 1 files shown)


bin/git-cp
```diff
@@ -32,9 +32,36 @@ elif [[ "$CURRENT_FILENAME" == "" ]]; then
...
     echo "Coping $CURRENT_FILENAME into $DESTINATION_FILENAME"
-    cp -r "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
 
-    git add "$DESTINATION_FILENAME"
-    git commit -m "Duplicate ${CURRENT_FILENAME} history in ${DESTINATION_FILENAME} history." -n
+    INTERMEDIATE_FILENAME="${CURRENT_FILENAME//\//__}-move-to-${DESTINATION_FILENAME//\//__}"
+
+    # We keep the existing file on the side in a commit
+    git mv "${CURRENT_FILENAME}" "${INTERMEDIATE_FILENAME}"
+    git commit -a -nm "Keep $CURRENT_FILENAME"
```
spacewander, line 47: Why do we need to use `-a` after `git mv`?

vanpipy: It is the test to avoid interruption because the original implementation was interrupted by conflict always. At first, it confused me and I thought something missing maybe, so the option `-a` was added. The option `-a` is useless for the current implementation and I forget to remove it. Fixed already.

spacewander (commented):
I tried the PR and found out it will create multiple commits for a `cp` operation? The previous version only created one.

vanpipy:
> I tried the PR and found out it will create multiple commits for a `cp` operation? The previous version only created one.

Keeping only one commit will overwrite the history always and I did not find a better way to keep the history and only one commit. But i think keeping the history of the file is meaningful. That is the reason the PR comes out. If the history did not keep anymore, the document of the `git-cp` needs to sync the action.

spacewander (commented):
I run `git cp AUTHORS AUTHORS.txt` with the new git-cp, but the `AUTHORS.txt` doesn't have the same history as `AUTHORS`.

vanpipy:
> I run `git cp AUTHORS AUTHORS.txt` with the new git-cp, but the `AUTHORS.txt` doesn't have the same history as `AUTHORS`.

I tried and had seen the history, but there is one difference - one more column `original file` is added to the history of the file copied.

diff

The git version I used is 2.25.1

spacewander:
> > I run `git cp AUTHORS AUTHORS.txt` with the new git-cp, but the `AUTHORS.txt` doesn't have the same history as `AUTHORS`.
> 
> I tried and had seen the history, but there is one difference - one more column `original file` is added to the history of the file copied.
> 
> diff
> 
> The git version I used is 2.25.1

Ah, it is my mistake.

spacewander (approved):

Issue #979, opened 2022-05-31
git-effort:fork: retry: Resource temporarily unavailable


Lonenso:
I just heard git-extras this powerful tool, and try to use `git effort` and encounter this error.

```
❯ ulimit -a
-t: cpu time (seconds)              unlimited
-f: file size (blocks)              unlimited
-d: data seg size (kbytes)          unlimited
-s: stack size (kbytes)             8192
-c: core file size (blocks)         0
-v: address space (kbytes)          unlimited
-l: locked-in-memory size (kbytes)  unlimited
-u: processes                       2784
-n: file descriptors                65535

❯ launchctl limit
	cpu         unlimited      unlimited
	filesize    unlimited      unlimited
	data        unlimited      unlimited
	stack       8388608        67104768
	core        0              unlimited
	rss         unlimited      unlimited
	memlock     unlimited      unlimited
	maxproc     2784           4176
	maxfiles    65535          65535

```
macOS Monterey: 12.3.1, intel
git-extra: 6.4.0

since I'm using this command on a private repo, I find another [repo](https://github.com/pubnub/objective-c) to reproduce this.

I'm happy to provide more details if you guys need.

spacewander:
```
maxproc     2784           4176
```
I guess it is because the number of processes is not enough, as currently `git effort` will create a background process per path.

spacewander:
PR is welcome!

Lonenso:
since the PR is merged, close this issue.

PR #990, opened 2022-09-16
fix(git-effort):fork: retry: Resource temporarily unavailable #979

Lonenso:
- use wait -n to limit number of processes

requirements:
- OSX, Linux, BSD (You may need to browse their man page)*: 
  - use `nproc --all`, `getconf NPROCESSORS_ONLN`, `getconf __NPROCESSORS_ONLN`, `grep -c processor /proc/cpuinfo` to get processor counts.
  - provide a fallback value `8`
- Bash 3.2+ (If you aren't sure, see [the Bash changelog](https://www.tldp.org/LDP/abs/html/bash2.html)): 
  - `wait -n` was introduced in Bash 4.0

### DIFF
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -140,10 +140,12 @@ sort_effort() {
 }
 
 #
-# get processor count
+# get processor count, default 8
 #
 procs() {
-  (nproc --all || getconf NPROCESSORS_ONLN || getconf _NPROCESSORS_ONLN || grep -c processor /proc/cpuinfo) 2>/dev/null
+  if ! (nproc --all || getconf NPROCESSORS_ONLN || getconf _NPROCESSORS_ONLN || grep -c processor /proc/cpuinfo) 2>/dev/null; then
+      echo 8
+  fi
 }
 
 declare -a paths=()
@@ -214,7 +216,7 @@ heading
 
 # send paths to effort
 nPaths=${#paths[@]}
-nProcs=${$(procs):-8}
+nProcs=$(procs)
 nJobs="\j"
 for ((i=0; i<nPaths; ++i))
 do
--- a/bin/git-effort
+++ b/bin/git-effort
@@ -139,6 +139,13 @@ sort_effort() {
   < $tmp sort -rn -k 2
 }
 
+#
+# get processor count
+#
+procs() {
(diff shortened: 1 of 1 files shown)


Lonenso:
@spacewander  Hi, could you help to review this? thanks.

bin/git-effort
```diff
@@ -207,11 +207,15 @@ heading
 
 # send paths to effort
 nPaths=${#paths[@]}
+nProcs=$(getconf NPROCESSORS_ONLN)
```
spacewander, line 210: I get `getconf: Unrecognized variable `NPROCESSORS_ONLN'` when running this command. We should provide a fallback for it.

Lonenso: Sure, since I can not find a perfect command to support all distributions, platforms, I simply choose 8 as a magic number.

bin/git-effort
```diff
@@ -207,11 +214,15 @@ heading
 
 # send paths to effort
 nPaths=${#paths[@]}
+nProcs=${$(procs):-8}
```
spacewander, line 217: I got `${$(procs):-8}: bad substitution` error here. Maybe we can move the fallback code to function `procs`?

Lonenso: > I got ${$(procs):-8}: bad substitution error here
It seems you're using `sh` instead of `bash`. Could you tell me which shell are you using? 
> Maybe we can move the fallback code to function procs?
Sure.

spacewander: My bash doesn't work with this...
```
 ￥ bash -c 'nProcs=${$(procs):-8}'
bash: ${$(procs):-8}: bad substitution
 ￥ bash --version
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

spacewander (approved):

bin/git-effort
Lonenso: reproduced and fixed.

Lonenso: Before I merged this, since I used `wait -n` which was introduced in bash 4.0, should I change the CONTRIBUTING.md?

Issue #938, opened 2021-09-05
feature request: git summary --line by file extension


chanansh:
currently, git summary counts all files. Sometimes large files leak into the repo and I would like to filter by file extension, e.g. .py

tfendin:
@chanansh Doesn't `git summary --line *.py` give you the expected result? If not, please elaborate.

vanpipy:
There is an interesting question.

what is the result of the `git summary --line` with a pattern like `*.py`?

Following the example of a `git summary --line`, there are two keywords - `lines` and `authors`.

So if the `*.py` does it work and the result should appear in the `lines` and `authors` together.

PR #1005, opened 2022-11-15
Make git-summary accept multiple paths for the --lines argument

aronhoff:
Currently only the first provided path is processed (`$1`). Therefore it gives incorrect results when multiple paths are given, or even for single bash globs such as `*.py`.

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -41,8 +41,12 @@ if [ -n "$MERGES_ARG" ] && [ -n "$SUMMARY_BY_LINE" ]; then
 fi
 
 
-commit=""
-test $# -ne 0 && commit=$*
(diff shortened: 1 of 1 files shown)


aronhoff:
Fixes #938

bin/git-summary
```diff
@@ -167,18 +171,14 @@ single_file() {
 # list the author for all file
 #
 lines() {
-  if [ -n "$1" ]; then
-    git ls-files -- $1 | single_file
-  else
-    git ls-files | single_file
-  fi
+  git ls-files -- $@ | single_file
```
spacewander, line 174: Could you add a comment to disable shellcheck warning?

Issue #1007, opened 2022-11-30
Feature: Allow to combine `git bulk` with other commands of the git-extras


guenhter:
It would be nice if this would work:

```
git bulk summary
```

Currently if this is executed, the summary of only the very first found git repo is printed.

spacewander:
PR is welcome!

PR #1008, opened 2022-12-01
Allow to run git-extras within git-bulk

guenhter:
When you execute

```
git bulk summary
```

it only summarizes the first found repo. 
The reason for it seems to be some stdin ... read ... blah stuff (https://stackoverflow.com/a/13800476/5247292) 

Therefore I changed the way the single repos of the bulk operation are worked off to avoid stdin when in the loop

Fix https://github.com/tj/git-extras/issues/1007

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -133,12 +133,15 @@ function executBulkOp () {
     eval cd "\"$rwsdir\""
     local actual=$(pwd)
     [ "${quiet?}" != "false" ] && echo 1>&2 "Executing bulk operation in workspace ${inverse}$actual${reset}"
-    eval find -L . -name ".git" | while read line; do
+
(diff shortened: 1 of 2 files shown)


spacewander:
It seems that the `summary` only shows part of the data. Running the command on my machine, got:
```
 git bulk -a summary
Core command "summary" accepted.
-> executing git summary

 project  : git-extras
 repo age : 12 years
 active   : 795 days
 commits  : 1688
 files    : 315
 authors  :
```

guenhter:
Hm, interesting.

guenhter:
Seems like I fall into the trap of `git shortlog`

```
If no revisions are passed on the command line and either standard input is not a terminal or there is no current branch, git shortlog will output a summary of the log read from standard input, without reference to the current repository.
```

I fixed that in the git-extras by passing always a reference to the git shortlog

PR #1016, opened 2023-01-06
chore: More Bash improvements and removing deprecations

hyperupcall:
- Improve quoting, especially in places where filepaths are used
- No longer use `which`, which is deprecated in Debian and has issues - use `command -v` instead
  - This ends up **modifying** `brew-release.path` to match (just to note)
- No longer use deprecated `egrep`
- No longer use deprecated `[ -o` or ``

### DIFF
--- a/bin/git-changelog
+++ b/bin/git-changelog
@@ -550,7 +550,7 @@ main() {
   fi
 
   if [[ -z "$changelog" ]]; then
-    changelog="$(ls | egrep 'change|history' -i | head -n1)"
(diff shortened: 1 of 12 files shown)


spacewander:
Have you tested if `command -v` works with a fresh Mac?

hyperupcall:
The usage is specified in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799.2018edition). And I have used it before, passing Mac CI on my other Bash apps 

For your benefit I just tested it with Bash 3.2 on Catalina and it works flawlessly

spacewander (approved):

PR #1020, opened 2023-01-12
Make the bulk output slimmer

guenhter:
Combine two output lines of the git-bulk command into one for the non-guarded mode.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -59,19 +59,19 @@ function listall { git config --global --get-regexp bulkworkspaces; }
 
 # guarded execution of a git command in one specific repository
 function guardedExecution () {
-  if $guardedmode; then
-    echo 1>&2 -n "${inverse}git $gitcommand${reset} -> execute here (y/n)? "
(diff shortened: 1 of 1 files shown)


bin/git-bulk
```diff
@@ -141,7 +142,6 @@ function executBulkOp () {
       eval cd "\"$gitrepodir\"" # into git repo location
       local curdir=$(pwd)
       local leadingpath=${curdir#${actual}}
-      [ "${quiet?}" != "true" ] && echo 1>&2 "Current repository: ${leadingpath%/*}/${bldred}${curdir##*/}${reset}"
```
spacewander, line 144: Why remove this prompt?

guenhter: It is included now in the the line where it prints: "Executing git something". Now it it

"Executing git something in repository yourrepo"

spacewander: I see

spacewander:
Let's merge master to solve the conflict.

spacewander (approved):

PR #1021, opened 2023-01-14
chore: Fix more Bash inconsistencies

hyperupcall:
This mainly fixes the following:
- Not only are more path-related variables quoted (ie. `git-touch`, `git-scp`), but variables like `$BRANCH` are.
- Add more `-r` to `read` invocations
- Stuff like `$(echo $(printf '%s\n' 'thing'))` is changed to `$(printf '%s\n' 'thing')`

### DIFF
--- a/bin/git-psykorebase
+++ b/bin/git-psykorebase
@@ -19,7 +19,7 @@ function usage()
     echo "  -c|--continue:		Continue after the user updates conflicts."
 }
 
-while [[ $# > 0 ]]
+while [[ $# -gt 0 ]]
 do
   key="$1"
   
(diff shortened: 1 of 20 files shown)


bin/git-psykorebase
```diff
@@ -19,7 +19,7 @@ function usage()
     echo "  -c|--continue:		Continue after the user updates conflicts."
 }
 
-while [[ $# > 0 ]]
+while [[ $# -gt 0 ]]
```
guenhter, line 22: Why was this changed? `>` is be perfectly and expressive find within `[[ ... ]]`, isn't it?

hyperupcall: `>` and `<` perform [lexicographical comparison](https://www.shellcheck.net/wiki/SC2071) rather than a numerical comparison in this context, so it was changed to `-gt`

guenhter: I see. Thx for the explanation. With this new knowledge I now have to have a look at some of my own scripts as well :)

spacewander (approved):

PR #1022, opened 2023-01-14
Add EditorConfig file

hyperupcall:
While working on #1021, I noticed that some files like [git-delta](https://github.com/tj/git-extras/blob/master/bin/git-delta) don't end in a newline. As a result, that last "line" (`git diff --name-only --diff-filter=$filter $branch`) is not considered part of the text file, according to POSIX.

To fix this, I propose adding an [EditorConfig](https://editorconfig.org) file. When editors read this (some require a plugin like VSCode) file, it adjusts the formatting based on the rules. One thing to note is that rules that specify whitespace are only take affect to each inserted text. For example, with `indent_style = space`, when editing a file that predominately uses tabs, only the spaces will be inserted on each `<tab>` - other characters are not modified.

It seems both 2 and 4-space shell scripts are used in this repository, so I decided to omit that constraint from the `.editorconfig` file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,23 +1,41 @@
 name: ci
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 jobs:
+  lint:
+    runs-on: 'ubuntu-22.04'
+    steps:
+      - name: 'Get Changed Files'
+        id: 'files'
+        uses: 'masesgroup/retrieve-changed-files@v2'
+        with:
+          format: 'json'
+      - name: Check out code.
+        uses: actions/checkout@v2
+      - name: 'Check EditorConfig Lint'
+        run: |
+          sudo apt install -y jq golang
+          go install 'github.com/editorconfig-checker/editorconfig-checker/cmd/editorconfig-checker@latest'
+
+          readarray -t changed_files <<<"$(jq -r '.[]' <<<'${{ steps.files.outputs.added_modified }}')"
+          ~/go/bin/editorconfig-checker ${changed_files[@]}
+
   build:
     strategy:
       fail-fast: false
       matrix:
         platform:
-          - ubuntu-18.04
+          - ubuntu-22.04
           - macos-latest
     runs-on: ${{ matrix.platform }}
     steps:
       - name: Check out code.
         uses: actions/checkout@v2
       - name: Linux Install
-        if: matrix.platform == 'ubuntu-18.04'
+        if: matrix.platform == 'ubuntu-22.04'
         run: |
           sudo apt-get install -y bsdmainutils
       - name: Script
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.patch]
+insert_final_newline = false
+
+[*.md,*.patch]
+trim_trailing_whitespace = false


spacewander (commented):
Can we add a job to check it in the CI?

hyperupcall:
Is there anything in particular you wish to check? For consistent formatting for all of the files then maybe [shfmt](https://github.com/mvdan/sh) can be added in a different PR with CI. But EditorConfig isn't really the tool for that and there doesn't exist any first-party tools for checking EditorConfig rules.

spacewander:
Some search brings me to https://github.com/editorconfig-checker/editorconfig-checker.

hyperupcall:
`editorconfig-checker` is a third-party tool so I was hesitant to add it. Since there isn't an autofix mode, there is no automatic way of it conforming to the `.editorconfig` without something like shfmt to do the heavy lifting for us.

hyperupcall:
What are your thoughts on shfmt?

spacewander:
I don't like a formatter, which forces other people to use a style that is not used in their daily life.

hyperupcall:
Fair enough - I don't use shfmt either.

How about adding `editorconfig-checker` to CI to only run on changes files?

spacewander:
> Fair enough - I don't use shfmt either.
> 
> How about adding `editorconfig-checker` to CI to only run on changes files?

LGTM

hyperupcall:
Epic :+1: 

I also changed Ubuntu version since 18.04 is being currently phased out with [brownouts](https://github.com/actions/runner-images/issues/6002)

.github/workflows/ci.yml
```diff
@@ -1,23 +1,41 @@
...
   pull_request:
-    branches: [ master ]
+    branches: ['*']
 jobs:
+  lint:
+    runs-on: 'ubuntu-22.04'
+    steps:
+      - name: 'Get Changed Files'
+        id: 'files'
+        uses: 'jitterbit/get-changed-files@v1'
```
spacewander, line 13: I see some warnings in https://github.com/tj/git-extras/actions/runs/3979079395/jobs/6821262963
```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

Can it be avoided?

.github/workflows/ci.yml
```diff
@@ -1,23 +1,41 @@
 name: ci
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: ['*']
```
spacewander, line 6: Let's change it back once this PR is approved.

hyperupcall:
The lint checks work with the [new changes](https://github.com/tj/git-extras/actions/runs/4011065754/jobs/6888238775).

I also changed the `pull_request.branches` back, but I'm not sure how contributors are supposed to know if their files are formatted correctly

.github/workflows/ci.yml
hyperupcall: I switched to `masesgroup/retrieve-changed-files@v2`, which fixed this issue

spacewander:
> The lint checks work with the [new changes](https://github.com/tj/git-extras/actions/runs/4011065754/jobs/6888238775).
> 
> I also changed the `pull_request.branches` back, but I'm not sure how contributors are supposed to know if their files are formatted correctly

They can check this locally? It would be good to add an editorconfig plugin in the editor. Anyway, they can also figure it out by reading the output of the CI.

hyperupcall:
Oh oops 🤦 - for some reason I thought that they CI wouldn't run if `pull_request.branches` wasn't `*`. I see the other PRs show CI.

Is this ready to go?

spacewander (approved):

spacewander:
> Oh oops 🤦 - for some reason I thought that CI wouldn't run if `pull_request.branches` wasn't `*`. But I now see that the other PRs show CI.
> 
> Is this ready to go?

Merged. Thanks!

PR #946, opened 2021-12-29
Consider init.defaultBranch in git_extra_default_branch, changed default to main

tfendin:
Changed according to issue #943 

If the config variable git-extras.default-branch use the value from init.defaultBranch. Also changed the default branch to main.

PR #1033, opened 2023-02-12
fix: Unify handling of nprocs

### LINKED ISSUE #29: should git release use -a ?
I expect most release tags are created with -a.

### LINKED ISSUE #943: Configuration Variable Default Branch
Hello!

Acording to the man documentation there is the configuration variable `git-extras.default-branch`.

Is it possible to get this info directly from the vanilla git configuration variable `init.default-branch`?
And, make the default to `main` following the latest trend?

If so, this git-extras config var can be only for environmental purposes.

What do you think?

hyperupcall:
The default was also changed to 2, as it should be.

Been looking through some issues, and the following should be closed as they are already implemented:

- Closes #29 (-a [already used](https://github.com/tj/git-extras/blob/a008308267bb8e0042b1f8bf7606902eb4980b8a/bin/git-release#L131))
- Closes #943 (via https://github.com/tj/git-extras/pull/946)

### DIFF
--- a/helper/git-extra-utility
+++ b/helper/git-extra-utility
@@ -22,7 +22,9 @@ git_extra_default_branch() {
 # get processor count, default 2
 #
 git_extra_get_nproc() {
-    if ! (nproc || getconf NPROCESSORS_ONLN || getconf _NPROCESSORS_ONLN || grep -c ^processor /proc/cpuinfo) 2>/dev/null; then
-        printf '%s\n' 2
+    local default=${1:-2}
+
+    if ! { nproc || getconf NPROCESSORS_ONLN || getconf _NPROCESSORS_ONLN || grep -c ^processor /proc/cpuinfo; } 2>/dev/null; then
+        printf '%s\n' "$default"
     fi
 }
--- a/helper/git-extra-utility
+++ b/helper/git-extra-utility
@@ -17,3 +17,12 @@ git_extra_default_branch() {
         echo "main"
     fi
 }
+
+#
+# get processor count, default 2
+#
+git_extra_get_nproc() {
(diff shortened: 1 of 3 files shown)


spacewander:
> The default was also changed to 2, as it should be.
> 
> Been looking through some issues, and the following should be closed as they are already implemented:
> 
>     * Closes [should git release use -a ? #29](https://github.com/tj/git-extras/issues/29) (-a [already used](https://github.com/tj/git-extras/blob/a008308267bb8e0042b1f8bf7606902eb4980b8a/bin/git-release#L131))
> 
>     * Closes [Configuration Variable Default Branch #943](https://github.com/tj/git-extras/issues/943) (via [Consider init.defaultBranch in git_extra_default_branch, changed default to main #946](https://github.com/tj/git-extras/pull/946))

Thanks for your report

helper/git-extra-utility
```diff
@@ -17,3 +17,12 @@ git_extra_default_branch() {
         echo "main"
     fi
 }
+
+#
+# get processor count, default 2
+#
+git_extra_get_nproc() {
+    if ! (nproc || getconf NPROCESSORS_ONLN || getconf _NPROCESSORS_ONLN || grep -c ^processor /proc/cpuinfo) 2>/dev/null; then
+        printf '%s\n' 2
```
spacewander, line 26: Can we allow the caller to push the default value?

spacewander (commented):
Nowadays, we are no longer enthusiastic about adding new helper functions. This is because it is not common practice and makes development cumbersome.

helper/git-extra-utility
hyperupcall: :+1:

hyperupcall:
> Nowadays, we are no longer enthusiastic about adding new helper functions. This is because it is not common practice and makes development cumbersome.

Should the current mechanism be replaced with something in which a bash file with the helper functions is sourced at runtime?

PR #1034, opened 2023-02-13
Add the current branch git summary

guenhter:
Simply adds the current branch name to the summary output.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -153,6 +153,7 @@ $ git summary
 
 project     : git-extras
 repo age    : 10 months ago
+branch      : master
 last active : 3 weeks ago
 active on   : 93 days
 commits     : 163
@@ -226,11 +227,11 @@ This is how the `tebular` output style and `oneline` output style look like
 
 ```bash
 $ git summary --output-style tabular
-# Repo     | Age       | Last active | Active on | Commits | Uncommitted
-git-extras | 13 years  | 7 hours ago | 807 days  | 1703    | 3
+# Repo     | Age       | Last active | Active on | Commits | Uncommitted | Branch
+git-extras | 13 years  | 7 hours ago | 807 days  | 1703    | 3           | master
 
 $ git summary --output-style oneline
-git-extras / age: 13 years / last active: 7 hours ago / active on 807 days / commits: 1703 / uncommitted: 3
+git-extras / age: 13 years / last active: 7 hours ago / active on 807 days / commits: 1703 / uncommitted: 3 / branch: master
 ```
 
 ## git effort
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -172,6 +172,10 @@ single_file() {
   done
 }
 
+current_branch_name() {
+  git rev-parse --abbrev-ref HEAD
+}
+
 #
 # list the author for all file
 #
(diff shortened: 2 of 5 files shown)


Commands.md
```diff
@@ -226,11 +226,11 @@ This is how the `tebular` output style and `oneline` output style look like
 
 ```bash
 $ git summary --output-style tabular
-# Repo     | Age       | Last active | Active on | Commits | Uncommitted
```
spacewander, line 229: We need to update the non-output-style version too.

guenhter: Sorry. This slipped through. It is not updated. I double checked all locations and the `branch` is not contained everywhere.

bin/git-summary
```diff
@@ -211,14 +215,15 @@ print_summary_by_line() {
 
 print_summary() {
   if [ "$OUTPUT_STYLE" == "tabular" ]; then
-    tabular_headers="# Repo $SP Age $SP Last active $SP Active on $SP Commits $SP Uncommitted"
-    echo -e "$tabular_headers\n$project $SP $(repository_age) $SP $(last_active) $SP $(active_days $commit) days $SP $(commit_count $commit) $SP $(uncommitted_changes_count)" | column -t -s "$COLUMN_CMD_DELIMTER"
+    tabular_headers="# Repo $SP Age $SP Last active $SP Active on $SP Commits $SP Uncommitted $SP Branch"
+    echo -e "$tabular_headers\n$project $SP $(repository_age) $SP $(last_active) $SP $(active_days $commit) days $SP $(commit_count $commit) $SP $(uncommitted_changes_count) $SP $(current_branch_name)" | column -t -s "$COLUMN_CMD_DELIMTER"
   elif [ "$OUTPUT_STYLE" == "oneline" ]; then
-    echo "$project / age: $(repository_age) / last active: $(last_active) / active on $(active_days $commit) days / commits: $(commit_count $commit) / uncommitted: $(uncommitted_changes_count)"
+    echo "$project / age: $(repository_age) / last active: $(last_active) / active on $(active_days $commit) days / commits: $(commit_count $commit) / uncommitted: $(uncommitted_changes_count) / branch: $(current_branch_name)"
```
spacewander, line 221: Look like we need to escape the branch name, as it can contain `/`

guenhter: Do you mean by "escape" to just put it withing double quotes?

On the other hand, is it a problem if the branch name contains a `/`, because it is anyway just some output...

spacewander: I wonder if we escape the `/`, can awk or column handle it correctly?

guenhter: I guess if you want to `awk` this line, you always can use ` / ` (space before and after) to distinguish it from slashes in a text

spacewander: Maybe we can mention the `space before and after` in the doc?

guenhter: Done. I mentioned the separator for the tabular and the oneline option in the documentation and also that they are surrounded by spaces.

spacewander (approved):

spacewander:
Would you solve the conflict? Thanks!

guenhter:
Done.

spacewander (commented):
You can use `git ignore -g` to ignore .idea files.
https://github.com/tj/git-extras/blob/master/man/git-ignore.md

guenhter:
Sorry for committing .idea again. It's gone now.

spacewander (approved):

PR #1047, opened 2023-05-17
Improve `git-ignore` command

### LINKED ISSUE #38: git ignore  doesn't interact with tab completion
feature request!
:)

### LINKED ISSUE #347: .gitignore without an initialized repo
A lot of times when I'm just creating a project, I want to add a few files and directories to the `.gitignore`, before setting up git.

The behavior before was to just run `git ignore node_modules` and it would create a `.gitignore` file, not it's errors out with `Not a git repo!`. I understand why this is here, but it forces you to manually create the file and add these files before committing.

### LINKED ISSUE #427: Is makefile.msys duplicate?
I find out that git-extras can be installed with the Makefile under msysgit, with a few path modified.
So I wonder whether `makefile.msys` is duplicate? Or maybe my msysgit environment is different?

By the way, my version of `make` is `4.1`.

Willing to know the story of `makefile.msys` from @southly

hyperupcall:
Closes #38
  - `git ignore<TAB>` now shows file/directory names
  
Closes #347
  - No longer prints "fatal: not a git repository"-type errors to console. If no (local) Git repository is present, handle gracefully

Closes #427 (stale issue about file that does not exist)

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-: ${GIT_DIR:=$(git rev-parse --git-dir)}
+GIT_DIR=$(git rev-parse --git-dir 2>/dev/null)
 
 function show_contents {
   local file="${2/#~/$HOME}"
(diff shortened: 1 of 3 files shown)


bin/git-ignore
```diff
@@ -11,18 +11,30 @@ function show_contents {
   fi
 }
 
+function maybe_cd_to_git_root {
+  local result=
+  if result=$(git rev-parse --show-toplevel 2>/dev/null); then
```
spacewander, line 16: Let's add this cmd in https://github.com/tj/git-extras/blob/master/not_need_git_repo

bin/git-ignore
```diff
@@ -34,22 +46,22 @@ function add_global {
...
 function add_local {
-  cd "$(git root)"
+  maybe_cd_to_git_root
   add_patterns .gitignore "$@"
 }
 
 function show_private {
-  cd "$(git root)"
+  maybe_cd_to_git_root
   show_contents Private "${GIT_DIR}/info/exclude"
```
spacewander, line 60: We must require a git repo for setting the private ignorefile. As it's private to the git repo.

hyperupcall:
Sorry for the slow reply, I have addressed the feedback

spacewander (approved):

PR #1053, opened 2023-06-22
feat(git-fork): add flags comprehension

mjholub:
Initially I just wanted to add support for directly forking the current repo (say you clone something, notice something you want/need to change, and you don't want to use `git-fork $(git remote get-url origin)`, so I added the `-c` and `--current` flags for convenience, along with the help, token and target dir flags (because I'd rather keep my forks in *~/projects/forks*, than the subdirectory of the upstream repo, which I usually keep in *~/src*)

### DIFF
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -1,67 +1,120 @@
 #!/usr/bin/env bash
 
 abort() {
-  echo "$@"
-  exit 1
+        echo "$@"
+        exit 1
 }
 
-url="$1"
-test -z "$url" && url=$(git remote get-url origin 2> /dev/null) && origin=true
-# validate repo url
-test -z "$url" && abort "github repo needs to be specified as an argument"
+print_help() {
+        echo "Usage: git fork [options] [github-repo-url]"
+        echo "Options:"
+        echo "  -h, --help          Print this help message"
+        echo "  -c, --current       Use the URL of the current Git repository as the source"
+        echo "  -t, --token <token> GitHub personal access token"
+        echo "  -d, --target-dir <directory>  Target directory for cloning the forked repository"
+}
 
-# validate user
-echo "Enter your github username"
-read -r user
-[ -n "$user" ] || abort "git username required"
-# personal access token
-# config name is github-personal-access-token '_' is not allowed in git config
+parse_args() {
+        while getopts ":hct:d:" opt; do
+                case $opt in
+                h)
+                        print_help
+                        exit 0
+                        ;;
+                c)
+                        origin=true
+                        ;;
+                t)
+                        github_personal_access_token="$OPTARG"
+                        ;;
+                d)
+                        target_dir="$OPTARG"
+                        ;;
+                \?)
+                        abort "Invalid option: -$OPTARG"
+                        ;;
+                :)
+                        abort "Option -$OPTARG requires an argument"
+                        ;;
+                esac
+        done
+}
 
-github_personal_access_token=$(git config git-extras.github-personal-access-token)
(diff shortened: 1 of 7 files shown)


man/git-fork.md
```diff
@@ -7,31 +7,47 @@ git-fork(1) -- Fork a repo on github
...
 ## DESCRIPTION
 
-  If a github repo url is given, fork the repo. Like clone but forks first.
+  If a GitHub repo URL is given, the command forks the specified repository. It performs the following actions:
 
-  1. forks the repo on github
-  2. clones the repo into the current dir
-  3. adds the original repo as a remote called `upstream`
+  1. Forks the repo on GitHub.
+  2. Clones the repo into the current directory or the one specified with `d` or `--target-dir`.
```
spacewander, line 13:
```suggestion
  2. Clones the repo into the current directory or the one specified with `-d` or `--target-dir`.
```

etc/bash_completion.sh
```diff
@@ -1,161 +1,190 @@
...
-    '--start-commit'
-    '--no-merges'
-    '--prune-old'
-    '--stdout'
-    '--help'
-  )
-  local merged_opts_str=""
-  merged_opts_str+="$(printf "%s " "${s_opts[@]}")"
-  merged_opts_str+="$(printf "%s " "${l_opts[@]}")"
+_git_changelog() {
```
spacewander, line 4: Could you avoid changing the indent of other commands' functions? Thanks!

etc/bash_completion.sh
```diff
@@ -1,161 +1,190 @@
...
+                COMPREPLY=($(compgen -d -- "${cur}"))
+                return 0
+                ;;
+        *) ;;
+        esac
+
+        if [[ "${cur}" == -* ]]; then
+                COMPREPLY=($(compgen -W "${opts}" -- ${cur}))
+                return 0
+        else
```
spacewander, line 114: We can use `fi` here or move the common `return 0` out?

etc/git-extras.fish
```diff
@@ -118,18 +118,60 @@ complete -c git -x -n "__fish_git_using_command delete-tag" -a '(__fish_git for-
...
+            return
+
+        case '*'
+
+            # Complete options
+            set -l completions (commandline -pco)
+            for completion in $completions
+                echo $completion
+            end
+            return
```
spacewander, line 155: It seems the execution is already return in every branch?

bin/git-fork
```diff
@@ -1,67 +1,120 @@
...
+                # Rename and add remote for the existing repository
+                git remote rename origin upstream
+                git remote add origin "${remote_prefix}${user}/${project}.git"
+                git fetch origin
+        else
+                # Clone forked repo into the specified or default directory
+                local clone_dir="${target_dir:-$project}"
+                git clone "${remote_prefix}${user}/${project}.git" "$clone_dir"
+                # Add reference to the original fork for merging upstream changes
+                cd "$clone_dir" || exit
```
spacewander, line 89: Should we exit with non-zero?

bin/git-fork
```diff
@@ -1,67 +1,120 @@
...
+main() {
+        parse_args "$@"
+
+        # Prompt for GitHub username if not provided
+        echo "Enter your GitHub username:"
+        read -r user
+        [ -n "$user" ] || abort "GitHub username is required"
+
+        # Prompt for personal access token if not provided
+        if [ -z "$github_personal_access_token" ]; then
```
spacewander, line 104: It seems we no longer get the token from git config anymore

PR #1054, opened 2023-06-29
I have made two improvements to the git-bulk:

icodebuster:
1. Previously, if the "repository.txt" file did not end with a blank line, the last entry in the list of repositories was omitted during the cloning process. This limitation has been addressed, and now all repositories listed in the file, including the last one, will be successfully cloned regardless of whether the file ends with a blank line or not.

2. There is support for cloning repositories into custom folder names when using the "repository.txt" file.

repositories\.txt
\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-
https://host-of-git/repo-1.git Repo-1-Code
https://host-of-git/repo-2.git Repo-2-Code
https://host-of-git/repo-3.git Repo-3-Code

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -36,7 +36,7 @@ function addworkspace {
             source=$(realpath "$source" 2>/dev/null)
             if [ -f "$source" ]; then
                 pushd "$wsdir" > /dev/null
-                while IFS= read -r line || [[ -n $line ]]; do
+                while IFS= read -r line; do
                   if [ -n "$line" ]; then
                     git clone $line;
                   fi
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -36,8 +36,12 @@ function addworkspace {
             source=$(realpath "$source" 2>/dev/null)
             if [ -f "$source" ]; then
                 pushd "$wsdir" > /dev/null
-                while read -r line; do git clone "$line"; done < "$source";
-                popd > /dev/null
+                    while IFS= read -r line || [[ -n $line ]]; do
+                      if [ -n "$line" ]; then
+                        git clone $line;
+                      fi
+                    done < "$source"
+                    popd > /dev/null
             else
                 echo 1>&2 "format of URL or file unknown"
             fi
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -38,6 +38,7 @@ function addworkspace {
                 pushd "$wsdir" > /dev/null
                 while IFS= read -r line; do
                   if [ -n "$line" ]; then
+                    # the git clone command to take the complete line in the repository.txt as separate argument. This facilitated the cloning of the repository with a custom folder name.
                     git clone $line;
                   fi
                 done < "$source"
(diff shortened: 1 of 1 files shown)


bin/git-bulk
```diff
@@ -36,8 +36,12 @@ function addworkspace {
             source=$(realpath "$source" 2>/dev/null)
             if [ -f "$source" ]; then
                 pushd "$wsdir" > /dev/null
-                while read -r line; do git clone "$line"; done < "$source";
-                popd > /dev/null
+                    while IFS= read -r line || [[ -n $line ]]; do
+                      if [ -n "$line" ]; then
```
spacewander, line 40: Would you honor the original indentation? Thanks!

icodebuster: Done

bin/git-bulk
```diff
@@ -36,7 +36,11 @@ function addworkspace {
             source=$(realpath "$source" 2>/dev/null)
             if [ -f "$source" ]; then
                 pushd "$wsdir" > /dev/null
-                while read -r line; do git clone "$line"; done < "$source";
+                while IFS= read -r line || [[ -n $line ]]; do
+                  if [ -n "$line" ]; then
+                    git clone $line;
```
spacewander, line 41: Let‘s add a comment that it's dedicated not to quoting the var.

spacewander (approved):

bin/git-bulk
icodebuster: I did not get you. Could you please elaborate?
Thank you.

spacewander: It's intended not to use `"$line"` here, right?

bin/git-bulk
```diff
@@ -36,7 +36,11 @@ function addworkspace {
             source=$(realpath "$source" 2>/dev/null)
             if [ -f "$source" ]; then
                 pushd "$wsdir" > /dev/null
-                while read -r line; do git clone "$line"; done < "$source";
+                while IFS= read -r line || [[ -n $line ]]; do
```
spacewander, line 39: Do we need to use `-n "$line"` here?

icodebuster: `-n "$line"` checks whether the variable $line is non-empty. It's needed since the `repository.txt` file can contain empty line.

icodebuster: Yes, you are right, we don't need `-n $line`. I have updated the code.

bin/git-bulk
icodebuster: If double quotes are used, the resulting git clone command will be:

```
git clone "https://host-of-git/repo-1.git Repo-1-Code"
```

This command may result in an invalid URL since the repository URL and the destination folder are enclosed in double quotes.

However, if the double quotes are removed, the git clone command will be:

```
git clone https://host-of-git/repo-1.git Repo-1-Code
```

By removing the double quotes, the command will create a separate folder named "Repo-1-Code" for the cloned repository, which is the enhancement I am suggesting.

spacewander: @icodebuster 
Would you add the explanation as a comment in the code?

icodebuster: I have added a short comment as an explanation. If you have any alternate comments/suggestions, do let me know.

spacewander (approved):

PR #1056, opened 2023-07-19
fix: Improve Bash hygiene

hyperupcall:


bin/git-summary
```diff
@@ -216,9 +216,9 @@ print_summary_by_line() {
 print_summary() {
   if [ "$OUTPUT_STYLE" == "tabular" ]; then
     tabular_headers="# Repo $SP Age $SP Last active $SP Active on $SP Commits $SP Uncommitted $SP Branch"
-    echo -e "$tabular_headers\n$project $SP $(repository_age) $SP $(last_active) $SP $(active_days $commit) days $SP $(commit_count $commit) $SP $(uncommitted_changes_count) $SP $(current_branch_name)" | column -t -s "$COLUMN_CMD_DELIMTER"
+    echo -e "$tabular_headers\n$project $SP $(repository_age) $SP $(last_active) $SP $(active_days "$commit") days $SP $(commit_count "$commit") $SP $(uncommitted_changes_count) $SP $(current_branch_name)" | column -t -s "$COLUMN_CMD_DELIMTER"
```
spacewander, line 219: The quote is missing intently. We have `shellcheck disable=SC2086` mark in `active_days` and other similar functions.

hyperupcall: To me it doesn't look like the `shellcheck disable=SC2086` should even be in `active_days`. And actually, looking at all the ShellCheck disable directives, the ones that deal with `$commit` look incorrect. So to me it seems we are dealing with a larger issue that has been surpressed.

From the top, `git-summary` has a synposis of `git-summary [--line] [--dedup-by-email] [<committish>]`, and accepts a _single_ `committish` (even if it accepted multiple committishes, it wouldn't make sense anyways since the underlying git subcommands acccept only a single commitish). The `$commit` variable is either assigned `HEAD` or `$*` (which corresponds to `<committish>`. The `$*` looks fishy itself, but doesn't need to be changed.

Concerning the `active_days` function, the only argument passed to it is `$commit`, which is eventually passed to `git log`. In other functions, `$commit` is passed to `git shortlog`. All these git subcommands have the commonality of accepting a single `[<revision range>]`.

`[<revision range>]` is defined in `gitrevisions(7)`, and some valid revisions include things that contain spaces, like `HEAD^{/fix nasty bug}`, `HEAD@{5 minutes ago}`. So by not quoting `$commit`, it word splits on committishes that include spaces.

hyperupcall: However, there are cases in which commit may be empty and word splitting is intended - so those of course will have to be handled as well

bin/git-changelog
```diff
@@ -280,7 +280,8 @@ commitList() {
   for (( i=0; i<"${_tags_list_keys_length}"; i++ )); do
     local __curr_tag="${tags_list_keys[$i]}"
     local __prev_tag="${tags_list_keys[$i+1]:-null}"
-    local __curr_date="$(_valueForKeyFakeAssocArray "${__curr_tag}" "${tags_list[*]}")"
+    local __curr_date=
```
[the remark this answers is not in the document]
hyperupcall: I suppose this is a stylistic preference (the small behavior difference isn't relevant to every Bash project), so I went ahead and removed the extra `=`.

bin/git-summary
```diff
@@ -45,30 +45,29 @@ if [ -n "$MERGES_ARG" ] && [ -n "$SUMMARY_BY_LINE" ]; then
...
 fi
 project=${PWD##*/}
 
 #
 # get date for the given <commit>
 #
-date() {
+commit_date() {
   # the $1 can be empty
   # shellcheck disable=SC2086
```
spacewander, line 61: We can remove the shellcheck now?

hyperupcall: I removed one extra one, but the rest cannot be removed because of `MERGES_ARG`.

spacewander: It seems we can quote MERGES_ARG as we don't expect any whitespace in it.

hyperupcall: Unfortunately not because that would mean an empty argument will be passed to `git`. It's funtionally similar to doing this:

```sh
$ git log '' --oneline HEAD
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

hyperupcall: In other words it shouldn't be quoted so the whitespace can be skipped by the shell during word splitting.

spacewander: I see. Thanks for clarifying that.

spacewander (approved):

PR #1059, opened 2023-07-27
Filter authors by commit count

tfendin:
In git summary

- Show the number of authors before listing them
- Added option --author-commit-limit to filter the list of authors

This gives better UX when running it on very large repos.

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -3,11 +3,7 @@
 
 cd "$(git root)" || { echo "Can't cd to top level directory";exit 1; }
 
-authors_file=
-_exit() {
-  rm -f "${authors_file}"
-}
-trap '_exit' EXIT
+AUTHOR_LISTING=
 
 SUMMARY_BY_LINE=
 DEDUP_BY_EMAIL=
@@ -147,7 +143,7 @@ format_authors() {
   # a rare unicode character is used as separator to avoid conflicting with
   # author name. However, Linux column utility will escape tab if separator
   # specified, so we do unesaping after it.
-  LC_ALL=C awk '
+  echo "${AUTHOR_LISTING:?}" | LC_ALL=C awk '
   BEGIN {
      commitLimit = strtonum('"${AUTHOR_COMMIT_LIMIT}"');
   }
@@ -211,7 +207,7 @@ uncommitted_changes_count() {
 }
 
 number_of_authors() {
-  wc -l ${authors_file:?}|cut -d$' ' -f1
+  echo "${AUTHOR_LISTING:?}" | wc -l
 }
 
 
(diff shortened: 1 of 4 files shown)


bin/git-summary
```diff
@@ -235,15 +257,17 @@ print_summary() {
       echo " files       : $(file_count)"
     fi
     echo " uncommitted : $(uncommitted_changes_count)"
-    echo " authors     : "
+    authors_file=$(git_extra_mktemp)
     if [ -n "$DEDUP_BY_EMAIL" ]; then
       # the $commit can be empty
       # shellcheck disable=SC2086
-      git shortlog $MERGES_ARG -n -s -e $commit | dedup_by_email | format_authors
+      git shortlog $MERGES_ARG -n -s -e $commit | dedup_by_email >${authors_file}
```
spacewander, line 264: Can't we store the output in a variable instead of a file?

bin/git-summary
```diff
@@ -228,10 +224,9 @@ print_summary_by_line() {
     echo
     echo " project     : $project"
     echo " lines       : $(line_count "${paths[@]}")"
-    authors_file=$(git_extra_mktemp)
-    lines "${paths[@]}" | sort | uniq -c | sort -rn >${authors_file}
+    AUTHOR_LISTING=$(lines "${paths[@]}" | sort | uniq -c | sort -rn)
     echo " authors     : $(number_of_authors)"
-    format_authors <${authors_file}
+    format_authors
```
spacewander, line 229: Since AUTHOR_LISTING and format_authors are in the same function, could we pass AUTHOR_LISTING as an argument to format_authors?

tfendin: Yes, now both `format_authors` and `number_of_authors` take the listing as an argument.

bin/git-summary
```diff
@@ -136,10 +143,15 @@ format_authors() {
...
   # specified, so we do unesaping after it.
-  LC_ALL=C awk '
+  echo "${1:?}" | LC_ALL=C awk '
+  BEGIN {
+     commitLimit = strtonum('"${AUTHOR_COMMIT_LIMIT}"');
+  }
   { args[NR] = $0; sum += $0 }
   END {
     for (i = 1; i <= NR; ++i) {
+      if (strtonum(args[i]) < commitLimit)
```
spacewander, line 153: The `strtonum` seems to be GNU awk extension, can we find an alternative?

bin/git-summary
```diff
@@ -22,6 +25,10 @@ for arg in "$@"; do
             OUTPUT_STYLE="$2"
             shift
             ;;
+        --author-commit-limit)
+            AUTHOR_COMMIT_LIMIT="$2"
```
spacewander, line 29: Do we need to check if the commit limit exists?

bin/git-summary
tfendin: Seems like it's unnecessary to call `strtonum` at all, the comparison should work anyway.

bin/git-summary
tfendin: Perhaps we should. I Just copied the code for `--output-style`.

tfendin:
The completion files should be updated as well.

tfendin:
I'm thinking of scrapping this PR and add an option to limit the number of shown authors instead, perhaps with a sensible default value. That seems easier for the user. Any thoughts?

hyperupcall:
> Any thoughts?

I think either possibilities are good. I can a case where either one makes sense. If an option is added that limits the number of shown authors, it should default to the value that preserves the output. I _believe_ that means the default would be no limit and all contributors are listed.

PR #1066, opened 2023-09-08
feat(auto-test): CI with pytest

### LINKED ISSUE #1058: Feature request: the continue integration
Hi, guys,

There is nothing test with the command like the `git-abort`. It is hardly remember any detail of any command when time goes by.

> The Only Constant in Life Is Change.

So I think is useful to create the unit test for each command to leave the description how the command has been expected to run. For now, there are 74 commands, that means 74 uts needed. It is a big job. But first, we can find some ways to save life.

At least, three ways to create the unit test I think, the detials as below:

1. Integration tests in Bash
    * It is great and no more side effect, but i did not know the detail how to, so if someone gives a hand about how to, grateful!
2. Integration tests in Python
    * Save life always
3. Integration tests in Node
    * My favourite cause I use it a lot and so many pkgs to support the process

I think we can take a vote here, any new one is welcome!

One more, the github action or [Travis](https://travisonline.com/) or any other?

#### Reference
* [About continuous integration](https://docs.github.com/en/actions/automating-builds-and-tests/about-continuous-integration)
* [CI/CD explained](https://resources.github.com/ci-cd/)
* [pytest](https://github.com/pytest-dev/pytest/)

vanpipy:
Close #1058

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,23 @@ jobs:
           pip install codespell==2.2
           git grep --cached -l '' | grep -v -e 'History\.md' -e 'AUTHORS' -e 'man/.*\.1' -e 'man/.*\.html' | xargs codespell --ignore-words=.github/.ignore_words
 
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest==7.4.0
+          pip install GitPython==3.1.36
+      - name: Unit test
+        run: make test
+
   build:
     strategy:
       fail-fast: false
--- a/Readme.md
+++ b/Readme.md
@@ -18,6 +18,18 @@ Go to the [Commands](Commands.md) page for basic usage and examples.
 
 __GIT utilities__ -- repo summary, repl, changelog population, author commit percentages and more
 
+## Test
+
+Nowaday the continues integrations test is coming and you can access it via the `make test` easily.
+
+The CI depends on
+
+* `python==3.10`
+* `pytest==7.4.0`
+* `GitPython==3.1.36`
+
+So the version or higher is recommended.
+
 ## Contributing
 
 Interested in contributing? Awesome!
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,18 @@
+# Sharing fixtures
+# Ref: https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
+
+import pytest
+from helper import TempRepository
+
+@pytest.fixture(scope="module")
+def temp_repo():
+    repo = TempRepository()
+    git = repo.get_repo_git()
+    tmp_file_a = repo.create_tmp_file()
+    tmp_file_b = repo.create_tmp_file()
+    repo.switch_cwd_under_repo()
+    git.add(".")
+    git.config("--local", "user.name", "test")
+    git.config("--local", "user.email", "test@git-extras.com")
+    git.commit("-m", "chore: initial commit")
+    return repo
--- a/tests/test_git_abort.py
+++ b/tests/test_git_abort.py
@@ -1,0 +1,65 @@
+from git import GitCommandError
+
+class TestGitAbort:
+    def test_init(self, temp_repo):
+        git = temp_repo.get_repo_git()
+        tmp_file = temp_repo.get_file(0)
+        git.branch("A")
+        git.branch("B")
+        git.checkout("A")
+        temp_repo.writefile(tmp_file, "a")
+        git.add(".")
+        git.commit("-m", "A")
+        git.checkout("B")
+        temp_repo.writefile(tmp_file, "b")
+        git.add(".")
+        git.commit("-m", "B")
+        git.status()
+
+    def test_cherry_pick(self, temp_repo):
+        git = temp_repo.get_repo_git()
+        try:
+            git.cherry_pick("A")
+        except GitCommandError as err:
+            print(err)
+        result = git.status()
+        assert "Unmerged path" in result
+        temp_repo.invoke_extras_command("abort")
+        result = git.status()
+        assert "nothing to commit, working tree clean" in result
+
+    def test_merge(self, temp_repo):
+        git = temp_repo.get_repo_git()
+        try:
+            git.merge("A")
+        except GitCommandError as err:
+            print(err)
+        result = git.status()
+        assert "Unmerged path" in result
+        temp_repo.invoke_extras_command("abort")
+        result = git.status()
+        assert "nothing to commit, working tree clean" in result
+
+    def test_rebase(self, temp_repo):
+        git = temp_repo.get_repo_git()
+        try:
+            git.rebase("A")
+        except GitCommandError as err:
+            print(err)
+        result = git.status()
+        assert "Unmerged path" in result
+        temp_repo.invoke_extras_command("abort")
+        result = git.status()
+        assert "nothing to commit, working tree clean" in result
+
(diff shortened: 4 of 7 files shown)


tests/test_git-abort.py
```diff
@@ -0,0 +1,60 @@
...
+        result = subprocess.run(shlex.split("git cherry-pick A"), capture_output=True)
+        result = subprocess.run(shlex.split("git status"), capture_output=True)
+        assert "Unmerged path" in result.stdout.decode()
+        subprocess.run(shlex.split(os.path.join(git_extras_cwd, 'bin', 'git-abort')))
+        result = subprocess.run(shlex.split("git status"), capture_output=True)
+        assert "nothing to commit, working tree clean" in result.stdout.decode()
+
+    def test_merge(self, git_repo):
+        git_extras_cwd = git_repo[0]
+        git_extras_cwd = git_repo[0]
```
spacewander, line 35: Duplicate code?

tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    result = subprocess.run(shlex.split("git add ."), capture_output=True)
+    print(result.stdout.decode())
+    print(result.stderr.decode())
+    subprocess.run(shlex.split("git config --local user.name \"test\""))
+    subprocess.run(shlex.split("git config --local user.email \"test@git-extras.com\""))
+    result = subprocess.run(shlex.split("git commit -m 'chore: initial commit'"), capture_output=True)
+    print(result.stdout.decode())
+    print(result.stderr.decode())
+    yield [git_extras_cwd, tmp_dir, tmp_file_a, tmp_file_b]
+    shutil.rmtree(tmp_dir, ignore_errors=True)
```
spacewander, line 30: Better to keep the tmp dir so we can look into it once thing goes wrong

tests/test_git-abort.py
```diff
@@ -0,0 +1,60 @@
+import os
```
spacewander, line 1: `test_git_abort.py` would be a better name

tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+@pytest.fixture(scope="module")
+def git_repo():
+    git_extras_cwd = os.getcwd()
+    tmp_dir = tempfile.mkdtemp()
+    tmp_file_a = tempfile.mkstemp(dir=tmp_dir)
+    tmp_file_b = tempfile.mkstemp(dir=tmp_dir)
+    os.chdir(tmp_dir)
+    result = subprocess.run(shlex.split("git init"), capture_output=True)
```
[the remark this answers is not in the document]
hyperupcall: Do you think it would be difficult to know what command is really being ran if using an external library? There could be a mismatch between the test suite and what users would usually run on the command line.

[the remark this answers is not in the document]
spacewander: It provides a simple DSL to run git command as in the shell:
https://gitpython.readthedocs.io/en/stable/tutorial.html#using-git-directly

This way is readable.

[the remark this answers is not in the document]
hyperupcall: Oh I see, thanks! That's quite readable

.github/workflows/ci.yml
```diff
@@ -35,6 +35,27 @@ jobs:
...
+  test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: ['3.10']
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version  }}
```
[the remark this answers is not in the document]
vanpipy: It is only 3.10 now, just keep the matrix for later.

tests/test_git-abort.py
```diff
@@ -0,0 +1,60 @@
...
+        subprocess.run(shlex.split("git add ."))
+        subprocess.run(shlex.split("git commit -m \"B\""))
+        subprocess.run(shlex.split("git status"))
+
+    def test_cherry_pick(self, git_repo):
+        git_extras_cwd = git_repo[0]
+        result = subprocess.run(shlex.split("git cherry-pick A"), capture_output=True)
+        result = subprocess.run(shlex.split("git status"), capture_output=True)
+        assert "Unmerged path" in result.stdout.decode()
+        subprocess.run(shlex.split(os.path.join(git_extras_cwd, 'bin', 'git-abort')))
```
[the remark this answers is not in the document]
vanpipy: Great, I will create a helper for it.

tests/test_git-abort.py
vanpipy: Yep, remove it later.

tests/test_git-abort.py
vanpipy: I juse think `test a command`, so `test git-abort` is `test_git-abort`. And `test the command` should be `test git abort`, it is a noum, but i think both are ok. i will reset it later.

Readme.md
```diff
@@ -12,6 +12,10 @@ Just getting started? Check out these screencasts:
 
 See the [Installation](Installation.md) page.
 
+## Test
+1. `pip install pytest`
```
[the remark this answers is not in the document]
vanpipy: Move to `make test` and explain in the readme.

.github/workflows/ci.yml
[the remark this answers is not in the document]
vanpipy: Keep only one python 3.10.

tests/test_git_abort.py
```diff
@@ -0,0 +1,72 @@
...
+
+class TestGitAbort:
+    def test_init(self, git_repo):
+        repo, tmp_dir, tmp_a, tmp_b = git_repo
+        repo.git.branch("A")
+        repo.git.branch("B")
+        repo.git.checkout("A")
+        with open(tmp_a[1], "w", encoding="utf-8") as file_a_on_A:
+            file_a_on_A.write("a")
+            file_a_on_A.close()
```
spacewander, line 15: The close is done automatically in `with open` syntax.

tests/test_git_abort.py
```diff
@@ -0,0 +1,72 @@
...
+        repo.git.add(".")
+        repo.git.commit("-m", "B")
+        repo.git.status()
+
+    def test_cherry_pick(self, git_repo):
+        repo = git_repo[0]
+        try:
+            repo.git.cherry_pick("A")
+        except git.GitCommandError:
+            pass
```
spacewander, line 31: Let's print the error if the cmd failed unexpectedly

tests/test_git_abort.py
```diff
@@ -0,0 +1,72 @@
...
+        result = repo.git.status()
+        assert "Unmerged path" in result
+        invoke_git_extras_command("git-abort")
```
spacewander, line 34: Would be better to move `git-` prefix to the helper?

tests/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+    tmp_dir = tempfile.mkdtemp()
+    tmp_file_a = tempfile.mkstemp(dir=tmp_dir)
+    tmp_file_b = tempfile.mkstemp(dir=tmp_dir)
+    repo = git.Repo.init(tmp_dir)
+    os.chdir(tmp_dir)
+    repo.git.add(".")
+    repo.git.config("--local", "user.name", "test")
+    repo.git.config("--local", "user.email", "test@git-extras.com")
+    repo.git.commit("-m", "chore: initial commit")
+    return [repo, tmp_dir, tmp_file_a, tmp_file_b]
```
spacewander, line 24: Would be better to return an object, which provides some method like `.create_tmp_file`. So people don't need to remember the argument order.

vanpipy: Got the idea.

tests/test_git_abort.py
vanpipy: Remove the prefix in the util class and keep it in the original function.

tests/test_git_abort.py
vanpipy: Dump the error when unexpected.

tests/test_git_abort.py
vanpipy: Remove the `close`.

spacewander (approved):

PR #1072, opened 2023-09-11
git-feature: add configurable branch separator

### LINKED ISSUE #1069: [feature request] Add configurable feature branch separator
After having been burned a few too many times by the use of `/` to separate branch parts (it works, but it doesn't necessarily work the way that people think it should unless they're git experts), I have an aversion to this separator being used and do not allow it at the workplace. It would be useful to be able to have a default configuration and/or a flag to be able to specify this value.

`git config --get git-extras.feature.separator` and/or `-s|--separator` would be useful from my perspective.

This way I could do `git config git-extras.feature.separator -` and I would get `feature-branchname` instead of `feature/branchname`.

halostatue:
Closes #1069

This allows the use of a separator other than `/` for feature or other
alias branches. While a command-line option has been provided (`-s` or
`--separator`), this will most often be used via
`git-extras.feature.separator`.

### DIFF
--- a/bin/git-feature
+++ b/bin/git-feature
@@ -13,21 +13,23 @@ if [ -z "$branch_separator" ]; then
 fi
 
 merge_mode="--no-ff"
+finish=false
 declare -a argv
 while test $# != 0
 do
   case $1 in
   -a|--alias )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       shift # shift -a|-alias
       branch_prefix=$1
     else
-      argv+=("$1") # treat tail '-a' as <name>
+      echo >&2 "option $1 requires a value"
+      exit 1
     fi
     ;;
   -r|--remote )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       remote=$2
       shift
@@ -36,12 +38,13 @@ do
     fi
     ;;
   -s|--separator )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       branch_separator=$2
       shift
     else
-      argv+=("$1") # treat tail '-s' as <name>
+      echo >&2 "option $1 requires a value"
+      exit 1
     fi
     ;;
   --squash )
@@ -50,6 +53,15 @@ do
   --from )
     start_point=$2
     shift
+    ;;
+  -- )
+    # terminate argument parsing
+    shift
+    argv+=("$@")
+    break
+    ;;
+  finish )
+    finish=true
     ;;
   * )
     argv+=("$1")
@@ -63,18 +75,14 @@ concatargs() {
   branch="$branch_prefix$branch_separator$str"
 }
 
-if test "${argv[0]}" = "finish"; then
-  test -z "${argv[1]}" && echo "$branch_prefix" "<name> required." 1>&2 && exit 1
-  branch="$branch_prefix$branch_separator${argv[1]}"
+test -z "${argv[0]}" && echo "$branch_prefix" "<name> required." 1>&2 && exit 1
+
+concatargs "${argv[@]}"
+
+if "${finish}"
+then
   git merge ${merge_mode} "$branch" && git delete-branch "$branch"
 else
-  test -z "${argv[0]}" && echo "$branch_prefix" "<name> required." 1>&2 && exit 1
-  if test -n "${argv[1]}"; then
-    concatargs "${argv[@]}"
-  else
-    branch="$branch_prefix$branch_separator${argv[0]}"
-  fi
-
   if [[ -n $remote ]] && [[ -z $start_point ]]
   then
     git create-branch -r "$remote" "$branch"
--- a/bin/git-feature
+++ b/bin/git-feature
@@ -6,6 +6,12 @@ if [ -z "$branch_prefix" ]; then
    branch_prefix="feature"
 fi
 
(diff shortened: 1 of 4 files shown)


bin/git-feature
```diff
@@ -45,19 +60,19 @@ done
...
 concatargs() {
   str=$(IFS='-'; echo "$*")
-  branch="$branch_prefix"/$str
+  branch="$branch_prefix""$branch_separator"$str
 }
 
 if test "${argv[0]}" = "finish"; then
   test -z "${argv[1]}" && echo "$branch_prefix" "<name> required." 1>&2 && exit 1
-  branch="$branch_prefix"/"${argv[1]}"
+  branch="$branch_prefix""$branch_separator""${argv[1]}"
```
hyperupcall, line 68: It would be good to simplify quoting (here, and elsewhere) like so:

```suggestion
  branch="$branch_prefix$branch_separator${argv[1]}"
```

halostatue: Applied.

bin/git-feature
```diff
@@ -29,6 +35,15 @@ do
       remote="origin"
     fi
     ;;
+  -s|--separator )
+    if [[ -n $2 ]]
+    then
+      branch_separator=$2
+      shift
+    else
+      argv+=("$1") # treat tail '-s' as <name>
```
spacewander, line 44: I think we can remove this. Otherwise, adding new option will be cumbersome.

halostatue: What should we do if we get `git feature -s` or `git feature xyz -s`?

I had initially treated a missing second parameter as an error, but then I saw how `-a` was handled so I made it consistent.

My personal opinion is that the handling should be consistent, but that it should be as an error *unless* we also check for `--` to end option parsing.

Something like this (this is untested, but based on a pattern that I have written fairly frequently):

```diff
diff --git bin/git-feature bin/git-feature
index f33015c94902..0751ce32761a 100755
--- bin/git-feature
+++ bin/git-feature
@@ -18,16 +18,17 @@ while test $# != 0
 do
   case $1 in
   -a|--alias )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       shift # shift -a|-alias
       branch_prefix=$1
     else
-      argv+=("$1") # treat tail '-a' as <name>
+      echo >&2 "option $1 requires a value"
+      exit 1
     fi
     ;;
   -r|--remote )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       remote=$2
       shift
@@ -36,12 +37,13 @@ do
     fi
     ;;
   -s|--separator )
-    if [[ -n $2 ]]
+    if [[ -n $2 ]] && [[ $2 != -- ]]
     then
       branch_separator=$2
       shift
     else
-      argv+=("$1") # treat tail '-s' as <name>
+      echo >&2 "option $1 requires a value"
+      exit 1
     fi
     ;;
   --squash )
@@ -51,6 +53,12 @@ do
     start_point=$2
     shift
     ;;
+  -- )
+    # terminate argument parsing
+    shift
+    argv+=("$@")
+    break
+    ;;
   * )
     argv+=("$1")
     ;;
```

man/git-feature.md
```diff
@@ -1,81 +1,87 @@
...
+&lt;-a|--alias branch_prefix&gt;
 
-  use `branch_prefix` instead of `feature`
+Use `branch_prefix` instead of `feature`.
 
-  &lt;-r|--remote [remote_name]&gt;
+&lt;-s|--separator branch_separator&gt;
 
-  Setup a remote tracking branch using `remote_name`. If `remote_name` is not supplied, use `origin` by default.
+Use `branch_separator` instead of `/`.
```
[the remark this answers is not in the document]
halostatue: Added documentation for this and `git-extras.feature.prefix`.

bin/git-feature
spacewander: Let's try the diff you provide.

halostatue: I’ve added this as a second commit. When I was writing up the documentation to capture this properly, I came across some test cases which required a few more changes than originally provided.

If the further changes are accepted, I can squash this into one commit before merge with a cleaner commit message.

There are other changes that I would *personally* make (using the result of `shfmt`, using `(($#))` instead of `test $# != 0`, always using `[[…]]` instead of sometimes using `[…]`, a few other things), but I have avoided doing those in this PR because it would obfuscate the logic changes.

man/git-feature.md
```diff
@@ -2,42 +2,59 @@
...
-&lt;-a|--alias branch_prefix&gt;
+- `-a` <PREFIX>, `--alias` <PREFIX>:
+
+  The branch prefix to use, or `feature` if not supplied.
+
+- `-s` <SEPARATOR>, `--separator` <SEPARATOR>:
+
+  The separator to use for joining the branch prefix and the branch name, or `/` if not supplied.
+
+- `-r` [REMOTE_NAME], `--remote` [REMOTE_NAME]:
```
spacewander, line 27: I wonder why `[]` is used here while `<>` is used to wrap the argument above.

spacewander (approved):

man/git-feature.md
halostatue: Because `REMOTE_NAME` is optional; it does not require the second parameter but defaults to `origin` if just `-r` is provided. (As I’m reading it, `git create-branch` takes `-r` to create the branch and set its upstream as `REMOTE_BRANCH`. So, `-r` is just a pass-through to `git-create-branch(1)`.

spacewander: Thanks for your explanation

Issue #856, opened 2020-06-29
git psykorebase --continue is broken on macOS


lkrvscl:
# Problem

psykorebasing a branch on another, and having a resolved conflict:

```
$ git psykorebase --continue
Couldn't continue rebasing on ...
```

# Analysis
psykorebase depends on `sed`, however the macOS version does not have the feature needed.

psykorebase has this line under the `--continue` flag:
```
    set $(echo "$TARGET_BRANCH" | sed -e "s/-rebased-on-top-of-/\n/g")
```

which is supposed to be splitting the branch name into two lines, to be later used as branch names.

macOS `sed` has a problem with the `\n` substitution, as it expects the literal newline character after `\` not `n`, so the above code substitutes `n` instead of newline.

# Relevant macOS documentation

```bash
$ man sed
SED(1)                    BSD General Commands Manual                   SED(1)

NAME
     sed -- stream editor
...
     [2addr]s/regular expression/replacement/flags
...
             A line can be split by substituting a newline character into it.  To specify a newline character in the replacement string, precede it with a back-slash.
```

# sed compatibility comparison/guide
https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification

> In replacement strings used with the s command, assume that NO control-character escape sequences are supported, so, again, include control chars. as literals, as above.
> 
> Linux only:
> `sed 's/-/\t/' <<<$'a-b' # -> 'a<tab>b'`
> macOS and Linux:
> `sed 's/-/'$'\t''/' <<<'a-b'`
> `sed 's/-/'"$(printf '\t')"'/' <<<'a-b'`

In this case the working solution is a bit more complex (tested only on macOS):

```bash
    set $(echo "$TARGET_BRANCH" | sed -e "$(printf 's/-rebased-on-top-of-/\\\n/g')")
```

spacewander:
It seems the code wants to split `x-rebased-on-top-of-y` into `x` and `y`. Maybe we can use bash's "pattern substitution" to get rid of `sed`? Before the splitting, we need to check if the branch name matched the pattern.

rascalDan:
I've got the same problem with git-extra v7.0.0 and sed 4.9 on Gentoo Linux. Hopefully I'll get a chance to try to diagnose if it's the same root cause and try a few solutions.

rascalDan:
[...]

PR #1073, opened 2023-09-13
git-psykorebase: determine branch names with bash pattern substitution

rascalDan:
Removes the use of `sed` which should address https://github.com/tj/git-extras/issues/856


hyperupcall:
Closes #856

spacewander (approved):

PR #1075, opened 2023-09-15
fix: More improvements via ShellCheck

hyperupcall:


bin/git-changelog
```diff
@@ -146,6 +146,7 @@ _fetchCommitRange() {
   local start_tag="$2"
   local final_tag="$3"
 
+  # shellcheck disable=SC2086
   if [[ "$list_all" == true ]]; then
```
spacewander, line 150: Why add shellcheck disable here?

hyperupcall: This shellcheck comment applies to the `if` body, which contains use of variables that ate not quoted

hyperupcall: There are multiple violations and the body isn't very long, which is why I put a single one at the top

spacewander: > This shellcheck comment applies to the `if` body, which contains use of variables that ate not quoted

I see. Would you add a comment to indicate it's applied to the body?

hyperupcall: Added the comment :+1:

spacewander (approved):

PR #1077, opened 2023-09-21
test(git-alias): add its unit test

### LINKED ISSUE #1076: Test(git-alias): add the unit test for it
## File
[git-alias](https://github.com/tj/git-extras/blob/master/bin/git-alias)

## Test cases
* list all aliases
* list the matched aliases
* set an alias globally
* set an alias locally

The original `git-alias` did not implement the forth one, so refactor it together.

## Problem
<del>Maybe need to create a test.md to track the process?</del>

No more, but anyone has more?

vanpipy:
Close #1076

### DIFF
--- a/bin/git-alias
+++ b/bin/git-alias
@@ -1,16 +1,45 @@
 #!/usr/bin/env bash
 
+options=""
+
 usage() {
 cat <<HERE
-usage: git alias                         # list all aliases
-   or: git alias <search-pattern>        # show aliases matching pattern
-   or: git alias <alias-name> <command>  # alias a command
+usage: git alias [options]                          # list all aliases
+   or: git alias [options] <search-pattern>         # show aliases matching pattern
+   or: git alias [options] <alias-name> <command>   # alias a command
+options:
+
+    --global  Show or create alias in the system config
+    --local   Show or create alias in the repository config
 HERE
 }
 
+if [[ "$1" == "--local" || "$1" == "--global" ]]; then
+    options=$1
+    shift
+fi
+
 case $# in
-  0) git config --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort ;;
-  1) git alias | grep -e "$1" ;;
-  2) git config --global alias."$1" "$2" ;;
+  0)
+      if [[ -z "$options" ]]; then
+        git config --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort
+      else
+        git config "$options" --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort
+      fi
+      ;;
+  1)
+      if [[ -z "$options" ]]; then
+        git config --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort | grep -e "$1"
+      else
+        git config "$options" --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort | grep -e "$1"
+      fi
+      ;;
+  2)
+      if [[ -z "$options" ]]; then
+        git config alias."$1" "$2"
+      else
+        git config "$options" alias."$1" "$2"
+      fi
+      ;;
   *) >&2 echo "error: too many arguments." && usage && exit 1 ;;
 esac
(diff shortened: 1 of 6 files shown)


man/git-alias.md
```diff
@@ -6,13 +6,23 @@ git-alias(1) -- Define, search and show aliases
 `git-alias`  
 `git-alias` &lt;search-pattern&gt;  
 `git-alias` &lt;alias-name&gt; &lt;command&gt;  
+`git-alias` --global &lt;alias-name&gt; &lt;command&gt;  
+`git-alias` --local &lt;alias-name&gt; &lt;command&gt;  
 
 ## DESCRIPTION
 
   List all aliases, show one alias, or set one (global) alias.
```
spacewander, line 14:
```suggestion
  List all aliases, show one alias, or set one (global or local) alias.
```

man/git-alias.md
```diff
@@ -6,13 +6,23 @@ git-alias(1) -- Define, search and show aliases
...
 
 ## OPTIONS
 
+  &lt;--global&gt;
+
+  write and read the config from ~/.gitconfig file only.
```
spacewander, line 20: Would be better to focus on the behavior of the cmd, instead of the implementation detail.

vanpipy: Got this idea, I will refactor the description.

man/git-alias.md
vanpipy: Done

bin/git-alias
```diff
@@ -1,16 +1,31 @@
...
+    shift
+fi
+
 case $# in
-  0) git config --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort ;;
-  1) git alias | grep -e "$1" ;;
-  2) git config --global alias."$1" "$2" ;;
+  0) git config $options --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort ;;
+  1) git config $options --get-regexp 'alias.*' | sed 's/^alias\.//' | sed 's/[ ]/ = /' | sort | grep -e "$1" ;;
+  2) git config $options alias."$1" "$2" ;;
```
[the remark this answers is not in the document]
vanpipy: Same solution as above cause the `$parameter` is string already.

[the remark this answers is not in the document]
spacewander: The quote added here is just to make the shellcheck happy. Without it, the shellcheck will report `Double quote to prevent globbing and word splitting.`

man/git-alias.md
```diff
@@ -6,13 +6,23 @@ git-alias(1) -- Define, search and show aliases
 `git-alias`  
 `git-alias` &lt;search-pattern&gt;  
 `git-alias` &lt;alias-name&gt; &lt;command&gt;  
+`git-alias` [--global] &lt;alias-name&gt; &lt;command&gt;
```
spacewander, line 9: It seems we missed a case that the scope parameter is used with `search-pattern`?

man/git-alias.md
```diff
@@ -6,13 +6,23 @@ git-alias(1) -- Define, search and show aliases
...
 
 ## DESCRIPTION
 
   List all aliases, show one alias, or set one (global) alias.
 
 ## OPTIONS
 
+  --global
+
+  Show aliased matched in the system config
```
spacewander, line 20: Let's document that we can also use the option to alias a command in different scope.

bin/git-alias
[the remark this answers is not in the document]
vanpipy: Understood, but in fact the `git config ""` will get the `error: key does not contain a section:  `. So I did not understand the happy shellcheck indeed.

man/git-alias.md
vanpipy: Mean to give an example?

bin/git-alias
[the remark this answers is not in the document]
spacewander: The `$option` won't contain any space, but shellcheck is not smart enough to figure out it. What we do is just to make it silent.

man/git-alias.md
spacewander: @vanpipy 
It didn't.

man/git-alias.md
spacewander:
```suggestion
  Show or create alias in the system config
```

spacewander (approved):

spacewander:
@vanpipy 
Merged. Thanks!

PR #1080, opened 2023-09-24
makefile: Allow bypassing conflict check

### LINKED ISSUE #295: Manpage for `git-obliterate` missing.
The Readme references a `git-obliterate`-command, while the related man-page is non-existant.
I would add it myself, but I'm unsure about the formatting conventions.

### LINKED ISSUE #375: set-upstream utility
would anyone be interested in a utility that does this? i can make a PR if so.

``` sh
#!/usr/bin/env sh
# git-pushup
#   Pushes the current branch via --set-upstream.
#   It's a shortcut for `git push --set-upstream origin <current branch>`.

remote=$1
[ -z $remote ] && remote=origin

git push --set-upstream $remote `git rev-parse --abbrev-ref HEAD`
```

### LINKED ISSUE #541: Install not possible in headless environment due to conflicting git alias
The conflict prompt prevents installing git-extras in a headless environment.

```
git-undo conflicts with an alias, still install it and disable the alias? [y/n]
```

There should be a flag or option to always install or skip these type of conflicting commands.

### LINKED ISSUE #968: Show option: expand alias in terminal before it gets executed (like `abbr -a` in fish shell)
### Feature request
For only .gitconfig file, i want this feature
```
[alias]
    cl = clone
    ci = commit
```

- After typing `git cl` and press <kbd>space</kbd> key, alias should expand
  e.g `cl` option will turn into `clone`

- For some reason, if i made a mistake, i can press <kbd>Esc</kbd> or <kbd>Ctrl+space</kbd> to undo alias expansion
  e.g `clone` will revert back to `cl`

- Again correct it from `cl` to `ci` and press <kbd>space</kbd> key
  e.g `git ci` will turn into `git commit`

hyperupcall:
Introduces a `SKIP_CONFLICT_CHECK` so installs can always be unattended.

- Changes wording of `conflicts with an alias, still install it and idsable the alias? [y/n]`
  - Removes the part that says "disable the alias" because any existing aliases aren't (and shouldn't be) disabled
- Adds `test` to the `.PHONY` target
- Improves quoting for paths with spaces
- Improves variable names

And,

- Closes #541
- Closes #375 (stale, built into git)
- Closes #295 (stale, already fixed)
- Closes #968 (stale, out of scope)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,22 +45,24 @@ install: check
 	@if [ "$(INSTALL_VIA)" = brew ]; then \
 		git apply brew-release.patch || { echo "Can't apply brew release patch"; exit 1; } \
 	fi
-	@mkdir -p $(DESTDIR)$(MANPREFIX)
-	@mkdir -p $(DESTDIR)$(BINPREFIX)
-	@echo "... installing bins to $(DESTDIR)$(BINPREFIX)"
-	@echo "... installing man pages to $(DESTDIR)$(MANPREFIX)"
+	mkdir -p $(DESTDIR)$(MANPREFIX)
+	mkdir -p $(DESTDIR)$(BINPREFIX)
+	@echo '... installing bins to '$(DESTDIR)$(BINPREFIX)
+	@echo '... installing man pages to '$(DESTDIR)$(MANPREFIX)
 	$(eval TEMPFILE := $(shell mktemp -q $${TMPDIR:-/tmp}/git-extras.XXXXXX 2>/dev/null || mktemp -q))
 	@# chmod from rw-------(default) to rwxrwxr-x, so that users can exec the scripts
 	@chmod 775 $(TEMPFILE)
 	$(eval EXISTED_ALIASES := $(shell \
-		git config --get-regexp 'alias.*' | awk '{print "git-" substr($$1, 7)}'))
+		git config --get-regexp 'alias\..*' | awk '{print "git-" substr($$1, 7)}'))
 	@$(foreach COMMAND, $(COMMANDS), \
-		disable=''; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,22 +45,24 @@ install: check
 	@if [ "$(INSTALL_VIA)" = brew ]; then \
 		git apply brew-release.patch || { echo "Can't apply brew release patch"; exit 1; } \
 	fi
-	@mkdir -p $(DESTDIR)$(MANPREFIX)
-	@mkdir -p $(DESTDIR)$(BINPREFIX)
-	@echo "... installing bins to $(DESTDIR)$(BINPREFIX)"
-	@echo "... installing man pages to $(DESTDIR)$(MANPREFIX)"
+	mkdir -p $(DESTDIR)$(MANPREFIX)
```
spacewander, line 48: It seems that these cmds will be broken when the directories contain space. If it is true, would you also improve them in  this PR?

spacewander (approved):

Makefile
hyperupcall: From my testing it seems you're not supposed to quote variable interpolations like this because the values will automatically be escaped. that's why I changed the quotes on lines 50 and 51 - to no longer include `$(DESTDIR)$(BINPREFIX)`.

hyperupcall: Interestingly, single quotes are OK, but double quotes breaks things.

Looking at other popular projects, it seems the norm is not to quote things: https://github.com/torvalds/linux/blob/6465e260f48790807eef06b583b38ca9789b6072/Makefile#L1283, which seems to line up with other information on how to use GNU Make

spacewander: I see

spacewander:
Merged. Thanks!

PR #1084, opened 2023-10-13
test(git-archive-file): add unit test

### LINKED ISSUE #1081: Test(git-archive-file): add unit test for it
## File
[git-archive-file](https://github.com/tj/git-extras/blob/master/bin/git-archive-file)

## Testcases
* If on the branch `tags*`, create a zip file which named `{dirname}.{git describe}.zip` in the root.
* If on any branch not like `tags*`
    + and the branch is not `git_extra_default_branch`, create a zip file which is named `{dirname}.{git describe --always --long}.{branch_name}.zip` in the root.
    + and the branch is `git_extra_default_branch`, create a zip file which is named `{dirname}.{git describe --always --long}.zip` in the root.
* Base above, replace the `/` or `\` with `-` when the final filename has one.
    + branch and version name maybe include the slash `/`, but cannot include the backslash `\`
    + dirname maybe include the backslash `\`, but cannot include the slash `/`

## References
* [git_extra_default_branch](https://github.com/tj/git-extras/blob/master/helper/git-extra-utility#L8)

vanpipy:
Close #1081

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -1,21 +1,23 @@
-import os
-import subprocess
-import shutil
-import tempfile
-import git
+import os, subprocess, stat, shutil, tempfile, git
+
+CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
+GIT_EXTRAS_BIN = os.path.join(CURRENT_DIR, "..", "bin")
+GIT_EXTRAS_HELPER = os.path.join(CURRENT_DIR, "..", "helper")
 
 def invoke_git_extras_command(name, *params):
-    current_dir = os.path.dirname(os.path.abspath(__file__))
-    git_extras_bin = os.path.join(current_dir, "..", "bin")
-    script = [os.path.join(git_extras_bin, name), *params]
+    script = [os.path.join(GIT_EXTRAS_BIN, name), *params]
     print(f"Run the script \"{script}\"")
     return subprocess.run(script, capture_output=True)
 
 class TempRepository:
     def __init__(self, repo_work_dir = None):
+        self._system_tmpdir = tempfile.gettempdir()
         if repo_work_dir == None:
             repo_work_dir = tempfile.mkdtemp()
+        else:
+            repo_work_dir = os.path.join(self._system_tmpdir, repo_work_dir)
         self._cwd = repo_work_dir
+        self._tempdirname = self._cwd[len(self._system_tmpdir) + 1:]
         self._git_repo = git.Repo.init(repo_work_dir)
         self._files = []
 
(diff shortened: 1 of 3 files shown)


tests/test_archive_file.py
```diff
@@ -0,0 +1,40 @@
+import os
+
+# Testcases issue https://github.com/tj/git-extras/issues/1081
+class TestGitAbort:
```
spacewander, line 4: We need to use a different name

bin/git-archive-file
```diff
@@ -26,7 +26,6 @@ fi
 
 # rename invalid chars for the file path
 FILENAME=${FILENAME//\//-}
-FILENAME=${FILENAME//\\/-}
```
spacewander, line 29: This line is different from the above

tests/helper.py
```diff
@@ -63,6 +66,33 @@ def teardown(self):
...
+    def invoke_installed_extras_command(self, name, *params):
+        command_name = "git-" + name
+        print(f"Invoke the git-extras command - {command_name}")
+        origin_extras_command = os.path.join(GIT_EXTRAS_BIN, command_name)
+        temp_extras_command = os.path.join(self._cwd, command_name)
+        helpers = [
+                os.path.join(GIT_EXTRAS_HELPER, "git-extra-utility"),
+                os.path.join(GIT_EXTRAS_HELPER, "is-git-repo")]
+
+        if os.path.exists(temp_extras_command) == False:
```
spacewander, line 82:
```suggestion
        if not os.path.exists(temp_extras_command):
```

tests/test_archive_file.py
vanpipy: Fixed

bin/git-archive-file
vanpipy: Recovered

tests/helper.py
vanpipy: Done

bin/git-archive-file
vanpipy: Need more time for this case. The preset playground did not support the preset dirname.

vanpipy: Alright, i take a wrong direction to figure the point out. There is only one way to pass the filter truly that `$BRANCH` has the slash or backslash. I will leave this in the issue.

vanpipy: Done

spacewander (approved):
Would you rebase to the master branch before I can merge it?

vanpipy:
Done and great, all checks passed.

tests/test_archive_file.py
```diff
@@ -0,0 +1,71 @@
+import os, pytest
+
+# Testcases issue https://github.com/tj/git-extras/issues/1081
```
spacewander, line 3: We can remove the link now?

vanpipy: Removed

spacewander (approved):

spacewander:
Merged. Thanks!

PR #1091, opened 2023-10-20
Allow `git cp` to create destination folder automatically

### LINKED ISSUE #1089: Create destination folder during `git cp` if it doesn't exist
Currently when running `git cp <src_folder/src_file> <dest_folder/dest_file>`, the `<dest_folder>` must pre-exist for the command to succeed. 

Suggestion: create `<dest_folder>` automatically if needed.

weiw005:
fixed #1089

### DIFF
--- a/bin/git-cp
+++ b/bin/git-cp
@@ -38,13 +38,15 @@ else
         exit 40
     fi
 
-    DESTINATION_DIR="$(dirname "$DESTINATION_FILENAME")"
+    if [[ "$DESTINATION_FILENAME" == */ ]]; then
+      echo 1>&2 "$DESTINATION_FILENAME is not a file path."
+      exit 80
+    fi
+    DESTINATION_DIR="${DESTINATION_FILENAME%/*}"
     if [ ! -d "$DESTINATION_DIR" ]; then
-      echo "Creating destination directory: $DESTINATION_DIR"
-      mkdir -p "$DESTINATION_DIR"
-      if [ $? -ne 0 ]; then
+      if ! mkdir -p "$DESTINATION_DIR"; then
         echo 1>&2 "Failed to create destination directory: $DESTINATION_DIR"
-        exit 80
+        exit 160
       fi
     fi
 
--- a/bin/git-cp
+++ b/bin/git-cp
@@ -42,8 +42,8 @@ else
       echo 1>&2 "$DESTINATION_FILENAME is not a file path."
       exit 80
     fi
-    DESTINATION_DIR="${DESTINATION_FILENAME%/*}"
-    if [ ! -d "$DESTINATION_DIR" ]; then
+    if [[ "$DESTINATION_FILENAME" == */* ]]; then
+      DESTINATION_DIR="${DESTINATION_FILENAME%/*}"
       if ! mkdir -p "$DESTINATION_DIR"; then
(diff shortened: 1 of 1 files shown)


hyperupcall (changes requested):
It looks like there is a race and an extra subshell here. What about something like this?:

```sh
DESTINATION_FILENAME=${DESTINATION_FILENAME%/}
if ! mkdir -p "${DESTINATION_FILENAME%/*}"; then
  echo 1>&2 "Failed to create destination directory: $DESTINATION_DIR"
  exit 80
fi
```

weiw005:
> It looks like there is a race and an extra subshell here. What about something like this?:
> 
> ```shell
> DESTINATION_FILENAME=${DESTINATION_FILENAME%/}
> if ! mkdir -p "${DESTINATION_FILENAME%/*}"; then
>   echo 1>&2 "Failed to create destination directory: $DESTINATION_DIR"
>   exit 80
> fi
> ```

Thanks! Removed the subprocess command. PTAL.

(BTW my knowledge with shell script is very limited, so open to further suggestions)

bin/git-cp
```diff
@@ -38,6 +38,18 @@ else
         exit 40
     fi
 
+    if [[ "$DESTINATION_FILENAME" == */ ]]; then
+      echo 1>&2 "$DESTINATION_FILENAME is not a file path."
+      exit 80
+    fi
+    DESTINATION_DIR="${DESTINATION_FILENAME%/*}"
+    if [ ! -d "$DESTINATION_DIR" ]; then
```
spacewander, line 46: We can use `mkdir -p` directly as the cmd won't return error if the dir exists?

bin/git-cp
```diff
@@ -38,6 +38,18 @@ else
         exit 40
     fi
 
+    if [[ "$DESTINATION_FILENAME" == */ ]]; then
+      echo 1>&2 "$DESTINATION_FILENAME is not a file path."
+      exit 80
+    fi
+    DESTINATION_DIR="${DESTINATION_FILENAME%/*}"
```
spacewander, line 45: What if the DESTINATION_FILENAME doesn't contain the directory, like `./foo`?

bin/git-cp
weiw005: Make sense. Removed this check.

bin/git-cp
weiw005: Good question! Although the new code still works for `./foo` (because `DESTINATION_DIR` will simply be `./` and creating it is a harmless no-op), another case that I tested did lead to incorrect behavior which is when the `DESTINATION_FILENAME ` is just a filename. The previous revision would create a folder of that name, and make a copy of the source file into that folder (rather than copy to a file of that name).

I've now switched the condition to be checking whether `DESTINATION_FILENAME` contains any slash characters - if not the folder creation is skipped.

spacewander (approved):

spacewander:
@hyperupcall 
Please take a look when you have free time.

hyperupcall (approved):

weiw005:
@spacewander @hyperupcall Thanks for the review! Would either of you be able to click the merge button for me? I don't have write access to this repo.

hyperupcall:
@weiw005 @spacewander will be able to do it

spacewander:
Merged. Thanks!
Also thanks @hyperupcall for the review :)

PR #1096, opened 2023-10-24
git-cp: don't stop execution when merge.ff is not set

spacewander:


hyperupcall:
The title is a bit confusing - with these changes, doesn't execution now stop if `merge.ff` is not set? That's what `|| true` is trying to prevent, right?

spacewander:
@hyperupcall 
Will you approve this PR if you think it's OK?

hyperupcall:
Yes, approved, this is a great change :+1:

Issue #1102, opened 2023-11-03
Discuss: add the python project attribute to git extras repository


vanpipy:
I want to enhance the dependency management of the python for git extras via [poetry](https://github.com/python-poetry/poetry), but I don't know it make sense or just a big hammer at this moment, so I create this.

@hyperupcall and @spacewander , please take a look here.

vanpipy:
Emm.., I think is useless to add the package manager for git extras repo cause the installation is a problem or just waste the time and the information about the test in the readme should be moved to a new file in the test directory, it is special for the dev who want to create some testcases.

hyperupcall:
If we use Python, in my opinion Poetry is a no-brainer; it makes things easier to use.

I'm not convinced that we need `testpath`. If we need to mock a command, I think we should do it ourselves. But I'm not the one writing the majority of tests.

I agree with your points that instructions for setting up Python for testing should not be included in the main `README.md`.

spacewander:
> Emm.., I think is useless to add the package manager for git extras repo cause the installation is a problem or just waste the time and the information about the test in the readme should be moved to a new file in the test directory, it is special for the dev who want to create some testcases.

Agree with that.

> If you would like to add it, you certainly have my vote!

Vote +1 for this too.

vanpipy:
Thanks guys and I will try to restruct the test env for git extras later.

> I'm not convinced that we need testpath. 

Because of the definition of the `git-browse` that it will lead the user to a website which display current repository. Normally, the `xdg-open` can help us to do like that, but the `xdg-open` is not the part of our functions, we just use it in a special environment, here is linux or unix. I call the part like `xdg-open` as external API and the external API should be mocked when do unit tests. So I think we should has a mock library to help us. [Mocking system commands](https://testpath.readthedocs.io/en/latest/commands.html) is a great way to detach the external API for `git-browse` unit test.

hyperupcall:
@vanpipy I don't disagree that we need a way to mock commands, and I did mention:

> If we need to mock a command, I think we should do it ourselves.

The simple thing we need to do is add a directory, `tests/bin` to the `PATH` only when testing, which contains the mocked commands. I think this approach is more clear and approachable for someone not familiar with the codebase; it's an approach that isn't uncommon and is not coupled to a particular library.

vanpipy:
[...]

PR #1103, opened 2023-11-06
Overhaul CONTRIBUTING document

hyperupcall:
Inspired by [this](https://github.com/tj/git-extras/issues/1102#issuecomment-1793283840) comment, this updates the `CONTRIBUTING.md` and moves the information about testing dependencies to itself.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,9 +6,11 @@ Thanks for contributing! Please read this document before you make a PR.
 
 Any changes must support the following platforms:
 
-- macOS, Linux, BSDs (You may need to browse their man page)
-- Bash 3.2+ (If you aren't sure, see the [Bash changelog](https://git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=devel) to see a detailed list of version support for features)
-- Git 2.1+
+- macOS
+- Linux
+- OpenBSD (You may need to browse their man page)
+
+Your change must also be compatible with the dependency constraints that we specify in [Installation](./Installation.md). If you aren't sure if a feature is compatible, check the manual or release notes. For example, the Bash changelog is [here](https://git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=devel).
 
 If you aren't able to test your new command on a platform, make that clear in your PR; someone else may be able to test it on their system.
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -1,22 +1,38 @@
...
-<sup>*</sup>If you aren't able to test your new command on a platform,
-make that clear in your PR and someone else may be able to test it on their system.
+## Supported Platforms
 
-## To submit a new command, you should
+Any changes must support the following platforms:
 
-Let's assume your new command is named `foo`.
+- macOS, Linux, BSDs (You may need to browse their man page)
+- Bash 3.2+ (If you aren't sure, see the [Bash changelog](https://git.savannah.gnu.org/cgit/bash.git/tree/NEWS?h=devel) to see a detailed list of version support for features)
```
spacewander, line 10: Thanks for updating the stale doc. Now we need Bash 4+, and there is other new dependencies: https://github.com/tj/git-extras/blob/main/Installation.md#dependencies

CONTRIBUTING.md
```diff
@@ -1,22 +1,38 @@
...
+- Git 2.1+
```
spacewander, line 11: Maybe it is time to update the git version?

CONTRIBUTING.md
hyperupcall: Oops, I missed Installation.md, I'll reference that instead of duplicating the same information in CONTRIBUTING

CONTRIBUTING.md
hyperupcall: I'll update it to Git 2.17.1, which is a less ancient Git version and is the lowest common denominator between Ubuntu 18 and Debian 10. I'm not sure if it is a good idea to increase it further than that, because I don't know which versions we truly support? Would we have to check all our uses of `git` to find out, or do you have an idea yourself?

spacewander: 2.17 LGTM

spacewander:
Please make the CI pass, thanks!

hyperupcall:
@spacewander It's fixed now - sorry about that

spacewander:
> @spacewander It's fixed now - sorry about that

No need to be sorry for it.

spacewander (approved):

PR #1104, opened 2023-11-06
Improve defaults for testing suite

### LINKED ISSUE #486: Git flow
Need commands such as `git feature` for git flow

image

``` bash
git bugfix AD-123 Drop document after upload
git hotfix AD-543 User not authorise
```

### LINKED ISSUE #818: Request: update epel-6 / 7 to include git-extras 5.x
I have a need to use epel-6 rather than the latest version and would like the latest git-extras installed. Can this be added please?

### LINKED ISSUE #1083: Gustavo
Como subir um projeto pelo celular??

hyperupcall:
I started to test @vanpipy's changed on my machines. I think this changes will streamling the testing process for other people as well.

I also wish to close the following issues:

Closes #1083 (spam)
Closes #818 (question answered)
Closes #486 (out of scope)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,8 +43,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          ref: 'main'
       - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          ref: 'main'
       - name: Set up Python 3.10
         uses: actions/setup-python@v4
         with:
--- a/tests/test_git_archive_file.py
+++ b/tests/test_git_archive_file.py
@@ -28,8 +28,8 @@ class TestGitArchiveFile:
 
     def test_archive_file_on_any_not_tags_branch_with_default_branch(self, temp_repo):
         git = temp_repo.get_repo_git()
-        git.checkout("main")
-        git.config("git-extras.default-branch", "main")
+        git.checkout("default")
+        git.config("git-extras.default-branch", "default")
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -42,7 +42,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          ref: 'main'
```
spacewander, line 47: According to https://github.com/actions/checkout#usage,
should we leave this empty instead of hardcoding it as main? What problem does it solve?

tests/test_git_archive_file.py
```diff
@@ -28,8 +28,8 @@ def test_archive_file_on_any_not_tags_branch_without_default_branch(self, temp_r
 
     def test_archive_file_on_any_not_tags_branch_with_default_branch(self, temp_repo):
         git = temp_repo.get_repo_git()
-        git.checkout("master")
-        git.config("git-extras.default-branch", "master")
+        git.checkout("main")
```
spacewander, line 31: If I am correct, I think the branch here is set to master intently. Because the current branch of the  `temp_repo.get_repo_git()` seems to be already `main`.

@vanpipy 
What do you think?

hyperupcall: Oh I think I know what this is - I have customized `init.defaultBranch` to be `main`. Which is why I had to make that change. I'll fix the name of the branches used in the test to be less confusing

.github/workflows/ci.yml
hyperupcall: Yesterday I seemingly needed it for CI to pass (it didn't make sense at the time, I just did it). But I just removed it, things seem to work today

hyperupcall:
@spacewander this should be ready - The only change to CI was using `checkout@v4` (which still works) and updated the `git.Repo.init(repo_work_dir, b="default")` call to manually set the default branch. I set it to `default`, which is different than `main` and `master` to make things a bit less confusing (in my opinion)

spacewander (approved):

PR #1105, opened 2023-11-07
Added --full flag to "git-count"

havocesp:
- Added --full flag to count all existing commits from a local repo.

Example:

```bash

# git count without flags (count from current commit to last one)
$ git count
total 244

# git count with `--all` flag (same behaviour that previous but show a list of contributors)
$ git count --all
- James
- Foo
- Bar
243

# git count with `--full` flag (count all existing commits)
$ git count --full
277
```


bin/git-count
```diff
@@ -2,7 +2,8 @@
 
 if test "$1" = "--all"; then
   git shortlog -n -s | awk '{print substr($0,index($0,$2)) " (" $1 ")"}'
-  echo
```
spacewander, line 5: We should also output total for `--all`.

etc/bash_completion.sh
```diff
@@ -64,7 +64,7 @@ _git_contrib(){
 }
 
 _git_count(){
-  __gitcomp "--all"
+  __gitcomp "--all --full"
```
spacewander, line 67: Let's update other completions like zsh

spacewander (commented):
Let's update the doc under https://github.com/tj/git-extras/tree/main/man

etc/bash_completion.sh
havocesp: I have never use `zsh` but I try.

bin/git-count
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
-if test "$1" = "--all"; then
+if test "$1" = "--all" || test "$1" = "-a"; then
   git shortlog -n -s | awk '{print substr($0,index($0,$2)) " (" $1 ")"}'
-  echo
+elif test "$1" = "--full" || test "$1" = "-f";then
```
spacewander, line 5:
```suggestion
elif test "$1" = "--full" || test "$1" = "-f"; then
```

bin/git-count
spacewander: Without this change, the output will be different.

spacewander: @havocesp 
This comment is not resolved yet.

man/git-count.md
```diff
@@ -23,6 +24,12 @@ git-count(1) -- Show commit count
     $ git count
 
     total 1844
+ 
+ Output full commits total:
```
spacewander, line 28: Let's add more words to explain the difference between the regular total and the full total.

havocesp: The regular total count from current commit to the first one and the full total count all commits available at repository.

PR #1107, opened 2023-11-09
Man git count

havocesp:
Updated also man files about git-count --full flag.


hyperupcall:
Shouldn't this be added to #1105?

hyperupcall:
Closing this as this seems to be a duplicate of #1105

Issue #1106, opened 2023-11-08
git standup recursive search fails if -L missing


rm--:
[macOS 14.1.1, zsh, git-extras v7.1.0]

If I run `git standup` in a parent directory of some git-repositories.
`git standup` fails with:

```
find: : No such file or directory
```

If I run `bash -x /usr/local/bin/git-standup`, I see:

```
++ find '' . -maxdepth 2 -mindepth 0 -name .git
find: : No such file or directory
```

That's because of the quoted empty value of: `$INCLUDE_LINKS` in 
`PROJECT_DIRS=$(find "$INCLUDE_LINKS" . -maxdepth "$MAXDEPTH" -mindepth 0 -name .git)`
(https://github.com/tj/git-extras/blob/main/bin/git-standup#L243)

Since `$INCLUDE_LINKS` is only set if `-L` option is used, IMO there should be a handling of the empty value or the variable should have some valid value which `find` can handle.
(https://github.com/tj/git-extras/blob/main/bin/git-standup#L124)

spacewander:
PR is welcome!

PR #1110, opened 2023-11-11
fix: git-standup find error (#1106)

rm--:


bin/git-standup
```diff
@@ -240,7 +240,11 @@ if [[ $in_git_repo != 0 ]]; then
     ## Set delimiter to newline for the loop
     IFS=$'\n'
     ## Recursively search for git repositories
-    PROJECT_DIRS=$(find "$INCLUDE_LINKS" . -maxdepth "$MAXDEPTH" -mindepth 0 -name .git)
```
hyperupcall, line 243: Thank you for the catch. Another fix would be to remove the quotes from `$INCLUDE_LINKS`

rm--: I implemented this way because of the [shellcheck](https://github.com/koalaman/shellcheck) output:

```
➜ shellccheck git-extras/bin/git-standup

In git-extras/bin/git-standup line 244:
        PROJECT_DIRS=$(find . -maxdepth $MAXDEPTH -mindepth 0 -name .git)
                                        ^-------^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean:
        PROJECT_DIRS=$(find . -maxdepth "$MAXDEPTH" -mindepth 0 -name .git)

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

hyperupcall: I see - yeah that would be a false positive

rm--: I can change it to your suggestion. It's your decision.
But until now shellcheck like the file. ;)

hyperupcall: You can use something like `#shellcheck disable=SC2086` to disable the spurious warning

hyperupcall: thanks!

hyperupcall (approved):

spacewander (approved):

PR #1111, opened 2023-11-11
Add more comprensive dependencies

hyperupcall:
From [this](https://github.com/tj/git-extras/commit/7a65311b1c7e074760c61aa2ebc0f8bd3665bc51#r132308022) comment, some dependencies were missing. I have added more dependencies that are from `util-linux` or require manual installation on some systems.

### DIFF
--- a/Installation.md
+++ b/Installation.md
@@ -8,6 +8,13 @@ You may need to install them manually. They are:
 * bash 4.0+
 * Git 2.17+
 * `column`
+* `awk`
+* `find`
+* `tput`
+* cURL (only required for `git-fork`, `git-ignore-io`, `git-pull-request`)
+* `ps` (only required for `git-changelog`)
+* `rsync` (only required for `git-rscp`, `git-scp`)
+* `xargs` (only required for `git-delete-merged-branches`, `git-force-clone`, `git-sed`, `git-scp`)
 
 If `bash --version` shows a lower version, you need to update it.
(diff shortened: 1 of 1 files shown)


Installation.md
```diff
@@ -8,6 +8,14 @@ You may need to install them manually. They are:
 * bash 4.0+
 * Git 2.17+
 * `column`
+* `awk`
+* `column`
+* cURL
```
kang8, line 13: typo
```suggestion
* `curl`
```

kang8 (changes requested):

Installation.md
hyperupcall: cURL works too: https://en.wikipedia.org/wiki/CURL

Installation.md
```diff
@@ -8,6 +8,14 @@ You may need to install them manually. They are:
 * bash 4.0+
 * Git 2.17+
 * `column`
+* `awk`
+* `column`
```
spacewander, line 12: The column shows up twice?

Installation.md
```diff
@@ -8,6 +8,14 @@ You may need to install them manually. They are:
...
+* cURL
+* `find`
+* `ps`
+* `rsync`
```
spacewander, line 16: Let's doc what commands are using rsync. There are few commands that are using rsync, if the user doesn't need them, she doesn't need rsync too.

hyperupcall: Sounds good, I did that for a few more commands :+1:

spacewander (approved):

PR #1115, opened 2023-11-19
refactor(git-authors): add unit test for option `--output`

### LINKED ISSUE #1100: Refactor(git-authors): support the option `--output` and controlled `--no-email`
## File
[git-authors](https://github.com/tj/git-extras/blob/master/bin/git-authors)

## Support new usecases
1. `git authors --output TARGET_AUTOHRS_FILE`
2. `git authors --output TARGET_AUTOHRS_FILE --no-email`

Refactor the original `git-authors`.

vanpipy:
Close #1100

### DIFF
--- a/bin/git-authors
+++ b/bin/git-authors
@@ -24,7 +24,7 @@ authors() {
 }
 
 if [[ -z "$file" ]]; then
-  file=$(find . -mindepth 1 -maxdepth 1 -iregex '.*\(authors\|contributors\).*' | head -n1)
+  file=$(find . -mindepth 1 -maxdepth 1 \( -iname '*authors*' -o -iname '*contributors*' \) | head -n1)
   if [[ -z "$file" ]]; then
     file="AUTHORS"
   fi
--- a/bin/git-authors
+++ b/bin/git-authors
@@ -44,7 +44,7 @@ while [[ $# -gt 0 ]]; do
       no_email=1
       ;;
     --output )
-      if [[ -n $2 ]] && [[ $2 =~ ^(\.\/|\.\.\/|\/)?([a-zA-Z]\/?)+$ ]]; then
+      if [[ -n $2 ]] && [[ $2 =~ ^(\.\/|\.\.\/|\/)?([a-zA-Z_](\/)?)+$ ]]; then
         output=$2
         shift
       else
--- a/bin/git-authors
+++ b/bin/git-authors
@@ -1,40 +1,20 @@
 #!/usr/bin/env bash
 
-LIST=false
-NO_EMAIL=false
-FILE=""
+list=0
+no_email=0
+output=""
+file=""
 
-while [[ $# -gt 0 ]]; do
-  case $1 in
-    -l|--list )
-      LIST=true
-      shift
-      ;;
-    --no-email )
-      NO_EMAIL=true
-      shift
-      ;;
-    * )
-      break
-  esac
-done
-
-if ! $LIST; then
-  FILE=$1
-  if [ -z "$FILE" ]; then
-    FILE=$(find . -mindepth 1 -maxdepth 1 \( -iname '*authors*' -o -iname '*contributors*' \) | head -n1)
-    if [ -z "$FILE" ]; then
-      FILE='AUTHORS'
-    fi
-  fi
+if [[ $# = 1 ]] && [[ "$1" = "--no-email" ]]; then
+  echo >&2 "--no-email option only can be used with --list | -l | --output"
+  exit 1
 fi
 
 #
 # list authors sorted by number of commits (descending).
 #
-
 authors() {
-  if $NO_EMAIL; then
+  if [[ $no_email = 1 ]]; then
     # email will be used to uniq authors.
     git shortlog HEAD -sne | awk '{$1=""; sub(" ", ""); print}' | awk -F'<' '!x[$1]++' | awk -F'<' '!x[$2]++' \
       | awk -F'<' '{gsub(/ +$/, "", $1); print $1}'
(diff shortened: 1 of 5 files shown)


bin/git-authors
```diff
@@ -1,40 +1,20 @@
...
-  FILE=$1
-  if [ -z "$FILE" ]; then
-    FILE=$(find . -mindepth 1 -maxdepth 1 \( -iname '*authors*' -o -iname '*contributors*' \) | head -n1)
-    if [ -z "$FILE" ]; then
-      FILE='AUTHORS'
-    fi
-  fi
+if [[ $# = 1 ]] && [[ "$1" = "--no-email" ]]; then
+  echo >&2 "--no-email option only can be used with --list | -l | --output"
+  exit 1
```
hyperupcall, line 10: Checks can be good, but I'm looking at a [previous diff](https://github.com/vanpipy/git-extras/blob/263dcb4a79fa8e21e64f9d0f6a007148a6cec9d6/bin/git-authors), and it seems like this is a breaking change? `git authors --no-email` seems to have written to file `AUTHORS`, excluding emails?

man/git-authors.md
```diff
@@ -3,7 +3,9 @@ git-authors(1) -- Generate authors report
 
 ## SYNOPSIS
 
+`git-authors`
 `git-authors` [-l, --list] [--no-email]
+`git-authors` [--output] &lt;filename&gt; [--no-email]
```
hyperupcall, line 8: Thank you for adding docs. You don't need to escape `<` and `>`, and other characters in Markdown as you would in HTML.

hyperupcall (changes requested):
Thanks for your contribution, I have a few first comments. It would be easier to review, and less error-prone if you didn't add tests at the same time as changing the whole script. Or, if you did so, to make only the necessary changes.

bin/git-authors
vanpipy: The previous implement did not include this feature, just dump all authors has the email into `AUTHORS`. There is not breaking change to previous implement I thought now cause the behavior of `git authors AUTHORS` did not change. This condition is to forbid using `--no-email` with the default behavior when invoking `git authors` directly.

man/git-authors.md
vanpipy: I follow the document existed and I will try 😄

bin/git-authors
```diff
@@ -43,12 +23,47 @@ authors() {
   fi
 }
 
-#
-# authors.
-#
+if [[ -z "$file" ]]; then
+  file=$(find . -mindepth 1 -maxdepth 1 -iregex '.*\(authors\|contributors\).*' | head -n1)
```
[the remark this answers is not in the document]
vanpipy: Let me take a comparison, I did remember this.

bin/git-authors
```diff
@@ -43,12 +23,47 @@ authors() {
...
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    -l|--list )
+      list=1
+      ;;
+    --no-email )
+      no_email=1
+      ;;
+    --output )
+      if [[ -n $2 ]] && [[ $2 =~ ^[a-zA-Z] ]]; then
```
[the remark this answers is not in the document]
vanpipy: Thanks, I got this error and this is a missing I did not think about. For option `--output`, it looks supporting the paths makes sense.

man/git-authors.md
hyperupcall: I see, in that case, I would leave it for now. If it's in other places, I think a separate PR should address it. Thanks for letting me know :)

bin/git-authors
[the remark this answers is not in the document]
hyperupcall: If the file begins with a non-ASCII character like á, the file will not be written to. I suggest removing this sort of check altogether (just keep the -z/-n, how it was before)

bin/git-authors
hyperupcall: I'm not sure I understand. On main:
```console
$ ./bin/git-authors --no-email
...
```
This branch:
```console
./bin/git-authors --no-email
--no-email option only can be used with --list | -l | --output
```
That is a breaking change.

hyperupcall: Also, please do not mark these conversations as resolved, the CONTRIBUTING mentions that is the responsibility of the reviewer. It makes it more difficult to track the conversation

bin/git-authors
```diff
@@ -43,12 +23,47 @@ authors() {
...
+    file="AUTHORS"
+  fi
+fi
+
+if [[ $# = 0 ]]; then
+  authors >> "$file"
+  exit 0
+fi
+
+while [[ $# -gt 0 ]]; do
```
hyperupcall, line 38: Moving big sections around like the argument parsing makes it more difficult to read the diff. Can the argument parsing be moved back to where it was - at the top of the file?

bin/git-authors
```diff
@@ -43,12 +23,47 @@ authors() {
...
+  case $1 in
+    -l|--list )
+      list=1
+      ;;
+    --no-email )
+      no_email=1
+      ;;
+    --output )
+      if [[ -n $2 ]] && [[ $2 =~ ^(\.\/|\.\.\/|\/)?([a-zA-Z_](\/)?)+$ ]]; then
+        output=$2
```
hyperupcall, line 48: I notice the variables `file` and `output` behave and mean the same thing. Is it possible to be consistent and use the old, `file` variable instead of making a new one?

bin/git-authors
```diff
@@ -43,12 +23,47 @@ authors() {
...
-# authors.
-#
+if [[ -z "$file" ]]; then
+  file=$(find . -mindepth 1 -maxdepth 1 \( -iname '*authors*' -o -iname '*contributors*' \) | head -n1)
+  if [[ -z "$file" ]]; then
+    file="AUTHORS"
+  fi
+fi
+
+if [[ $# = 0 ]]; then
```
hyperupcall, line 33: There seems to be some duplication. There are two sections that have

```sh
authors >> "$somevariable"
exit 0
```

Can they be combined into one? If they share the same variable name, it should be relatively straightforward

hyperupcall (changes requested):

vanpipy:
It is not the right time to do the refactor, do this again when the coverage is ok.

PR #1124, opened 2023-12-13
fix(utimes): change /dev/stdin to -

rasa:
As not all environments have `/dev/stdin`. See https://github.com/ish-app/ish/issues/944

Sorry ‘bout that.

### DIFF
--- a/bin/git-utimes
+++ b/bin/git-utimes
@@ -128,4 +128,4 @@ git --no-pager status "${status_opts[@]}" . \
 # prefix is not stripped:
 git --no-pager whatchanged "${whatchanged_opts[@]}" . \
   | awk "${awk_flags[@]}" "${awk_script}" "${tmpfile}" - \
-  | BASH_ENV='' bash "${bash_opts[@]}" /dev/stdin
+  | BASH_ENV='' bash "${bash_opts[@]}" -


spacewander (approved):

spacewander:
> As not all environments have `/dev/stdin`. See [ish-app/ish#944](https://github.com/ish-app/ish/issues/944)
> 
> Sorry ‘bout that.

Thanks for your contribution.
No need to be sorry about this!

hyperupcall (approved):
Thanks for your contribution - I'll approve these changes, but I'm not certain they are necessary? According to the manpage:

> Bash handles several filenames specially when they are used in redirec‐tions, as described in the following table.  If the operating system on which bash is running provides these special files, bash will use them; otherwise  it  will emulate them internally with the behavior described below.

> /dev/stdin: File descriptor 0 is duplicated

Since our shebang points to Bash, I think this shouldn't be a problem. :)

rasa:
> Since our shebang points to Bash, I think this shouldn't be a problem. :)

@hyperupcall It’s not a bash thing, the device doesn’t exist in the OS:
```
iPad:~# echo whoami | bash -
root
iPad:~# echo whoami | bash /dev/stdin
bash: /dev/stdin: No such file or directory
iPad:~# uname -a
Linux iPad 4.20.69-ish SUPER AWESOME May 20 2023 23:41:32 i686 Linux
```

hyperupcall:
@rasa I don't quite understand - your example code doesn't run under Bash. I don't expect `echo whoami | bash /dev/stdin` to work, but I expect `bash -c "echo whoami | bash /dev/stdin"` to work. The shebang points to `bash`, not `sh`

rasa:
> your example code is not running under Bash.

It is, as `cat < <(ls -1 /dev)` is a bashism, that doesn't work in sh/ash/dash.

hyperupcall:
I see - perhaps I misread the Bash man page as those /dev/stdin files are not being used for redirection. The `bash: /dev/fd/63: No such file or directory` error does seems to suggest that this is being ran under Bash.

Thanks for the fix

Issue #1133, opened 2024-02-12
Add a flag to "git delete-squashed-branches" to not abort on first error


nkiesel:
I love this command, but I often have squashed branches which cannot be deleted because they are checked out locally in a different worktree.  And because the script immediately aborts in this situation, it will not try to delete other squashed branches.

I created a local copy which simply replaces `git branch -D "$branch"` with `git branch -D "$branch" || true`, but I could file a PR which instead uses a flag for this if you are interested. or is there a reason to not do that?

nkiesel:
something like the following (do not like the flag name):
```
--- /opt/homebrew/bin/git-delete-squashed-branches	2023-10-29 03:52:56
+++ /Users/norbert.kiesel/bin/git-delete-squashed-branches	2024-02-12 12:51:33
@@ -37,6 +37,17 @@

 set -euo pipefail

+try-next() {
+  false
+}
+
+if [[ $# -gt 0 && "$1" == "--all" ]]; then
+    try-next() {
+      true
+    }
+    shift
+fi
+
 if [[ $# -eq 0 ]]; then
   targetBranch=$(git rev-parse --abbrev-ref HEAD)
 else
@@ -48,6 +59,6 @@
     mergeBase=$(git merge-base "$targetBranch" "$branch")

     if [[ $(git cherry "$targetBranch" "$(git commit-tree "$(git rev-parse "$branch^{tree}")" -p "$mergeBase" -m _)") == "-"* ]]; then
-        git branch -D "$branch"
+        git branch -D "$branch" || try-next
     fi
 done
 ```

spacewander:
PR is welcome! What name do you want to use for this new flag?

spacewander:
Done

PR #1134, opened 2024-02-15
Add --proceed flag to git-delete-squashed-branches

nkiesel:

### DIFF
--- a/bin/git-delete-squashed-branches
+++ b/bin/git-delete-squashed-branches
@@ -2,6 +2,12 @@
 
 set -euo pipefail
 
+proceed=false
+if [[ $# -gt 0 && ("$1" == "--proceed" || "$1" == "-p") ]]; then
+    proceed=true
+    shift
+fi
+
 if [[ $# -eq 0 ]]; then
   targetBranch=$(git rev-parse --abbrev-ref HEAD)
 else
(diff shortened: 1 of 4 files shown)


spacewander (approved):

hyperupcall (changes requested):
The code change itself LGTM, but I'm wondering if this is the right way to go

To me, it seems the behavior with `--proceed` is what is expected when running the command. If there is 10 branches that should be deleted, I do not think it should stop on the first one, if the first one fails to delete, by default. Maybe the best approach is to just add `|| true` to the `git branch -D "$branch"`?

@spacewander what do you think?

hyperupcall:
Oops, I didn't realize this was already discussed in #1133. I've added the reference to PR description

nkiesel:
Yes, I would have preferred making this the default behavior. However, I was afraid that this breaks compatibility, so opted for a non-default option. Alternative could be to also support an env variable which changes the default behavior.

hyperupcall:
@nkiesel I agree, backwards compatability is important. In my opinion, it is OK because the fix restores the command behavior to what is expected when this command runs, especially as these commands are typically run interactively and the behavior and flags remain the same as similar commands like `git-delete-{branch,merged-branches}`.

I'm not going to block on this issue, the final verdict is up to @spacewander. We can always make a change later that will change the default behavior, which will both be backwards-compatible to, and obsolete, the `--proceed` flag.

hyperupcall (approved):

spacewander:
Let's merge this one first :smile:

norwd:
@tj / @spacewander, I think this PR is missing from `main` (it's in `master` though), should this be reopened and merged to `main`?

See: https://github.com/tj/git-extras/compare/main...master

spacewander:
@norwd 
Thanks for your nice catch. Let's open a new PR to the main branch.

PR #1136, opened 2024-02-23
refactor: test utils and simplify the case code

vanpipy:


tests/README.md
```diff
@@ -26,6 +26,18 @@ It is done or go without `poetry`,
...
+One command has a unit test, because one `git-*` command is just do one thing, so we can eat a piece of `git-*` command in one time.
+
+For example,
+
+1. The `git-alias` should has a test suite, so create `test_git_alias.py` in the directory `test`
+2. Create a test class `TestGitAlias` in the `test_git_alias.py`
+3. Create a test case `test_init`, and some test fixtures can be used, `temp_repo`, `named_temp_repo` etc.
+    * `temp_repo` is module scoped fixture which create a temporary directory and available in the test suite `test_git_alias.py`.
+    * `named_temp_repo` is just same as `temp_repo` except the custom directory renaming.
+4. Loop the third step until the 100% coverage of the function of the `git-alias`
```
spacewander, line 39:
```suggestion
4. Loop the third step until the 100% coverage of the function of the `git-alias`
```

Could we have a way to measure it?

tests/README.md
```diff
@@ -26,6 +26,18 @@ It is done or go without `poetry`,
 
 The second way maybe blocked the some missing dependencies at someday, so the first one is recommended.
 
+# What and how to create a unit test
+One command has a unit test, because one `git-*` command is just do one thing, so we can eat a piece of `git-*` command in one time.
+
+For example,
+
+1. The `git-alias` should has a test suite, so create `test_git_alias.py` in the directory `test`
```
spacewander, line 34:
```suggestion
1. The `git-alias` should have a test suite, so create `test_git_alias.py` in the directory `test`
```

tests/helper.py
```diff
@@ -98,3 +108,19 @@ def invoke_installed_extras_command(self, name, *params):
         script = [temp_extras_command, *params]
         print(f"Run the script \"{script}\"")
         return subprocess.run(script, capture_output=True)
+
+    def change_origin(self, origin_url):
+        try:
+            self._git_repo.git.remote("add", "origin", origin_url)
+        except GitCommandError:
+            pass
```
spacewander, line 116: We should log the error?

tests/README.md
spacewander: We should rename https://github.com/tj/git-extras/blob/main/tests/test_authors.py?

vanpipy: Yes, it should be, update later.

tests/helper.py
vanpipy: The log is better but the throw is not expected.

tests/README.md
vanpipy: I think the coverage of the unit test is a good way to measure it.

tests/helper.py
vanpipy: Update

tests/README.md
vanpipy: Update the typo and the pytest coverage will be update in another PR.

tests/README.md
vanpipy: The pytest coverage will be update in another PR.

spacewander (approved):

PR #1140, opened 2024-04-26
ci: update actions

boidolr:
Resolve deprecation warnings, see https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

For an example of the warnings see any recent ci run, for example https://github.com/tj/git-extras/actions/runs/8826310597


spacewander (approved):

spacewander:
@boidolr 
Would you also add the dependabot to update the action automatically? Thanks!

boidolr:
Sure, here you are: https://github.com/tj/git-extras/pull/1141

PR #1147, opened 2024-06-14
fix: git-summary commit count

xwjdsh:
If the user has set log.showSignature = ture in gitconfig file, then the `commits` field in the git-summary result is wrong. Replace `git log` with `git rev-list` to avoid the problem, `git rev-list` isn't affected by log.* config parameters, which is also the same as `git-count` command.

---
The result of `git log --oneline HEAD` is as follows, then `| wc -l | tr -d ' '` get wrong count.
```
e34dea9 (origin/main, origin/HEAD) gpg: Signature made Fri May  3 15:52:28 2024 JST
gpg:                using RSA key B5690EEEBB952194
gpg: Can't check signature: No public key
tests: update dependencies (#1142)
3bac05e gpg: Signature made Mon Apr 29 10:59:50 2024 JST
gpg:                using RSA key B5690EEEBB952194
gpg: Can't check signature: No public key
chore(deps): bump masesgroup/retrieve-changed-files from 2 to 3 (#1144)
...
```

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -77,7 +77,7 @@ active_days() {
 #
 commit_count() {
   # shellcheck disable=SC2086
-  git rev-list --count "$commit"
+  git rev-list $MERGES_ARG --count "$commit"
 }
 
 #
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -77,7 +77,7 @@ active_days() {
 #
 commit_count() {
   # shellcheck disable=SC2086
-  git log $MERGES_ARG --oneline "$commit" | wc -l | tr -d ' '
+  git rev-list --count "$commit"
 }
 
 #


bin/git-summary
```diff
@@ -77,7 +77,7 @@ active_days() {
 #
 commit_count() {
   # shellcheck disable=SC2086
-  git log $MERGES_ARG --oneline "$commit" | wc -l | tr -d ' '
+  git rev-list --count "$commit"
```
spacewander, line 80: We pass `--no-merges` via `$MERGES_ARG`. How could we handle this with `rev-list`?

xwjdsh: I missed it, updated.

spacewander (approved):

spacewander:
Merged. Thanks!

hyperupcall (commented):
LGTM

PR #1149, opened 2024-07-10
feat: add rename-file command

CodeByZach:
Rename a file or directory and ensure Git recognizes the change, regardless of filesystem case-sensitivity. It combines the functionality of the `mv` command and `git mv`. This is particularly useful for renaming files or directories to change only their case, which might not be detected by Git on case-insensitive filesystems.

I initially considered naming it `git-mv` due to its intuitive functionality. However, I noticed that the naming convention in the repository includes both commands with and without the dash following "git-". To avoid conflict with the native `git mv` command, I opted for `git-rename-file` instead.

- Added new command 'rename-file' to rename files and directories within a Git repository regardless of filesystem case-sensitivity.
- Updated AUTHORS and documentation in Commands.md.
- Implemented bash, zsh, and fish completions for 'rename-file'.
- Created manual pages for 'rename-file' in various formats (man, html, md).
- Updated index.txt to include 'rename-file' command.


bin/git-rename-file
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
+
```
spacewander, line 2: Let's add https://github.com/tj/git-extras/blob/e1a4e7195d54cf3e0a931bb53a2584697ba431f0/bin/git-rename-remote#L3 for new command

bin/git-rename-file
```diff
@@ -0,0 +1,99 @@
...
+check_case_sensitive_exists() {
+    local path="$1"
+    local dir
+    local base
+
+    dir=$(dirname "$path")
+    base=$(basename "$path")
+
+    if [ -e "$dir" ]; then
+        if (cd "$dir" && find . -maxdepth 1 -name "$base" | grep -q "$base"); then
```
spacewander, line 51: Could we use `find "$dir"` to save a child shell?

bin/git-rename-file
CodeByZach: I'm good with this.

bin/git-rename-file
CodeByZach: Can you show me what you mean? Thanks!

spacewander: We can rewrite the cmd into `find "$dir" -maxdepth 1 -name "$base" ...` so we can avoid using a sub shell wrapped with `()`

hyperupcall: Is `| grep -q $base` any different from redirecting like `>/dev/null 2>&1`? I believe `find` should already filter out the path;

bin/git-rename-file
```diff
@@ -0,0 +1,99 @@
...
+    if [ -e "$TEMP_NAME" ]; then
+        git mv -f "$TEMP_NAME" "$SOURCE"
+    fi
+}
+
+# Trap errors to trigger rollback
+trap 'rollback' ERR
+
+# Move the file to a temporary name within the Git repository
+git mv "$SOURCE" "$TEMP_NAME" || exit 1
```
hyperupcall, line 94: Once `set -e` is added to the top, these two `exit 1`'s can be removed;

bin/git-rename-file
```diff
@@ -0,0 +1,99 @@
...
+if [[ "$1" == "-h" || "$1" == "--help" ]]; then
+    usage
+    exit 0
+fi
+
+# Check for correct number of arguments
+if [ "$#" -ne 2 ]; then
+    echo "Error: Incorrect number of arguments."
+    echo ""
+    usage
```
hyperupcall, line 32:
```suggestion
    usage >&2
```

bin/git-rename-file
```diff
@@ -0,0 +1,99 @@
...
+# Check if destination already exists
+if check_case_sensitive_exists "$DESTINATION"; then
+    echo "Error: Destination '$DESTINATION' already exists."
+    exit 1
+fi
+
+# Check if the destination directory exists
+DEST_DIR=$(dirname "$DESTINATION")
+if ! check_case_sensitive_exists "$DEST_DIR"; then
+    echo "Error: Destination directory '$DEST_DIR' does not exist."
```
hyperupcall, line 78: Small nit, but it would be good to redirect all errors to standard error. I believe that's consistent with the rest of the code;

hyperupcall (changes requested):

hyperupcall:
Thanks, this looks good - I had a few suggestions;

bin/git-rename-file
CodeByZach: > Is `| grep -q $base` any different from redirecting like `>/dev/null 2>&1`? I believe `find` should already filter out the path;

The method I used was tested in multiple different scenarios. Please ensure whatever method you suggest is still able to efficiently and reliably detect case-sensitivity regardless of the case-sensitivity of the filesystem.

CodeByZach:
Apologies, just catching up here. What's left before this can be merged?

bin/git-rename-file
hyperupcall: Yeah I don't know, I think I see now what `-iname` does on case insensitive file systems, which explains the `grep`.

CodeByZach: > Yeah I don't know, I think I see now what `-iname` does on case insensitive file systems, which explains the `grep`.

So should I leave as-is?

hyperupcall:
@CodeByZach A must is to have the lint check in CI pass (if it's a lot of issues, running `shfmt` locally should speed things up). While you do that it would be a good idea to address the code change recommendations

CodeByZach:
> @CodeByZach Hey, a must is to have the lint check in CI pass (if it's a lot of issues, running `shfmt` locally should speed things up). While you do that it would be a good idea to address the code change recommendations

Seems easy enough.
image

bin/git-rename-file
hyperupcall: Oops, I missed this notification. Yeah, leave it as it is. My recommendation came from a misunderstanding

hyperupcall:
@CodeByZach Thank you for the update. I'll be able to merge this once @spacewander's comments [here](https://github.com/tj/git-extras/pull/1149#discussion_r1677280688) and [here](https://github.com/tj/git-extras/pull/1149#discussion_r1678682251) are addressed

CodeByZach:
> @CodeByZach Thank you for the update; I'll be able to merge this once @spacewander's comments [here](https://github.com/tj/git-extras/pull/1149#discussion_r1677280688) and [here](https://github.com/tj/git-extras/pull/1149#discussion_r1678682251) are addressed;

Should now be addressed.

spacewander (approved):

PR #1152, opened 2024-08-08
fix: Use absolute (relative to repo root) path to supplied filename

mattcen:
If a file is referenced from the subdir of a repository, the script confirms it exists and tacks the filename onto the URL, but doesn't check if the filename provided is relative to the repo root.

I couldn't immediately work out the best way to update the tests to:

1. Create a temp dir *inside* the existing temp repo
2. Create a new temp file within that new dir
3. Change CWD to the new dir
4. Run git-browse on that new file with a relative path


spacewander:
@mattcen 
Is this PR ready for review?

mattcen:
@spacewander  happy for review, with the caveat I never worked out how to update tests as per PR description

bin/git-browse
```diff
@@ -33,7 +33,7 @@ if [[ $remote_url =~ gitlab ]]; then
     # construct gitlab urls
     # https://gitlab.com/<user_or_group>/<repo>/-/blob/<commit_or_branch>/<filename>#L<line1>-<line2>
     if [[ -n ${filename} ]]; then
-        url="${url}/-/blob/${commit_or_branch}/${filename}"
+        url="${url}/-/blob/${commit_or_branch}/$(git ls-files --full-name "${filename}")"
```
spacewander: Would you change `$(git ls-files --full-name "${filename}")` into a var so we don't need to repeat it?

spacewander:
Looks like we need to update the tests.

bin/git-browse
mattcen: Done.

spacewander:
Would you update other test files too?

mattcen:
I could not immediately work out how they worked or how to update them in the time I had available, which is why this wasn't completed when I first opened the PR. I will try to take another look at them, but am not sure how soon that will be; any guidance or suggestions  you can provide would be much appreciated.

spacewander:
Maybe you can change https://github.com/tj/git-extras/blob/99207eee8ef73fb25099f8559d6477dbc4e14ff8/tests/helper.py#L47 to return a full path name like os.path.relpath(file, self._cwd)?

PR #1155, opened 2024-08-10
fix: Exit with error if HEAD commit doesn't yet exist on remote

### LINKED ISSUE #1153: git-browse doesn't check if a commit exists on upstream
When using git-browse to generate a web link to a repo, it constructs a URL based on the commit at the HEAD of the local repo, but doesn't actually check if this HEAD exists on the remote. This means that if I have a local commit that I've yet to push, and I run git-browse, it will fail because the remote end doesn't know about the commit the URL references.

Ideally, we should test whether the commit exists upstream first, using:

```
git branch --remotes --contains "${commit_or_branch}" 2>/dev/null | grep -q "\<${remote}\>/"
```

Which will exit non-zero if there's no remote ref for this commit. We could then display an error to the user.

mattcen:
Fixes #1153

### DIFF
--- a/bin/git-browse
+++ b/bin/git-browse
@@ -29,6 +29,14 @@ fi
 commit_hash=$(git rev-parse HEAD 2>/dev/null)
 commit_or_branch=${commit_hash:-${branch}}
 
+# Check that the commit branch actually *exists* on the remote first:
+if ! git branch --remotes --contains "${commit_or_branch}" 2>/dev/null |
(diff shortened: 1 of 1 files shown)


hyperupcall:
LGTM! CI failure looks unrelated.

spacewander:
The unit tests failed because we added an upstream check and they can't satisfy. Could we mock the upstream check in the test?

PR #1156, opened 2024-08-20
Add pathspec support in `git-missing`

Weyzu:
This PR add pathspec support for the `git-missing` command. Users can now limit the commit difference to specific directories or files. Addition is particularly useful for big repositories where users are not concerned with other directories outside their usual workspace.

Pathspec is optional, command will work as usual for users not using it.

## New Usage

```bash
git missing [<first branch>] <second branch> [<git log options>] [[--] <path>...]
```

## Example

Compare differences between the current branch and master, limited to the `src/` directory:

```bash
$ git missing master -- src/
< a1b2c3d feat: Add new component in src/
> d4e5f6g fix: Update existing module in src/
```

### DIFF
--- a/bin/git-missing
+++ b/bin/git-missing
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 usage() {
-    echo 1>&2 "usage: git missing [<first branch>] <second branch> [<git log options>]"
+    echo 1>&2 "usage: git missing [<first branch>] <second branch> [<git log options>] [[--] <path>...]"
 }
 
 if [ "${#}" -lt 1 ]
@@ -12,9 +12,20 @@ fi
 
 declare -a git_log_args=()
 declare -a branches=()
+declare pathspec=""
+declare parse_path=false
+
 for arg in "$@" ; do
 
+    if [[ $parse_path == true ]]; then
+        pathspec="$*"
+        break
+    fi
+
     case "$arg" in
+        --)
+            parse_path=true
+            ;;
         --*)
             git_log_args+=( "$arg" )
             ;;
@@ -22,6 +33,7 @@ for arg in "$@" ; do
             branches+=( "$arg" )
             ;;
     esac
+    shift
 done
 
 firstbranch=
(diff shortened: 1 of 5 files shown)


bin/git-missing
```diff
@@ -12,16 +12,28 @@ fi
 
 declare -a git_log_args=()
 declare -a branches=()
+declare pathspec=""
+declare parse_path=false
+
 for arg in "$@" ; do
 
+    if [[ $parse_path == true ]]; then
+        pathspec="$*"
```
hyperupcall, line 21: From this code, it looks like only a single `pathspec` is supported. If multiple are passed, it is treated as a single path. To fix this, `pathspec` should be an indexed array; then, you can do `pathspec+=("$@")` here.

bin/git-missing
```diff
@@ -38,4 +50,8 @@ else
     exit 1
 fi
 
-git log "${git_log_args[@]}" "$firstbranch"..."$secondbranch" --format="%m %h %s" --left-right
```
hyperupcall, line 41: I think a single codepath can be preserved if `pathspec` is an indexed array and you add `-- "${pathspec[@]}"` at the end?

bin/git-missing
```diff
@@ -12,16 +12,28 @@ fi
...
+            parse_path=true
+            ;;
         --*)
             git_log_args+=( "$arg" )
             ;;
         *)
             branches+=( "$arg" )
             ;;
     esac
+    shift
```
hyperupcall, line 36: Was this unintentional? I don't think this changes the behavior of the program at all and it could be removed

hyperupcall (changes requested):
Thanks for this fix! I did see a bug - that passing multiple files to be used as a `pathspec` is not supported. My comments have more details

bin/git-missing
Weyzu: Excellent point! I was perhaps to focused on my usual use case of using just one path. I introduced your proposal in the just added commit. (b8488cd)

bin/git-missing
Weyzu: Spot on, you're right. Simplified in b8488cd.

bin/git-missing
Weyzu: Yes, it's a leftover from my tinkering, sorry about that. Removed. (b8488cd)

Weyzu:
Sorry for slight delay in response. Thank you for the ultra fast review and quality feedback. 👍🏻

hyperupcall:
@Weyzu Thanks! Everything looks great to me! @spacewander what do you think?

spacewander (commented):
Would you update the md https://github.com/tj/git-extras/blob/main/man/git-missing.md and generate new doc from it? 
Thanks!

Weyzu:
@spacewander Sure thing, updated. :) (8716c18) Let me know if I phrased it all correctly.

hyperupcall (approved):

spacewander (approved):

spacewander:
Merged. Thanks!

PR #1160, opened 2024-09-16
Enhance `git-repl`

### LINKED ISSUE #1159: it would be nice if the git repl printed out the exit code of the previous command
In my own bashrc, I have it so that every prompt line prints out `x⏎` at the front, where x is the exit code of the previous command. This makes observing the behaviors of commands that return an exit code easier.

wyattscarpenter:
This makes 3 improvements to git-repl:

- It displays  the exit code of the previous command

- The message at the beginning now tells you how to quit. q ~and x~ have been added as commands that quit

- If git-repl is given arguments, it will be evaluate them as the first command.

Closes #1159

### DIFF
--- a/bin/git-repl
+++ b/bin/git-repl
@@ -10,9 +10,9 @@ while true; do
 
   # Prompt
   if test -n "$cur"; then
-    prompt="git ($cur)> "
+    prompt="$exit_status|git ($cur)> "
   else
-    prompt="git> "
+    prompt="$exit_status|git> "
   fi
 
   # Readline
@@ -32,11 +32,14 @@ while true; do
   esac
 
   if [[ $cmd == !*  ]]; then
-    eval ${cmd:1} 
+    eval ${cmd:1}
+    exit_status=$?
   elif [[ $cmd == git* ]]; then
     eval $cmd
+    exit_status=$?
   else
     eval git "$cmd"
+    exit_status=$?
   fi
 done
 
--- a/bin/git-repl
+++ b/bin/git-repl
@@ -45,14 +45,12 @@ while true; do
 
   if [[ $cmd == !*  ]]; then
     eval ${cmd:1}
-    exit_status=$?
   elif [[ $cmd == git* ]]; then
     eval $cmd
-    exit_status=$?
   else
     eval git "$cmd"
-    exit_status=$?
   fi
+  exit_status=$?
 done
 
 echo
--- a/bin/git-repl
+++ b/bin/git-repl
@@ -17,7 +17,7 @@ while true; do
 
   # Use arguments as a command if any are provided
   if [ $# -ne 0 ]; then
-    cmd="$@"
+    cmd=$*
     set --
   else
     # Readline
--- a/bin/git-repl
+++ b/bin/git-repl
@@ -25,7 +25,7 @@ while true; do
   if [ $# -ne 0 ]; then
     cmd=$*
     set --
-    echo "$prompt" "$cmd" # iI is as though you had entered a command.
+    echo "$prompt" "$cmd" # It is as though you had entered a command.
   else
     # Readline
     read -e -r -p "$prompt" cmd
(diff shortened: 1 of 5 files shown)


bin/git-repl
```diff
@@ -2,24 +2,29 @@
...
   fi
 
-  # Readline
-  read -e -r -p "$prompt" cmd
-
-  # EOF
-  test $? -ne 0 && break
+  # Use arguments as a command if any are provided
+  if [ $# -ne 0 ]; then
+    cmd="$@"
```
hyperupcall, line 20: nit: Change this to `$*`

hyperupcall (changes requested):
Everything mostly LGTM - I just had a single nitpick. I'm also thinking that adding `x` as a way to exit the repl might be an option to much/excessive. Not sure what @spacewander thinks about that

wyattscarpenter:
Oddly enough: I only added x because I wanted to add q and assumed that people who prefer the term exit over the term quit would object to my blatant favoritism if I didn't add x as well! I'm quite happy to remove it, myself

bin/git-repl
```diff
@@ -2,24 +2,29 @@
...
+echo "type 'ls' to ls files below current directory, '!command' to execute any command or just 'subcommand' to execute any git subcommand. 'quit', 'exit', 'q', 'x', ^D, or ^C to exit the git repl."
 
 while true; do
   # Current branch
   cur=$(git symbolic-ref HEAD 2> /dev/null | cut -d/ -f3-)
 
   # Prompt
   if test -n "$cur"; then
-    prompt="git ($cur)> "
+    prompt="$exit_status|git ($cur)> "
```
spacewander, line 13: `0|git ` is not so obvious. Maybe we can omit the `0` and only show non-zero exit code like this `exit code: 1 | git ...`? Any idea to make the exit code be more readable is welcome.

wyattscarpenter: Yeah, there are a lot of different ways it could be displayed. I have my bashrc set up to display the return code in red followed by a ⏎ (a unicode character which none of my terminals ever display correctly).
image
Fish shell seems to display the code in red in the prompt in brackets, if the return code isn't 0.
image
A friend of mine, if memory serves, has his shell configured to print a line after every process that exits non-zero with something like `12 ← the command exited with this status` highlighted in red. (I don't have a screenshot of this.)

spacewander: @wyattscarpenter 
Maybe we can try the fish shell style?

wyattscarpenter: @spacewander happy to! Do you want it colored? (Everything in the current git repl is monochrome, but theoretically there shouldn't be any problem adding the terminal escape code for the color, which should be supported everywhere.)

spacewander: @wyattscarpenter 
Adding color is great! Let's do it.

man/git-repl.md
```diff
@@ -5,14 +5,20 @@ git-repl(1) -- git read-eval-print-loop
...
 
   Commands can be prefixed with an exclamation mark (!) to be interpreted as
-  a regular command.
+  a regular shell command.
+
+  Any arguments to git repl will be taken collectively as the first command
+  to execute in the repl.
 
-  Type `exit` or `quit` to end the repl session.
+  Type `exit`, `quit`, or `q` to end the repl session. Ctrl-D an Ctrl-C
```
spacewander, line 20:
```suggestion
  Type `exit`, `quit`, or `q` to end the repl session. Ctrl-D and Ctrl-C
```

spacewander (commented):
Please update the doc after fixing the typo

man/git-repl.md
wyattscarpenter: Thanks!

spacewander (approved):

hyperupcall (approved):
Thanks! The new changes are much improved

PR #1161, opened 2024-09-17
Update git-alias.md: add brs to prevent incorrect line behavior

wyattscarpenter:
Currently these all appear on one line, in the rendered markdown viewer and also in the man page. That isn't what's wanted. What's wanted is for each to be on its own line, which this change implements, using the strategy I found in git-bulk.

Running `make` to generate the new documentation files apparently fixed a typo in another file, as well.


spacewander (approved):

spacewander:
Thanks for your contribution!

PR #1162, opened 2024-09-17
Use filetimes in check_integrity

wyattscarpenter:
Before this change, if you changed the documentation in foo.md, check_integrity.sh would not warn you that you had to update foo.html and foo.1. This is bad, because then the derived files could get out of sync. (It also was not checked by the CI, which also just uses check_integrity.sh)

This PR makes 4 changes:

1. check_integrity.sh now tells if you've updated the md since you updated the html and man page, using the file modified times. I got this idea from the tool we use to generate the files, which also checks that, which is why you have to touch the md if you want it to generate fresh derived files.
2. To facilitate the first point, check_integrity.sh now calls our very own `git-utimes --newer` in order to give the files the correct file modified times. This always runs, which has the downside of taking a couple more seconds. It has the upside of being simple, and always correct. (Well, I guess it could technically become incorrect if you deliberately touched the md to before the previous git commit date for it. But you can only cheat yourself that way.)
3. If check_integrity.sh is given no arguments, it checks the integrity of all the commands. This uses the `$(find bin | cut -b 5- | xargs)` method of getting all the command names, that was previously in the CI.
4. Update git-extras, which was out of sync.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
           sudo apt-get install -y bsdmainutils
       - name: Script
         run: |
-          ./check_integrity.sh $(find bin | cut -b 5- | xargs)
+          ./check_integrity.sh
       - name: Brew release
         if: matrix.platform == 'macos-latest'
         run: |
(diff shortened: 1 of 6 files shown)


spacewander (approved):

spacewander:
Merged. Thanks!

PR #1164, opened 2024-09-21
Update some documentation that was out of sync

wyattscarpenter:
As I was updating some other code, the check_integrity script informed me that these three files were out of sync... and indeed they were! In the first one, there's a whole new usage pattern documented, and in the others an http has been changed to an https.

It's nice that the script worked now, but why hasn't it worked to prevent this in the CI, such as the recent runs of https://github.com/tj/git-extras/pull/1160 (after the new CI should be in place)? For that matter, why didn't it show up to me locally earlier? Did something else change? Perhaps the --newer flag should be removed from our invocation of git-utimes in check_integrity?

(Perhaps this PR shouldn't be accepted until there is at least one other commit past the CI-altering commit https://github.com/tj/git-extras/commit/4d2d393142918594da02e498169dd692fe21e17c in the git history (or at least has been attempted to be put in the git history), because maybe that would trigger the new CI and quell my doubts.)

Since I had to run the instructions to update each file, I noticed they were hard to copy-paste, and changed them to be easier.

### DIFF
--- a/check_integrity.sh
+++ b/check_integrity.sh
@@ -6,7 +6,7 @@ err() {
 }
 
 make_doc() {
-    echo "'touch man/git-$1.md && make man/git-$1.{1,html}'"
+    echo "touch man/git-$1.md && make man/git-$1.{1,html}"
 }
 
 check_bash_script() {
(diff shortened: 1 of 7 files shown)


spacewander (approved):

spacewander:
Merged. Thanks!

PR #1169, opened 2024-09-30
Change git-cp to use cleaner branch approach

### LINKED ISSUE #1090: Allow override temp file location in `git cp`
When running `git cp` it will create a temp file following a certain convention. 

E.g. running `git cp a/b/c.kt d/e.kt` will create a temp file like this: `a__b__c.kt-move-to-d__e.kt`

While this behavior works most of the time, it may run into path-too-long error when the src/dest file path is very deep. 

Suggestion: allow a third parameter to specify a temp file location override.

MattiasEh:
Change git-cp to use cleaner branch approach, per https://stackoverflow.com/a/46484848/32841 and https://devblogs.microsoft.com/oldnewthing/20190919-00/?p=102904

This approach avoids the need for messy temporary files and therefore eliminates the class of problems related to those--which means that this [resolves #1090]( https://github.com/tj/git-extras/issues/1090).  It also makes for a cleaner history (at least IMHO) and should affect the original branch atomically.

My changes seem to work well for me, but I don't have a great way to test regressions for this fairly significant approach-change.

### DIFF
--- a/bin/git-cp
+++ b/bin/git-cp
@@ -50,37 +50,29 @@ else
       fi
     fi
 
-    MERGE_OPT=
-    ff=$(git config --get merge.ff || true)
-    if [[ "$ff" == "only" ]]; then
-        MERGE_OPT="--ff"
-    fi
 
     echo "Copying $CURRENT_FILENAME into $DESTINATION_FILENAME"
 
-    INTERMEDIATE_FILENAME="${CURRENT_FILENAME//\//__}-move-to-${DESTINATION_FILENAME//\//__}"
+    # Pre-check that the source and destination will work
+    git mv --dry-run "${CURRENT_FILENAME}" "${DESTINATION_FILENAME}"
 
-    # We keep the existing file on the side in a commit
-    git mv "${CURRENT_FILENAME}" "${INTERMEDIATE_FILENAME}"
-    git commit -nm "Keep $CURRENT_FILENAME"
+    # Make a new branch and switch to it
+    BRANCH_NAME="git-split-$(date +%s)"
+    git checkout -b "$BRANCH_NAME"
 
-    # We come back to the previous state and revert that change
-    INTERMEDIATE_SAVED=$(git rev-parse HEAD)
-    git reset --hard HEAD^
-
-    # We move the file to its new destination
+    # Move the original file to the new destination, in this branch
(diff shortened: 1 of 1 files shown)


spacewander:
@MattiasEh 
Would you merge the main branch and check the CI is passed?

Maybe we can change our CI code according to the https://github.com/jitterbit/get-changed-files/issues/19#issuecomment-854657729. (It will go as another pull request)

bin/git-cp
```diff
@@ -50,37 +50,29 @@ else
...
+    # Restore the original file to keep it in the history
+    git checkout HEAD~ "${CURRENT_FILENAME}"
+    git commit -nm "Restore $CURRENT_FILENAME"
 
-    # We keep both files
-    # shellcheck disable=SC2086
-    git merge $MERGE_OPT "${DESTINATION_SAVED}" "${INTERMEDIATE_SAVED}" -m "Duplicate ${CURRENT_FILENAME} history."
+    # Switch to the original branch and merge this back in.
+    git checkout -
+    git merge --no-ff "$BRANCH_NAME" -m "Copying $CURRENT_FILENAME into $DESTINATION_FILENAME"
```
spacewander, line 73: This produces a commit with the repeated message (there is already a message `Copying ....` before it).
Could we optimize it?

MattiasEh: I'd be happy to change the commit messages to be more descriptive.  The reason I made them like this is because this merge message is the key one directly on the branch, so I figured it should be a summary message for the whole operation.

Here's an example of what results when I used this version of git-cp, run thrice, to turn one file into four copies--setting up for a commit that prunes down all four into just the parts each should keep:
 image

With this context, what do you think are the best messages?

MattiasEh: Do you prefer these messages?
image

spacewander: These messages look good to me.

MattiasEh: Great!  I have pushed the better commit messages and taken this PR out of Draft.  Thanks, @spacewander .

spacewander:
@MattiasEh 
Merged. Thanks!

MattiasEh:
> @MattiasEh Merged. Thanks!

Happy to help!  And thank you, @spacewander !

Issue #839, opened 2020-05-21
git-pull-request (and others) support for acccess tokens


tstackhouse:
For repos that use SAML SSO, email, password, and 2fa otp aren't sufficient., and GitHub kicks back that you must configure a Personal Access Token.  It doesn't look like the git-pull-request script is able to support that.  Perhaps it could read from an env var `GITHUB_TOKEN` or something similar? Then we could keep those secure in e.g. a `.env` file.

spacewander:
We can use a git configure like `git-extras.github.token` to define whether the token is enabled and the name of the environment variable which stores the token. Pull request is welcome!

TrTmPB:
But why

vanpipy:
Hi, @spacewander ,

I did not get the idea of the `git-extras.github.token` cause the original implementation of the `git-pull-request` base on the `git-extras.github-personal-access-token` via the keyword `Authorization` on the HTTP.

1. Email and password
2. 2FA
3. ...

Each line may a way to create a PR in the command line. It looks impossiable to use only one `token` to reach the multiple goals?

PR #1177, opened 2024-11-13
Support `GITHUB_TOKEN` var for `git-fork` and `git-pull-request`

nim65s:
Hi,

`git-fork` & `git-pull-request` currently require to set github token in git config global file.
I would like to keep this file versionned and public on github, so I would prefer not having to put secrets inside.

A common solution to this issue is to pass that token through the `GITHUB_TOKEN` environment variable instead.

Here is a simple commit implementing and documenting that.

ref. #839

### DIFF
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -17,11 +17,7 @@ read -r user
 # personal access token
 # config name is github-personal-access-token '_' is not allowed in git config
 
-if test -z "${GITHUB_TOKEN}"; then
-  github_personal_access_token=$(git config git-extras.github-personal-access-token)
-else
-  github_personal_access_token=$GITHUB_TOKEN
-fi
+github_personal_access_token=$(git config --default "$GITHUB_TOKEN" git-extras.github-personal-access-token)
 
 test -z "$github_personal_access_token" && abort "GITHUB_TOKEN, or git config git-extras.github-personal-access-token required"
 
--- a/bin/git-fork
+++ b/bin/git-fork
@@ -17,9 +17,13 @@ read -r user
 # personal access token
 # config name is github-personal-access-token '_' is not allowed in git config
 
-github_personal_access_token=$(git config git-extras.github-personal-access-token)
+if test -z "${GITHUB_TOKEN}"; then
+  github_personal_access_token=$(git config git-extras.github-personal-access-token)
+else
+  github_personal_access_token=$GITHUB_TOKEN
+fi
 
-test -z "$github_personal_access_token" && abort "git config git-extras.github-personal-access-token required"
+test -z "$github_personal_access_token" && abort "GITHUB_TOKEN, or git config git-extras.github-personal-access-token required"
 
 # extract owner + project from repo url
 project=${url##*/}
--- a/man/git-fork.md
+++ b/man/git-fork.md
@@ -27,7 +27,7 @@ git-fork(1) -- Fork a repo on github
   
   Make sure the personal access token has the right `OAuth` scopes for the repo(s)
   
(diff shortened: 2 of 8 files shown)


bin/git-fork
```diff
@@ -17,9 +17,13 @@ read -r user
 # personal access token
 # config name is github-personal-access-token '_' is not allowed in git config
 
-github_personal_access_token=$(git config git-extras.github-personal-access-token)
+if test -z "${GITHUB_TOKEN}"; then
+  github_personal_access_token=$(git config git-extras.github-personal-access-token)
+else
+  github_personal_access_token=$GITHUB_TOKEN
```
spacewander, line 23: Need to quote $GITHUB_TOKEN to make shellcheck happy

bin/git-fork
```diff
@@ -17,9 +17,13 @@ read -r user
 # personal access token
 # config name is github-personal-access-token '_' is not allowed in git config
 
-github_personal_access_token=$(git config git-extras.github-personal-access-token)
+if test -z "${GITHUB_TOKEN}"; then
```
spacewander, line 20: Better to use `test -n`, and put the highest priority branch (using GITHUB_TOKEN) as the first one. This way is more readable.

man/git-fork.md
```diff
@@ -27,7 +27,7 @@ git-fork(1) -- Fork a repo on github
   
   Make sure the personal access token has the right `OAuth` scopes for the repo(s)
   
-  Use `git config --global --add git-extras.github-personal-access-token <your-personal-access-token>`
+  Use `GITHUB_TOKEN` environment variable, or `git config --global --add git-extras.github-personal-access-token <your-personal-access-token>`
```
spacewander, line 31: Why git-fork.1 is not updated?

nim65s:
Thanks for the review !
I've added the requested changes

bin/git-fork
```diff
@@ -17,9 +17,13 @@ read -r user
 # personal access token
 # config name is github-personal-access-token '_' is not allowed in git config
 
-github_personal_access_token=$(git config git-extras.github-personal-access-token)
+if test -n "${GITHUB_TOKEN}"; then
+  github_personal_access_token="$GITHUB_TOKEN"
+else
+  github_personal_access_token=$(git config git-extras.github-personal-access-token)
```
hyperupcall, line 23:
```suggestion
  github_personal_access_token=$(git config --default '' git-extras.github-personal-access-token)
```

Would recommend passing `--default` to make code more robust. (in the future, `errexit` may be enabled, and this invocation shouldn't exit the entire script if the config key could not be found).

hyperupcall (changes requested):
I support this `GITHUB_TOKEN` feature! I just had a few notes

nim65s:
@hyperupcall what would you think about dropping that switch, and using only one
```
github_personal_access_token=$(git config --default "${GITHUB_TOKEN}" git-extras.github-personal-access-token)
```
?

hyperupcall:
@nim65s That's perfect!

hyperupcall (approved):

man/git-pull-request.md
```diff
@@ -13,7 +13,7 @@ A personal access token is required for making the API call to open the pull req
 
 Make sure the personal access token has the right `OAuth` scopes for the repo(s)
 
-Use `git config --global --add git-extras.github-personal-access-token <your-personal-access-token>`
```
spacewander, line 16: Could you also regenerate the git-pull-request.1 file?

hyperupcall:
Great, thanks!

PR #1179, opened 2024-11-22
feat: add shellcheck ci check

oikarinen:
This was mentioned in some older PR. Autofixes in the same commit, some which seem actual bugs.

### DIFF
--- a/bin/git-guilt
+++ b/bin/git-guilt
@@ -31,7 +31,7 @@ do
     esac
 done
 
-cd "$(git-root)" # cd for git blame
+cd "$(git-root)" || exit # cd for git blame
 MERGED_LOG=$(git_extra_mktemp)
 if [[ $EMAIL == '-e' ]]
 then
@@ -44,10 +44,10 @@ for file in $(git diff --name-only "$@")
 do
     test -n "$DEBUG" && echo "git blame $file"
     # $1 - since $2 - until
-    git blame $NOT_WHITESPACE --line-porcelain "$1" -- "$file" 2> /dev/null |
+    git blame "$NOT_WHITESPACE" --line-porcelain "$1" -- "$file" 2> /dev/null |
         LC_ALL=C sed -n "$PATTERN" | sort | uniq -c | LC_ALL=C sed 's/^\(.\)/- \1/' >> "$MERGED_LOG"
     # if $2 not given, use current commit as "until"
-    git blame $NOT_WHITESPACE --line-porcelain "${2-@}" -- "$file" 2> /dev/null |
+    git blame "$NOT_WHITESPACE" --line-porcelain "${2-@}" -- "$file" 2> /dev/null |
         LC_ALL=C sed -n "$PATTERN" | sort | uniq -c | LC_ALL=C sed 's/^\(.\)/+ \1/' >> "$MERGED_LOG"
 done
 
@@ -71,7 +71,7 @@ END {
             printf("%d %s\n", contributors[people], people)
         }
     }
-}' $MERGED_LOG | sort -nr | # only gawk supports built-in sort function
+}' "$MERGED_LOG" | sort -nr | # only gawk supports built-in sort function
 while read -r line
 do
     people=${line#* }
@@ -103,7 +103,7 @@ do
             do
                 printf "-"
             done
-            printf "(%s)" $num
+            printf "(%s)" "$num"
         else
             for (( i = 0; i > num; i-- ))
             do
--- a/bin/git-guilt
+++ b/bin/git-guilt
@@ -44,10 +44,12 @@ for file in $(git diff --name-only "$@")
 do
     test -n "$DEBUG" && echo "git blame $file"
     # $1 - since $2 - until
-    git blame "$NOT_WHITESPACE" --line-porcelain "$1" -- "$file" 2> /dev/null |
+    # shellcheck disable=SC2086
+    git blame $NOT_WHITESPACE --line-porcelain "$1" -- "$file" 2> /dev/null |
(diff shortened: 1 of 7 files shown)


bin/git-obliterate
```diff
@@ -19,5 +19,5 @@ then
 else
     # don't quote $range so that we can forward multiple rev-list arguments
     git filter-branch -f --index-filter "git rm -r --cached ""$file"" --ignore-unmatch" \
-        --prune-empty --tag-name-filter cat -- $range
+        --prune-empty --tag-name-filter cat -- "$range"
```
spacewander, line 22: `range` may contain multiple options, let's add a `shellcheck disable`.

oikarinen: I'm not an expert here but converted range to array so that multiple values would be quoted correctly in (theoretical) case they'd contain IFS or glob characters. LMKWYT? Could also just `shellcheck disable=SC2086`. Thanks for quick feedback on these few items 😎

spacewander:
Let's merge main branch to make CI pass.

bin/git-obliterate
```diff
@@ -9,15 +9,15 @@ do
...
+if [ -z "${range[*]}" ]
 then
     git filter-branch -f --index-filter "git rm -r --cached ""$file"" --ignore-unmatch" \
         --prune-empty --tag-name-filter cat -- --all
 else
-    # don't quote $range so that we can forward multiple rev-list arguments
+    # $range is an array so that we can forward multiple rev-list arguments
     git filter-branch -f --index-filter "git rm -r --cached ""$file"" --ignore-unmatch" \
-        --prune-empty --tag-name-filter cat -- $range
+        --prune-empty --tag-name-filter cat -- "${range[@]}"
```
spacewander, line 22: Should be "${range[*]}"?

oikarinen: This is what [SC2086](https://www.shellcheck.net/wiki/SC2086) doc indicated and looks correct based on the [bash](https://www.gnu.org/software/bash/manual/html_node/Arrays.html) docs. 
> If the subscript is ‘@’ or ‘*’, the word expands to all members of the array name. These subscripts differ only when the word appears within double quotes. If the word is double-quoted, ${name[*]} expands to a single word with the value of each array member separated by the first character of the IFS variable, and ${name[@]} expands each element of name to a separate word.

And indeed that seems to be the case:
```
$ ranges=( "range 1" "range 2" "range 3" ); for range in "${ranges[*]}"; do echo "$range"; done
range 1 range 2 range 3
$ ranges=( "range 1" "range 2" "range 3" ); for range in "${ranges[@]}"; do echo "$range"; done
range 1
range 2
range 3
```

spacewander: Got it

spacewander (approved):

spacewander:
@oikarinen 
Would you solve https://github.com/tj/git-extras/pull/1180#issuecomment-2499607635?

hyperupcall:
Thanks for adding this! This finishes up the rest of the ShellCheck improvements that I originally made in #1075 and #1056

bin/git-guilt
```diff
@@ -44,10 +44,10 @@ for file in $(git diff --name-only "$@")
 do
     test -n "$DEBUG" && echo "git blame $file"
     # $1 - since $2 - until
-    git blame $NOT_WHITESPACE --line-porcelain "$1" -- "$file" 2> /dev/null |
+    git blame "$NOT_WHITESPACE" --line-porcelain "$1" -- "$file" 2> /dev/null |
```
hyperupcall, line 47: These new changes will cause `git blame` to fail whenever `-w` (or `--ignore-whitespace`) is not supplied at the command line.

bin/git-scp
```diff
@@ -108,10 +108,10 @@ function scp_and_stage
 	then
 		local _TMP=${0///}
 		echo "$list" > "$_TMP" &&
-		_sanitize $list &&
+		_sanitize "$list" &&
 		_info "Pushing to $remote ($(git config "remote.$remote.url"))" &&
 		rsync -rlDv --files-from="$_TMP" ./ "$(git config "remote.$remote.url")/" &&
-		git add --force $list &&
+		git add --force "$list" &&
```
hyperupcall, line 114: These new changes will now cause `git add` to fail in all cases. Before, it would only fail on files that contained whitespace.

bin/git-scp
```diff
@@ -108,10 +108,10 @@ function scp_and_stage
 	then
 		local _TMP=${0///}
 		echo "$list" > "$_TMP" &&
-		_sanitize $list &&
+		_sanitize "$list" &&
```
hyperupcall, line 111: Same comment as below.

hyperupcall (changes requested):
I spotted a few issues with the new changes. Could you go back and test any of the non-trivial changes to make sure they work?

oikarinen:
> I spotted a few issues with the new changes. Could you go back and test any of the non-trivial changes to make sure they work?

Thanks, if the test suite doesn't cover the changes probably best approach is to just ignore and perhaps have an issue(s) in backlog to fix those to limit the size of the PR.

hyperupcall:
> Thanks, if the test suite doesn't cover the changes probably best approach is to just ignore and perhaps have an issue(s) in backlog to fix those to limit the size of the PR.

Sounds good to me!

bin/git-scp
```diff
@@ -107,6 +107,7 @@ function scp_and_stage
 	if [ -n "$list" ]
 	then
 		local _TMP=${0///}
+        # shellcheck disable=SC2086
```
spacewander, line 110: Bad indent and bad location?

oikarinen: Hey, tabs vs spaces thing. Sorry I rebased the branch to main instead of merge from old habits :(

spacewander (approved):

hyperupcall:
Great, thanks!

PR #1180, opened 2024-11-24
fix(github-actions): changed files output for editorcondig-checker

oikarinen:
Seems that ci.yml includes a dead dependency action just to get list of changed files for the PR / push. This seems weird as the repo is all about git. Propose removing this extra dep with straightforward implementation that supports force push for the PR branches.
Also:
- github runners do have `jq` preinstalled, so that was redundant and now no longer needed. 
- separate the install step of the go dependency

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,25 +8,32 @@ jobs:
   lint:
     runs-on: 'ubuntu-latest'
     steps:
-      - name: 'Get Changed Files'
-        id: 'files'
-        uses: 'masesgroup/retrieve-changed-files@v3'
-        with:
-          format: 'json'
-        # a force push will cause the action above to fail. Don't do force push when people are
-        # reviewing!
       - name: Check out code.
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: 'Get Changed Files'
+        id: 'files'
+        run: |
+          {
+            echo "added_modified<<EOF"
+            git diff --name-only "${BASE_SHA:-$BEFORE_SHA}...${HEAD_SHA:-GITHUB_REF}"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -8,25 +8,32 @@ jobs:
...
       - name: Check out code.
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: 'Get Changed Files'
+        id: 'files'
+        run: |
+          {
+            echo "added_modified<<EOF"
+            git diff --name-only "${BASE_SHA:-$BEFORE_SHA}...${HEAD_SHA:-GITHUB_REF}"
```
spacewander, line 20: Would you print the diff result to the output so that we can check it if the CI doesn't work well?

spacewander (commented):
Thanks for your contribution!

.github/workflows/ci.yml
oikarinen: Hey, this can be verified by running the github actions in debug mode, then github will log the contents of the env variable. But for sure it would be easier by just printing that in the step. I will amend!

spacewander (approved):

spacewander:
@oikarinen 
Merged. Thanks!

spacewander:
@oikarinen 
The CI on the main branch is broken now. Maybe you forget to add a `$` before `GITHUB_REF`? I think you can verify the behavior from the push event by testing it on your own fork.

oikarinen:
> @oikarinen The CI on the main branch is broken now. Maybe you forget to add a `$` before `GITHUB_REF`? I think you can verify the behavior from the push event by testing it on your own fork.

Yeah that's it! i did test the pull_request and then called it good and the push path was not covered 🥹

oikarinen:
@spacewander here's the missing `$` https://github.com/tj/git-extras/pull/1184

PR #1183, opened 2024-11-26
fix(ci): use poetry

oikarinen:
For managing the python dependencies, running the unit tests and spellchecker. This makes it easier for everyone to have same environment as CI and allows for use of dependabot or similar for security and version updates if so desired. Also includes few cosmetic edits for style for the ci.yml


hyperupcall:
LGTM! I do prefer poetry myself. Just to let you know, the Python changes to CI are somewhat moot as the Python tests will eventually be replaced with BATS.

hyperupcall (approved):

PR #1191, opened 2025-02-13
feat(git-bulk): Add compatibility to match workspaces using environnement variable

pierreay:
# Problem

With the current implementation of `git-bulk` (`df53711` at the time of writing), using environment variables inside the `~/.gitconfig` file to specify a `bulk` workspace works partially.

For example, let's say that `$MYWORKSPACE` is set to `~/myworkspace`.
Let's say we use the following configuration in `~/.gitconfig`:
```bash
[bulkworkspaces]
    myworkspace = $MYWORKSPACE
```
Next, what currently **is working** is the following command, using `status` as a demo command to test `git-bulk`:
```bash
git bulk -w myworkspace status
```
But what is **not working** is the following:
```bash
cd ~/myworkspace && git bulk status
```

However, it is expected to **either** work or not work in the both situations.

# Rationale

I think it would be a good feature to have it working in both situations.
In this way, it would allow to easily use the same workspace name with different paths for different hosts, for example.

# Solution

In the function that test whether our shell `$PWD` is in a workspace or not, I simply dereference variable values if there is any and I use the `realpath` of both `$PWD` and the workspace directory.
In addition to enabling full usage of environnement variable, this allows to match a workspace directory even if there is a mismatch between our `$PWD` and the workspace directory because of a symlink.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -120,9 +120,9 @@ function wsnameToCurrent () {
   while read -r workspace; do
     if [ -z "$workspace" ]; then continue; fi
     parseWsName "$workspace"
-    # compatibility with environnement variable inside .gitconfig:
+    # compatibility with environment variable inside .gitconfig:
     # use combination of `eval` and `realpath` to:
-    # 1. dereference possible environnement variable contained in $rwsdir
+    # 1. dereference possible environment variable contained in $rwsdir
     # 2. match workspaces even across different symlinks
     rwsdir_rp=$(eval realpath $rwsdir 2>/dev/null)
     pwd_rp=$(eval realpath $PWD 2>/dev/null)
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -120,7 +120,14 @@ function wsnameToCurrent () {
   while read -r workspace; do
     if [ -z "$workspace" ]; then continue; fi
     parseWsName "$workspace"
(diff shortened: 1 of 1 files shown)


pierreay:
I stumbled upon one difference of behavior between my PR and your upstream implementation.

# Demonstration

Assuming we have this in the `~/.gitconfig`:
```
[bulkworkspaces]
    workspace = ~/workspace
    workspace-inner = ~/workspace/inner
```

Assume we run the following commands:
```bash
cd ~/myworkspace && git bulk status
cd ~/myworkspace/inner && git bulk status
cd ~/myworkspace/inner/git-repo && git bulk status
```
- In the previous bulk implementation, all commands would use the `workspace` workspace.
- In my current implementation:
1. The first command would use the `workspace` workspace
2. The second command would use `workspace-inner` workspace.
3. The third command would complain that we are not inside a workspace (the workspace root, but we are in the workspace considering directories recursively).

# Proposition

My current implementation is most intuitive for me... but I'm not the only one to use this. What do you think? While I'm suppose it is not a good idea to change the default of a command which is long used, maybe it would be great to add a flag to switch between the two behavior? It is a matter of changing only one test conditional.

# Sticking to default behavior

Anyway, to stick to the default previous behavior of the command, I just have to change in `bin/git-bulk:130` the `==` to `=~`.

hyperupcall (changes requested):

PR #1192, opened 2025-02-14
feat(git-summary): add stashed count alongside uncommitted

pierreay:
# Problem

In `git` helpers, when querying the status of a repository, it is often desired or expected to see if there is one or more stash(es) inside a repository alongside uncommitted changes or files (or at least, to have the choice to do so).

# Solution

Add a new column (after uncommitted, before author for long output style, before branch name for tabular or oneline output style) counting the number of stash(es) inside the repository.

### DIFF
--- a/bin/git-summary
+++ b/bin/git-summary
@@ -192,6 +192,9 @@ uncommitted_changes_count() {
   git status --porcelain | wc -l
 }
 
+stashed_changes_count() {
+  git stash list | wc -l
+}
 
 COLUMN_CMD_DELIMTER="¬" # Hopefully, this symbol is not used in branch names... I use it as a separator for columns
 SP="$COLUMN_CMD_DELIMTER|"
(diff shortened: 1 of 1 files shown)


pierreay:
# TODO if necessary

If this PR is of interest for the maintainers, I can update the documentation/manpage to explicitly state that we check for stashes and to update the examples with the new output.

hyperupcall:
Sounds good to me. @spacewander what do you think?

bin/git-summary
```diff
@@ -192,6 +192,9 @@ uncommitted_changes_count() {
   git status --porcelain | wc -l
 }
 
+stashed_changes_count() {
+  git stash list | wc -l
```
spacewander, line 196: Now the output is:
```
commits     : 1802
 files       : 345
 uncommitted :        0
 stashed     :        0
```

There is an extra tab before the 0.
Could you remove the tab prefix of the output? Also to the one in uncommitted_changes_count.

PR #1194, opened 2025-02-15
feat(git-bulk): add new option to no follow symlinks

pierreay:
Add a new option `--no-follow-symlinks` for `git-bulk` to not traverse symlinks under the workspace directories when searching for git repositories. This PR does not change the default behavior, it only add the option for the user.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -15,13 +15,12 @@ no_follow_symlinks=false
 # print usage message
 #
 usage() {
-  echo 1>&2 "usage: git bulk [-q|--quiet] [-g] ([-a]|[-w <ws-name>]) <git command>"
+  echo 1>&2 "usage: git bulk [--no-follow-symlinks] [-q|--quiet] [-g] ([-a]|[-w <ws-name>]) <git command>"
   echo 1>&2 "       git bulk --addworkspace <ws-name> <ws-root-directory> (--from <URL or file>)"
   echo 1>&2 "       git bulk --removeworkspace <ws-name>"
   echo 1>&2 "       git bulk --addcurrent <ws-name>"
   echo 1>&2 "       git bulk --purge"
   echo 1>&2 "       git bulk --listall"
-  echo 1>&2 "       git bulk --no-follow-symlinks"
 }
 
 cdfail() {
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -155,7 +155,7 @@ function executBulkOp () {
       find_flags+=(-L)
     fi
     # find all git repositories under the workspace on which we want to operate
-    allGitFolders=( $(eval find $find_flags . -name ".git" 2>/dev/null) )
+    readarray allGitFolders < <(find "${find_flags[@]}" . -name ".git" 2>/dev/null)
 
     for line in "${allGitFolders[@]}"; do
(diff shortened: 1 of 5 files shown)


bin/git-bulk
```diff
@@ -146,7 +147,15 @@ function executBulkOp () {
...
-    allGitFolders=( $(eval find -L . -name ".git") )
+    # build `find` flags depending on command-line options
+    local find_flags=()
+    if [[ "$no_follow_symlinks" == true ]]; then
+      find_flags+=(-P)
+    else
+      find_flags+=(-L)
+    fi
+    # find all git repositories under the workspace on which we want to operate
+    allGitFolders=( $(eval find $find_flags . -name ".git") )
```
hyperupcall, line 158:
```suggestion
    readarray allGitFolders <(find "${find_flags[@]}" . -name ".git")
```

Possible to do this instead? This won't break on filepaths that include a newline and removes the (unecessary?) eval

bin/git-bulk
```diff
@@ -146,7 +147,15 @@ function executBulkOp () {
     local actual=$PWD
     [ "${quiet?}" != "true" ] && echo 1>&2 "Executing bulk operation in workspace ${inverse}$actual${reset}"
 
-    allGitFolders=( $(eval find -L . -name ".git") )
+    # build `find` flags depending on command-line options
+    local find_flags=()
+    if [[ "$no_follow_symlinks" == true ]]; then
+      find_flags+=(-P)
+    else
+      find_flags+=(-L)
```
hyperupcall, line 155: This codepath changes the default. According to the manpage:

> Since it is the default, the -P option should be considered to be in effect unless either -H or -L is specified.

I don't think changing the default behavior is a good idea. Maybe the flag should be `--follow-symlinks` to make this clearer?

hyperupcall (approved):
Thanks! I think the feature is good, had a few comments on the implementation.

bin/git-bulk
pierreay: Thank you for the improvement. It is fixed with my new commit `5a7c33d`.
A few comments:
- Compared to your proposition, I added a missing `<` (the first one is the `stdin` redirection, the second one is the process substitution). I tested this implementation against paths with spaces, which is working. Note that I assume you meant "*filepaths that include a space*" and not "*filepaths that include a newline*", right?
- I removed the unnecessary `eval`, which was introduced here since the very first commit `57c0eb7` of `git-bulk`, for a non obvious reason to me.

bin/git-bulk
pierreay: Indeed, this codepath change the default of `find`... but not the default of the `git-bulk` behavior, where the `-L` flag was added in commit `3730339` in 2018.

I did this in this way to **not change the default `git-bulk` behavior**. But I agree with you that it may have been better to not change `find` default in the first place.
So, from this, what do you prefer? On my side, I don't have any strong opinion about this. :)

spacewander:
@hyperupcall 
Would you take a look at this PR again? Thanks!

bin/git-bulk
hyperupcall: > I assume you meant "filepaths that include a space" and not "filepaths that include a newline", right?

Yeah that's what I meant haha

hyperupcall (approved):
Thanks! With the new changes everything looks great

spacewander:
Merged. Thanks!

PR #1195, opened 2025-02-15
feat(git-bulk): add new option to not follow hidden directories

pierreay:
Add a new option `--no-follow-hidden` for `git-bulk` to not traverse hidden directories (containing a dot at the beginning) under the workspace directories when searching for git repositories. This PR does not change the default behavior, it only adds the option for the user.

Note that on the implementation, I did not changed the way `find` is searching for directories. Instead, I checked using a regex before performing the operation if we are under a hidden directory. This was the easiest way I found.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -16,13 +16,12 @@ no_follow_hidden=false
 # print usage message
 #
 usage() {
-  echo 1>&2 "usage: git bulk [--no-follow-symlinks] [-q|--quiet] [-g] ([-a]|[-w <ws-name>]) <git command>"
+  echo 1>&2 "usage: git bulk [--no-follow-symlinks] [--no-follow-hidden] [-q|--quiet] [-g] ([-a]|[-w <ws-name>]) <git command>"
   echo 1>&2 "       git bulk --addworkspace <ws-name> <ws-root-directory> (--from <URL or file>)"
   echo 1>&2 "       git bulk --removeworkspace <ws-name>"
   echo 1>&2 "       git bulk --addcurrent <ws-name>"
   echo 1>&2 "       git bulk --purge"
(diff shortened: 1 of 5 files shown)


bin/git-bulk
```diff
@@ -153,7 +154,10 @@ function executBulkOp () {
       eval cd "\"$gitrepodir\"" # into git repo location
       local curdir=$PWD
       local leadingpath=${curdir#"${actual}"}
-      guardedExecution "$@"
+      # do not execute if we do not want to consider a ".git" directory under a hidden directory
+      if [[ ! ( $no_follow_hidden == true && "$leadingpath" =~ "/." ) ]]; then
```
hyperupcall, line 158:
```suggestion
      if [ $no_follow_hidden = false ] || ! [ "$leadingpath" =~ "/." ]; then
```

hyperupcall (changes requested):
Thanks! Feature is good. Had a question about implementation

bin/git-bulk
pierreay: Indeed! I just fixed the `[` to `[[` (see `8121c97`).

hyperupcall (approved):
Thanks!

spacewander (approved):

spacewander (commented):
@pierreay 
Do we need to update the completion script?

pierreay:
> @pierreay Do we need to update the completion script?

Actually, I did it on a local branch, but not merged into the branch of the PR...sorry! And I forgot the completion in the same way for the already merged [#1194](https://github.com/tj/git-extras/pull/1194)... So I cherry-picked the right commit for the two PRs (this one and the mentioned one) at the same time. ;)

spacewander (approved):

spacewander:
@pierreay 
Thanks!

PR #1196, opened 2025-02-15
fix(git-bulk): quiet find errors by default

pierreay:
It may be a bad experience for a user to see a bunch of find errors (*e.g.*, file name too long, no permissions, *etc.*) when using `git-bulk`.  I feel like it is not the job of the tool to report all errors that `find` can throws when searching all files under a hierarchy. Therefore, I quiet the `find` errors by default in this PR. If showing its errors is really on purpose, maybe we could create a flag to do so?


hyperupcall (approved):
Agreed, sounds reasonable to me!

spacewander (approved):

spacewander:
@pierreay 
Merged. Thanks!

PR #1197, opened 2025-02-15
fix(git-bulk): fix workspace selection when cd fails

pierreay:
`cd` may fails for multiple reasons:
- mistake when editing `.gitconfig` manually
- previously existing workspace that have been removed
- ...

Currently, if `cd` fails, the `BulkOp` continue its execution ... in the workspace defined in a higher directory that where the user, despite the user specified a specific workspace (`-w`).

The user should be noticed of a failed `cd` (this is really not expected for a valid configuration) and the operations should stop to not execute something unexpected.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -116,8 +116,11 @@ function parseWsName () {
   rwsdir=${wsspec#* }
   if [[ ${rwsdir:0:1} == '$' ]]; then
     # Dereference the `rwsdir` value which is a variable name.
-    rwsdir=${rwsdir:1}
-    rwsdir=${!rwsdir}
+    rwsdir_varname=${rwsdir:1}
+    rwsdir=${!rwsdir_varname}
+    if [[ -z "${rwsdir}" ]]; then
+      echo 1>&2 "error: bad environnement variable: $rwsdir_varname" && exit 1
+    fi
   fi
   rwsname=${wsspec#*.} && rwsname=${rwsname%% *}
 }
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -142,7 +142,7 @@ function executBulkOp () {
   listall | while read -r workspacespec; do
     parseWsName "$workspacespec"
     if [[ -n $wsname ]] && [[ $rwsname != "$wsname" ]]; then continue; fi
-    eval cd "\"$rwsdir\""
+    eval cd "\"$rwsdir\"" || exit 1
     local actual=$PWD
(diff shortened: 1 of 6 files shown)


hyperupcall (approved):
Thanks!- Surprised this wasn't already here.

As a side-note, those `eval`s are a bit suspicious to me

spacewander (commented):
Maybe we can also drop the `cd` in this PR?

pierreay:
> Thanks!- Surprised this wasn't already here.
> 
> As a side-note, those other `eval`s are a bit suspicious to me

I was surprised too, maybe not too many people invested time into `git-bulk`.
Anyway, in my last commit `7daedfd`, I propose a patch which get rid of the `eval`s while keeping the ability to use environment variables inside the `.gitconfig` (which is a feature I use, see my other [PR#1191](https://github.com/tj/git-extras/pull/1191) for better context).

spacewander:
> > Maybe we can also drop the `cd` in this PR?
> 
> Which one? There is 3 of them ^^'. We cannot get rid of the 2nd one without changing `guardedExecution` which is designed to work in the current directory. The 3rd one do not seems neither useful nor harmful to me. The first one could be replaced by the path inside the `find` command and concatenated to the 2nd `cd`. But I don't see the pros of requiring to do so.

Sorry for my typo! Actually, I want to drop the `eval`, which you have done in the new commit.

bin/git-bulk
```diff
@@ -111,7 +111,14 @@ function checkWSName () {
 # parse out wsname from workspacespec
 function parseWsName () {
   local wsspec="$1"
+  # Get the workspace value from its specification in the `.gitconfig`.
+  # May be an absolute path or a variable name of the form: `$VARNAME`
   rwsdir=${wsspec#* }
+  if [[ ${rwsdir:0:1} == '$' ]]; then
+    # Dereference the `rwsdir` value which is a variable name.
+    rwsdir=${rwsdir:1}
+    rwsdir=${!rwsdir}
```
spacewander, line 120: Should we exit 1 when $rwsdir is expanded to ""?

spacewander (commented):
Should we doc the variable name feature added in this PR?

pierreay:
> Should we doc the variable name feature added in this PR?

Documentation has been added in `4054cb3` and `ebb6d37` for both `Commands.md` and man page. Note that in `0e4b6c4` I also mention the `.gitconfig` storing workspaces configuration, even if it is not strictly related to this PR.

spacewander (approved):

spacewander:
@pierreay 
Merged. Thanks!

PR #1198, opened 2025-02-16
fix(git-bulk): fix a bad integer expression

pierreay:
Fix a logic error inside the `allowedargcount()` function. This function may be called with one or two arguments. However, no default values are assigned to `$1` and `$2` that are used inside a numerical comparison. Therefore, when using a bad number of arguments for the following lines:

```
listall|purge) allowedargcount 1;;
addcurrent|removeworkspace) allowedargcount 2;;
```

Then, we would get the error `[: : integer expression expected`. To fix this, we assign the 0 default value to `$1` and `$2`, such that we trigger the error message destined to the user without any integer error when there is a bad number of argument and that the function is called with only 1 argument instead of 2.

### DIFF
--- a/bin/git-bulk
+++ b/bin/git-bulk
@@ -129,7 +129,7 @@ function wsnameToCurrent () {
 
 # helper to check number of arguments.
 function allowedargcount () {
-	if [ "$paramcount" -ne "$1"  ] && [ "$paramcount" -ne "$2" ]; then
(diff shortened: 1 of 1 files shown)


hyperupcall (approved):
Nice catch! This probably should have been a loop so the function is valid for any arity. But I think it's a bit moot since I think this fix is good too.

spacewander (approved):

spacewander:
@pierreay 
Merged! Thanks.

PR #1200, opened 2025-03-24
Mostly finish pytest to Bats conversion

hyperupcall:
This finalizes the conversion from Pytest to Bats, mostly so coverage can be calculated. Big shoutout to @vanpipy for making the original tests, which made things easier.

Code coverage is mentioned in the testing README.md. I got it to work with bashcov:

```sh
bashcov -- bats ./tests
```

`coverage/index.html` looks like this:

image

This is what opening a file looks like:

image

I wasn't able to get kcov to work, unfortunately. If anyone else is able to figure that out, I would appreciate it. But bashcov seems to work fine.

I updated CI to move the Bats testing to a separate job because it speeds things up, testing in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,8 @@ jobs:
       - name: Setup Bats
         id: setup-bats
         uses: bats-core/bats-action@3.0.0
+        with:
+          bats-version: 'v1.8.1'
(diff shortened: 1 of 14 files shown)


spacewander (approved):

spacewander:
@hyperupcall 
Merged. Thanks!

PR #1201, opened 2025-03-24
Format comparisons, functions, and redirections to be consistent

hyperupcall:
It can be difficult reading through scripts here with every file seemingly having a different convention (for argument parsing, help menu, function delarations, etc.). This cleans up some of that with a little [checkstyle.py script I made for asdf-vm](https://github.com/asdf-vm/asdf/blob/master/scripts/checkstyle.py) a while back (it has autofix). It catches some things that ShellCheck doesn't, or autofixes things that ShellCheck doesn't.

One thing I would like to do, is add a `chekstyle.py` here, so we can easily enforce those style issues (and possibly other automatically fixable things) for new PRs. Is there interest in that?

### DIFF
--- a/bin/git-clear
+++ b/bin/git-clear
@@ -2,7 +2,7 @@
 PROGNAME="git-clear"
 FORCE=0
 
-function _usage() {
+_usage() {
 cat << EOF
 usage: $PROGNAME options
 usage: $PROGNAME -h|help|?
@@ -38,6 +38,6 @@ else
     clean=y
 fi
 
-if [ "$clean" == "y" ]; then
+if [ "$clean" = "y" ]; then
     git clean -d -f -x && git reset --hard
 fi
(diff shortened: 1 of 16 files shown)


spacewander (commented):
Could we put this check in the CI?

bin/git-clear
```diff
@@ -2,7 +2,7 @@
 PROGNAME="git-clear"
 FORCE=0
 
-function _usage() {
+_usage() {
```
ferdnyc, line 5: Which means the results have all of:

```bash
# Definitions without a space, because they
# were changed from
function fname { ... }
# to
fname() { }

# Definitions without a space, because they were
# changed from
function fname() { ... }
# to
fname() { ... }

# Definitions with a space, because they were
# changed from
function fname () { ... }
# to
fname () { ... }
```

(And that's _just_ in the ones the script modified.) Would it be better to decide on a particular style and apply it consistently?

ferdnyc (commented):
This seems like a helpful bit of cleanup, @hyperupcall -- I agree with @spacewander, it would be good to have those checks in the CI as well.

hyperupcall:
Script added to CI!.

spacewander:
@hyperupcall 
Merged. Thanks!

PR #1208, opened 2025-06-20
git-adopt

nfultz:
This PR adds a small script for managing/updating a dotfiles repo (see also https://dotfiles.github.io).

I believe I have completed all the steps described in `CONTRIBUTING.md`. I'm not able to test on Mac or OpenBSD but I don't believe there is anything Linux-specific in this.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -1,5 +1,6 @@
 
  - [`git abort`](#git-abort)
+ - [`git adopt`](#git-adopt)
  - [`git alias`](#git-alias)
  - [`git archive-file`](#git-archive-file)
  - [`git authors`](#git-authors)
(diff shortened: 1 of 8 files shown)


spacewander (commented):
Is this tool used widely? If it's only used for dotfile management, maybe it should be somewhere else like dotfile-extras.

spacewander (commented):
Is this tool used widely? If it's only used for dotfile management, maybe it should be somewhere else like dotfile-extras.

nfultz:
> Is this tool used widely? If it's only used for dotfile management, maybe it should be somewhere else like dotfile-extras.

I'm not able to find the dotfile-extras project you are referring to on github or google.

Issue #831, opened 2020-04-10
Use ronn-ng to generate the documentation


spacewander:
Currently we are using ronn to convert markdown docs to man pages and html files. Ronn is not longer maintained so we want to use ronn-ng instead.

There are two blocking issues about ronn-ng:
* https://github.com/apjanke/ronn-ng/issues/44
* https://github.com/apjanke/ronn-ng/issues/45

thoward27:
I'd also suggest [pandoc](https://pandoc.org/index.html) for consideration.

elonderin:
and i propose asciidoc b/c it has a defined syntax

PR #1211, opened 2025-08-25
Fix minor typo in `git rename-branch` man page

danpaolella:
Show the option as `<old-branch>` and `<new-branch>` instead of `&lt;old-branch&gt;` and `&lt;new-branch&gt;`.

### DIFF
--- a/man/git-rename-branch.1
+++ b/man/git-rename-branch.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "GIT\-RENAME\-BRANCH" "1" "July 2019" "" "Git Extras"
+.TH "GIT\-RENAME\-BRANCH" "1" "September 2025" "" "Git Extras"
 .
 .SH "NAME"
 \fBgit\-rename\-branch\fR \- rename local branch and push to remote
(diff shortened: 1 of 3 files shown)


man/git-rename-branch.1
```diff
@@ -21,11 +21,11 @@ Rename local branch and push the new branch to remote
 .
 .nf
 
-&lt;old\-branch&gt;
```
spacewander, line 24: This file is generated with https://github.com/tj/git-extras/blob/main/man/Readme.md#description.
Better to fix in the original source or the generator.

spacewander: That's fine.

danpaolella: I eventually managed to run `ronn` and generate the doc files, [here I added a few notes](https://github.com/tj/git-extras/issues/831#issuecomment-3246118350) in case it's useful to somebody else.

I'd like to ask:
1. I added a new commit as per guidelines, would you like me to remove the previous revision for a cleaner history?
2. I re-formatted the source file for consistency with the other `rename` commands docs, but the `rename-file` options are rendered as a definition list in the HTML: would you like me to adjust the `rename` docs (and perhaps other commands), to have their options turned into DLs too?

spacewander:
@danpaolella 
Merged. Thanks!

Issue #1222, opened 2026-01-08
feature request: repl runs `status` when user hits enter


jjb:
something nice the now abandoned https://github.com/thoughtbot/gitsh does is run `status -sb` whenever you hit enter, would be cool to do that with repl, or make it customizable.

spacewander:
@jjb 
Would you like to work on this?

spacewander:
I would recommend adding a new option: `git config --global git-extras.repl.on-enter-command "status -sb"`

And run it after line 42 (empty command handling):
```
# Get configured command to run on enter
on_enter_cmd=$(git config --get git-extras.repl.on-enter-command)
if test -n "$on_enter_cmd"; then
  eval git $on_enter_cmd
fi
```

PR #1223, opened 2026-01-10
ability to specify command when hitting enter

jjb:
closes https://github.com/tj/git-extras/issues/1222

image

### DIFF
--- a/man/git-repl.1
+++ b/man/git-repl.1
@@ -35,6 +35,7 @@ Ends the repl session\.
 You can specify a default command to run when hitting enter:
 .P
 \fBgit config \-\-global git\-extras\.repl\.on\-enter\-command "git status \-sb"\fR
+.P
 You can configure which character is used at the end of the prompt (default \fB>\fR):
 .P
 \fBgit config \-\-global git\-extras\.repl\.prompt\-character "±"\fR


bin/git-repl
```diff
@@ -39,7 +39,13 @@ while true; do
   # Built-in commands
   case $cmd in
     ls) cmd=ls-files;;
-    "") continue;;
+    "")
+      on_enter_cmd=$(git config --get --default '' git-extras.repl.on-enter-command)
+      if test -n "$on_enter_cmd"; then
+        eval "$on_enter_cmd"
+      fi
+      continue
```
spacewander: The `continue` will skip `exit_status=$?` below?

hyperupcall (approved):
This LGTM assuming the comments by @spacewander are addressed!

bin/git-repl
jjb: before my change, when there was an empty command, the result was `continue`

after my change, now there is the optional command executed, and then always `continue` like before

(maybe you know this and are asking something else)

i just now tested the exit_status feature with and without `git-extras.repl.on-enter-command` being set and it still works

jjb: oh i see

i'm not sure if persisting the default commands status would be useful, but i also don't know the use case for normal commands either. so the user knows the exit code just in case different non-0 codes have meaning?

image

jjb: okay, fixed that, and it's a nice improvement anyway because we are running the command through the normal command evaling logic

iTerm2 2026-01-19 21 05 13

spacewander:
@jjb 
Merged. Thanks!

PR #1226, opened 2026-01-18
Add persistent history to `git-repl`

jjb:
- **persistent command history**
- **notes**

### DIFF
--- a/man/git-repl.1
+++ b/man/git-repl.1
@@ -1,6 +1,6 @@
 .\" generated with Ronn-NG/v0.10.1
 .\" http://github.com/apjanke/ronn-ng/tree/0.10.1
-.TH "GIT\-REPL" "1" "January 2026" "" "Git Extras"
+.TH "GIT\-REPL" "1" "February 2026" "" "Git Extras"
 .SH "NAME"
 \fBgit\-repl\fR \- git read\-eval\-print\-loop
 .SH "SYNOPSIS"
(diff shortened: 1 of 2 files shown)


jjb:
@spacewander @hyperupcall take a look

jjb:
i have comments in the code discussing different approaches we could take

bin/git-repl
```diff
@@ -4,6 +4,12 @@ git version
 echo "git-extras version ""$(git-extras -v)"
 echo "Type 'ls' to ls files below current directory; '!command' to execute any command or just 'subcommand' to execute any git subcommand; 'quit', 'exit', 'q', ^D, or ^C to exit the git repl."
 
+# global history for all projects. we could make option for per-project history
+# for my use, i want this global history, and it's also much more simple to implement
+# name is analogous to bash_history and zsh_history
+HISTFILE=~/.git_repl_history
```
spacewander: Better to make it configurable

jjb:
@hyperupcall wdyt?

spacewander (approved):
LGTM

jjb:
@hyperupcall to be clear - i'm totally happy with that 😄 you know shell better so i figured if you wanna put on your finishing touches as you see best - if you don't have time or otherwise prefer not to, i'll go ahead and try to implement whats you suggested

jjb:
@hyperupcall thanks! i just tested with an history file that has only whitespace (a mash of newlines and spaces), and that breaks it after your regex removal

hyperupcall:
Oops! I just committed my suggestions and put the `shopt -s extglob` at the top. Do you think it needs anything else?

PR #1227, opened 2026-01-19
Fix cleanup of temporary files

pabs3:


Makefile
spacewander: What about building the script content in a subshell and redirecting it directly to the destination? This avoids mktemp, chmod on temp files from the beginning.

PR #1228, opened 2026-01-29
Support passing ref to `git-commits-since`

eskultety:
The commits-since command only considers dates. It's useful to dump a list of commits based on a ref, i.e. a tag, branch, or even another commit. This patch introduces a new optional '--ref|-r' flag that toggles a ref-specific variant of 'git log'.

Docs for the respective module has been updated & regenerated as well.

### DIFF
--- a/bin/git-commits-since
+++ b/bin/git-commits-since
@@ -1,5 +1,36 @@
 #!/usr/bin/env bash
 
-SINCE="last week"
-test $# -ne 0 && SINCE=$*
-git log --pretty='%h: %an - %s' --after="@{$SINCE}"
+REF=""
+SINCE=""
+
+while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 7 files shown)


bin/git-commits-since
```diff
@@ -1,6 +1,31 @@
...
-git log --pretty='%an - %s' --after="@{$SINCE}"
\ No newline at end of file
+
+echo "... commits since ${REF:-$SINCE}" >&2
+
+if [[ -n "$REF" ]]; then
+    git log --pretty='%an - %s' "$REF..HEAD"
+else
+    git log --pretty='%an - %s' --after="@{$SINCE}"
+fi
```
spacewander: Let's fix the editorconfig lint

hyperupcall (approved):
LGTM

PR #1236, opened 2026-03-04
feat: add `resolve` cmd

### LINKED ISSUE #1234: add `resolve` command
It works like this (updated to fix bug):
```sh
git diff --name-only --diff-filter=U --relative -z |
    xargs -0r $(git var GIT_EDITOR) && \
    git diff --check . && \
    git add . && \
    git continue
```
[I have an approximation of it as an alias](https://github.com/Rudxain/dotfiles/blob/9cfb62b2c9d8e5bc628ee6d319025f577e3ef73d/.config/git/config#L65-L68)

Rudxain:
closes #1234

> [!note]
> I'll `push -f` only while in `draft` state, to add missing docs

### DIFF
--- a/bin/git-resolve
+++ b/bin/git-resolve
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+set -eufo pipefail
+IFS=$' \t'
+
+#shellcheck disable=SC2046
+# ignore files outside WD, to avoid overload
+git diff --name-only --diff-filter=U --relative -z |
+    while IFS= read -r -d '' f; do
+        # `--quiet` implies `--exit-code` which is incompatible with `--check`;
+        # however, it could be special-cased internally, so it could work
+        # (untested)
+        if ! git diff --check -- "$f" &>/dev/null; then
+            printf '%s\0' "$f"
+        fi
+    done |
+    # WARN: `-r` is too new
+    # https://github.com/tj/git-extras/pull/1233#issuecomment-3957233996
+    xargs -0r $(git var GIT_EDITOR)
+
+# user might have removed other conflict-markers,
+# so check full WT
+git diff --name-only --diff-filter=U -z |
+    while IFS= read -r -d '' f; do
+        # WARN: `f` is relative to repo root!
+        # this might be broken
+        if git diff --check -- "$f"; then
+            git add -- "$f"
+        fi
+    done
+
+git continue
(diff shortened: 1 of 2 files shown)


bin/git-resolve
Rudxain, line 1: If `rerere` is enabled (even if `autoUpdate` is `false`), and a conflict matches, this will jump directly to `git continue`. This is a potential pitfall

hyperupcall (commented):
Thank you for the pull request - just a few general questions, assuming all maintainers agree on adding this feature: https://github.com/tj/git-extras/issues/1234#issuecomment-4447385479

- Is it required to ignore files outside the working directory to "avoid overload"? In my opinion this is unintuitive and I don't think it causes overload: the command should behave the same no matter where the user is in the git repository.
- I assume the `git diff --check` ensures that no files with conflict markers are committed. However, it looks like it also checks if there are whitespace errors. I worry that for some user-supplied files, a "whitespace error" (false positive) will always be activated, is there some way to avoid this?

Code comments:
- I don't think many other bash programs in this repository set `IFS` to `$' \t'`, so I would prefer if this file would be consistent with that.
- Would removing the `git continue` at the bottom avoid the potential pitfall that you mention? I think it would be better to not run `git continue` in general and let the user do it when they are ready.

Rudxain:
> Is it required to ignore files outside the working directory [...]

Not really, it's a preventive measure in case there are too many conflicted files. It ensures `EDITOR` doesn't use too much memory and reduces cognitive load for the user. I believe it's a sensible default since the user can easily change their CWD (and if CWD = worktree-root, then no files are ignored), but I agree that consistency is important as well.

> `git diff --check` [...] also checks if there are whitespace errors. I worry that for some user-supplied files, a "whitespace error" (false positive) will always be activated, is there some way to avoid this?

I have the same concern. I haven't tested, but this command should work as alternative:
```sh
git -c core.whitespace='' diff --check
```
> I don't think many other bash programs in this repository set `IFS` to `$' \t'`, so I would prefer if this file would be consistent with that.
d
I agree, but it has an effect on `xargs -0r $(git var GIT_EDITOR)` as I removed `\n` just in case, but it doesn't seem necessary.

> Would removing the `git continue` at the bottom avoid the potential pitfall that you mention?

IIRC, yes

> I think it would be better to not run `git continue` in general and let the user do it when they are ready

I agree. I can remove it now. It can be re-added in the future if a workaround for `rerere` is found

Rudxain:
I've tested on GNU nano (version 9.1), and it fails with "Standard input is not a terminal". It's because of `xargs` :(

beceqada32ps3 (approved):

PR #1239, opened 2026-03-15
feat: add git-delete-gone-branches command

codersofthedark:
Closes #1220

## What

Adds a new `git delete-gone-branches` command that deletes all local branches whose remote-tracking branch has been deleted (shown as `[gone]` in `git branch -vv`).

## Why

After PRs are merged and remote branches are cleaned up on GitHub/GitLab, local branches are left behind with no remote. There's no existing git-extras command to clean these up — `git delete-merged-branches` handles a different case (merged but remote still exists).

## Usage

```bash
# Delete all gone branches
$ git delete-gone-branches
Deleted branch feature-123 (was abc1234).
Deleted branch feature-456 (was def5678).

# Preview without deleting
$ git delete-gone-branches --dry-run
Would delete the following branches:
  feature-123
  feature-456
```

## Changes

- `bin/git-delete-gone-branches` — the command
- `man/git-delete-gone-branches.md` — man page
- `Commands.md` — added to index and description section
- `etc/git-extras-completion.zsh` — zsh completion with `--dry-run` flag

## Checklist

- [x] Script starts with `#!/usr/bin/env bash`
- [x] Man page written
- [x] Added to `Commands.md`
- [x] Updated `etc/git-extras-completion.zsh`
- [ ] `check_integrity.sh` — requires local install of ronn to generate `.1`/`.html` man files; happy to add if needed

PR #1240, opened 2026-03-15
feat: add git-delete-gone-branches

mkindahl:
When using GitHub and using the "Delete" button after merging a branch into a upstreams repository, the remote branch is gone but the local branch stays.

This command removes all "gone" branches that are still checked out.

I have tested this on macOs and Linux, but nothing else.

### DIFF
--- a/Commands.md
+++ b/Commands.md
@@ -18,6 +18,7 @@
  - [`git cp`](#git-cp)
  - [`git create-branch`](#git-create-branch)
  - [`git delete-branch`](#git-delete-branch)
+ - [`git delete-gone-branches`](#git-delete-gone-branches)
  - [`git delete-merged-branches`](#git-delete-merged-branches)
  - [`git delete-squashed-branches`](#git-delete-squashed-branches)
  - [`git delete-submodule`](#git-delete-submodule)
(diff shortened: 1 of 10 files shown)


mkindahl:
I get the following when running `check_integrity.sh` despite having added a line similar to the one for `git-delete-merged-branches`.

```
mats@fury:~/repos/git-extras$ bash check_integrity.sh 
Add description of git-delete-gone-branches in Commands.md
```

hyperupcall:
Thank you for your contribution.

One main issue I see, is that if the branch includes the string "gone", it will automatically be removed, even if it's not actually gone from the remote.

Interestingly, we received a similar PR in #1239. I like the general approach that was decided on, but I'm not sure I'll be giving my approval for the pull request, for some unrelated reasons. But I don't want to ask you to change code if I'm not completely sure that PR #1239 isn't going to be merged.

mkindahl:
> Thank you for your contribution.
> 
> One main issue I see, is that if the branch includes the string "gone", it will automatically be removed, even if it's not actually gone from the remote.
> 
> Interestingly, we received a similar PR in #1239. I like the general approach that was decided on, but I'm not sure I'll be giving my approval for the pull request, for some unrelated reasons. But I don't want to ask you to change code if I'm not completely sure that PR #1239 isn't going to be merged.

The approach in #1239 does indeed look like a more solid approach. It might make sense to wait until it is closed before closing this PR, but feel free to close this PR once you know the situation.

Issue #1246, opened 2026-04-21
`git-utimes` desired behaviour for directories


Xenira:
Hello, I have a question regarding the desired behaviour for `git-utimes` and directories.

We had an environment running with git `2.30.2-1` (bullseye) and updated to `2.39.5-0` (bookworm).
Before the update directories and files were touched. After the update only files are touched.
This was with version `6.4.0` of `git-extras` but `7.4.0` seems to affect only files as well. I therefore think this is due a change in `git` itself.

As git does not track directories this is technically correct.
We would still like to have the timestamps of the directories to be updated.

My question is what the desired default behaviour is and if you would accept a PR that adds this back in (maybe behind a flag if not default behaviour)?

spacewander:
What about adding a flag to touch the directories?

Issue #1245, opened 2026-04-17
'git whatchanged' is nominated for removal


Xenira:
When executing `git utimes` i get the following error message:

```
'git whatchanged' is nominated for removal.

hint: You can replace 'git whatchanged <opts>' with:
hint:	git log <opts> --raw --no-merges
hint: Or make an alias:
hint:	git config set --global alias.whatchanged 'log --raw --no-merges'

If you still use this command, here's what you can do:

- read https://git-scm.com/docs/BreakingChanges.html
- check if anyone has discussed this on the mailing
  list and if they came up with something that can
  help you: https://lore.kernel.org/git/?q=git%20whatchanged
- send an email to <git@vger.kernel.org> to let us
  know that you still use this command and were unable
  to determine a suitable replacement

fatal: refusing to run without --i-still-use-this
```

git version: 2.53.0
git-extras version: 7.4.0

hyperupcall:
Thanks for the report!. This was fixed in #1212 but it has not been part of a release yet.

hyperupcall:
@spacewander Do you think a release would be appropriate soon since `git-utimes` is not usable for newer versions of Git? The more we wait the more distros (not to mention Homebrew on macOS) will upgrade their Git version with this message.

spacewander:
Yes. We could kick off a release early. Would you like to create a release yourself? Feel free to reach me via email on my GitHub profile if you want to know the release details.

hyperupcall:
@spacewander Saw the release issue, I won't be able to do it this release anyways.

spacewander:
> [@spacewander](https://github.com/spacewander) Saw the release issue, I won't be able to do it this release anyways.

No worry. I will cover it.

PR #1247, opened 2026-04-22
feat(git-utimes): add ability to update directory timestamp

Xenira:
New flag `--touch-dirs` or `-d` allows to update directory timestamp when using `git utimes` command.

Refs: #1246

I haven't done much with awk till now. Only tested under linux with git `2.53.0`.

### DIFF
--- a/bin/git-utimes
+++ b/bin/git-utimes
@@ -16,20 +16,18 @@ log_opts=(
 op=eq
 
 for arg in "$@"; do
-    case "$arg" in
-        --newer)
-            op=le
-            ;;
-        --touch-dirs | -d)
-            log_opts+=(--dirstat=files,0)
-            ;;
-        *)
-            >&2 echo "unknown argument $arg found"
-            exit 1
-            ;;
-    esac
-
-    shift
+  case "$arg" in
+    --newer)
+      op=le
+      ;;
+    --touch-dirs | -d)
+      log_opts+=(--dirstat=files,0)
+      ;;
+    *)
+      >&2 echo "unknown argument $arg found"
+      exit 1
+      ;;
+  esac
 done
 
 # BSD systems
--- a/bin/git-utimes
+++ b/bin/git-utimes
@@ -7,12 +7,30 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-if [[ "${1:-}" == "--newer" ]]; then
-  op=le
-  shift
-else
-  op=eq
-fi
+# %ct: committer date, UNIX timestamp / %at: author date, UNIX timestamp
+# --diff-filter=d: exclude deleted files
+log_opts=(
+  --format='%ct'
+  --diff-filter=d
+)
+op=eq
+
+for arg in "$@"; do
+    case "$arg" in
+        --newer)
+            op=le
+            ;;
+        --touch-dirs | -d)
+            log_opts+=(--dirstat=files,0)
+            ;;
+        *)
+            >&2 echo "unknown argument $arg found"
+            exit 1
+            ;;
+    esac
+
+    shift
+done
 
 # BSD systems
 if date -j &>/dev/null; then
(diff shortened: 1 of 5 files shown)


bin/git-utimes
```diff
@@ -7,12 +7,30 @@
...
+        --touch-dirs | -d)
+            log_opts+=(--dirstat=files,0)
+            ;;
+        *)
+            >&2 echo "unknown argument $arg found"
+            exit 1
+            ;;
+    esac
+
+    shift
```
spacewander: The shift modifies $@ but has no effect on the for loop's iteration (it already has its own copy).

bin/git-utimes
```diff
@@ -7,12 +7,30 @@
...
+)
+op=eq
+
+for arg in "$@"; do
+    case "$arg" in
+        --newer)
+            op=le
+            ;;
+        --touch-dirs | -d)
+            log_opts+=(--dirstat=files,0)
```
spacewander: --dirstat=files will override --diff-filter=d, so a directory will be touched with a timestamp from a deletion-only commit. Is this expected?

bin/git-utimes
```diff
@@ -7,12 +7,30 @@
...
+# %ct: committer date, UNIX timestamp / %at: author date, UNIX timestamp
+# --diff-filter=d: exclude deleted files
+log_opts=(
+  --format='%ct'
+  --diff-filter=d
+)
+op=eq
+
+for arg in "$@"; do
+    case "$arg" in
```
spacewander: Inconsistent indentation

man/git-utimes.1
```diff
@@ -1,6 +1,6 @@
-.\" generated with Ronn-NG/v0.9.1
-.\" http://github.com/apjanke/ronn-ng/tree/0.9.1
-.TH "GIT\-UTIMES" "1" "May 2022" "" "Git Extras"
+.\" generated with Ronn-NG/v0.10.1
+.\" http://github.com/apjanke/ronn-ng/tree/0.10.1
+.TH "GIT\-UTIMES" "1" "January 1980" "" "Git Extras"
```
spacewander: I wonder why we are back to 1980

bin/git-utimes
Xenira: Thinking about it we might want to include deletes when touching directories.
That would mirror the unix behaviour where the directory timestamp is updated on file deletes as well.

man/git-utimes.1
Xenira: Didnt notice that. For some reason ronn can't get a valid date on my system. Fixed.

bin/git-utimes
```diff
@@ -91,13 +105,17 @@ FILENAME==tmpfile {
 }
 # skip blank lines
```
spacewander, line 106: We should update this comment

bin/git-utimes
```diff
@@ -91,13 +105,17 @@ FILENAME==tmpfile {
...
-  if (substr($1, length($1), 1) ~ /D/) {
-    next
-  }
   if (NF == 1) {
-    ct=$1
-    next
+    # dirstat line
+    if ($1 ~ /^.+% .+\/$/) {
+      # remove percentage before path
+      gsub(/^.+% /, "", $1)
```
spacewander, line 112: When calling in a suddir, will a path be stripped twice? One in here, another in line 120

krlmlr:
Unrelated, but could be perhaps addressed as part of this, since we're touching it here?

With recent Git 2.52.0 and git-extras 7.4.0, I see:

```
✗ git utimes                                 
'git whatchanged' is nominated for removal.

hint: You can replace 'git whatchanged <opts>' with:
hint:   git log <opts> --raw --no-merges
hint: Or make an alias:
hint:   git config set --global alias.whatchanged 'log --raw --no-merges'

If you still use this command, here's what you can do:

- read https://git-scm.com/docs/BreakingChanges.html
- check if anyone has discussed this on the mailing
  list and if they came up with something that can
  help you: https://lore.kernel.org/git/?q=git%20whatchanged
- send an email to <git@vger.kernel.org> to let us
  know that you still use this command and were unable
  to determine a suitable replacement

fatal: refusing to run without --i-still-use-this
```

Can you confirm?

Xenira:
This is already resolved on master. See #1245

spacewander:
Let me kick off a new release over the weekend.

PR #1253, opened 2026-05-17
chore: `get-regexp` is deprecated

Rudxain:
According to [the docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt---get-regexpname-regexp), `config --get-regexp` is deprecated in favor of `config get --all --show-names --regexp`, so I replaced all the occurrences I've found

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ install: check
 	@# chmod from rw-------(default) to rwxrwxr-x, so that users can exec the scripts
 	@chmod 775 $(TEMPFILE)
 	$(eval EXISTED_ALIASES := $(shell \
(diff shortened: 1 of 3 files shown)


spacewander (commented):
git config get raises the runtime requirement to Git 2.46+. The new git config subcommands were introduced with Git 2.46, while older supported/system Git versions, for example Apple Git 2.39.5, reject git config get --all --show-names --regexp ... with usage output.

Let's keep the old one. Does git have a plan to drop the get-regexp?

Rudxain:
> raises the runtime requirement to Git 2.46+

oh :(

> Does git have a plan to drop the get-regexp?

According to [v3.0 docs](https://git-scm.com/docs/BreakingChanges), not yet, so I guess this PR can be closed

PR #1256, opened 2026-05-25
perf(git-locked): replace `sed` by `cut`

Rudxain:
This is a micro-optimization. I did it on my own, but I asked Copilot if it could be improved, it suggested these (not tested):
```sh
git ls-files -v | sed -n '/^S /s///p'
git ls-files -v | awk '/^S /{print substr($0,3)}'
```

### DIFF
--- a/bin/git-locked
+++ b/bin/git-locked
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-git ls-files -v | grep ^S | sed -e 's|S ||'
+git ls-files -v | grep ^S | cut -c3-


hyperupcall:
Thanks, since this is labeled as an optimization as there any numbers to show it?

Rudxain:
> any numbers to show it?

no :sweat_smile:, but here's my reasoning for why it should be faster:
- `sed` has to interpret an expression, while `cut` only has to parse a range
- `sed` has to check that the expression matches every line (too much branching), while `cut` unconditionally slices every line. There could be a speed "tie" if the `sed` counterpart used the expression `s|^..||`

hyperupcall:
Those could make sense, but I"m not sure if any of those statements are really true without benchmarking and reading the tools' code. Unless there's shown to be a big difference, I think it's best to leave it the way it is. Code churn has a cost

Issue #1257, opened 2026-05-30
git-ignore remove/edit functionality


kdergachev:
Currently git-ignore only adds rules to ignore files (and prints them); adding an option to remove rules or just open the respective ignore file for editing seems like a nice addition. Currently I have a raw version of it that functions locally, so if this additional functionality is welcome, I can tidy it up and make a PR.

spacewander:
PR is welcome!

PR #1258, opened 2026-06-02
Added edit/remove functionality to git-ignore

kdergachev:
PR for issue #1257

These commits introduce flags -r/--remove and -e/--edit to git-ignore to allow removing patterns the same way they can be added and for opening ignore files in a text editor. Some other minor changes like '--' flag-argument separator introduced. New version is mostly backwards-compatible with old one.

Some quirks/things to consider:
- Flags for actions/locations override each other and the last is the actual one.
- Temporary file creation for remove (found that filling tmp file with double loop over lines and patterns is the best way to remove a list of patterns while not interpreting patterns as regex or any other type of special symbols, maybe there is a better way)
- Should edit action create a file if one does not exist?
- Unnecessary arguments are not checked for and just omitted

Submitting this as a draft, and if everything's OK/after changes are made will modify documentation accordingly (same for output with -h/--help flags).

### DIFF
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -126,7 +126,7 @@ _usage() {
 remove_patterns() {
   local args
   local tmpfile
-  args="${@:3}"
+  args="${*:3}"
   tmpfile="$(get_tmp_file)"
   local file="${2/#~/$HOME}"
   if [ -f "$file" ]; then
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -124,16 +124,16 @@ _usage() {
 }
 
 remove_patterns() {
-  local args
+  declare -a args
   local tmpfile
-  args="${*:3}"
+  args=( "${@:3}" )
   tmpfile="$(get_tmp_file)"
   local file="${2/#~/$HOME}"
   if [ -f "$file" ]; then
     echo "Removing patterns from $1 gitignore ($2)..."
     while read -r line; do
       local drop=
-      for pattern in $args; do
+      for pattern in "${args[@]}"; do
         test "$pattern" = "$line" && drop="yes"
       done
       if [ -z "$drop" ]; then
--- a/bin/git-ignore
+++ b/bin/git-ignore
@@ -2,6 +2,7 @@
 
 GIT_DIR=$(git rev-parse --git-dir 2>/dev/null)
 
+
 show_contents() {
   local file="${2/#~/$HOME}"
   if [ -f "$file" ]; then
@@ -27,6 +28,17 @@ cd_to_git_root() {
   if result=$(git rev-parse --show-toplevel 2>/dev/null); then
     cd "$result" || exit
   fi
+}
+
+get_tmp_file() {
+  local tmpfile=
+  if ! git rev-parse --git-dir &>/dev/null || [ -z "$GIT_DIR" ]; then
+    tmpfile="$(mktemp -q "${TMPDIR:-/tmp}"/git-extras-ignore.XXXXXX 2>/dev/null)"
+  else
+    cd_to_git_root --warn &>/dev/null
+    tmpfile="$(mktemp -q "$GIT_DIR"/git-extras-ignore.XXXXXX 2>/dev/null)"
+  fi
+  echo "${tmpfile:-$(mktemp -q)}"
 }
 
 global_ignore() {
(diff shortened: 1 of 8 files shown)


bin/git-ignore
```diff
@@ -94,29 +106,157 @@ add_patterns() {
...
+    echo "Editing $1 gitignore ($2)..." && $(git var GIT_EDITOR) "$file"
+    echo "Done."
+  else
+    echo "There is no $1 .gitignore yet." >&2
+    exit 1
+  fi
+}
+
+_usage() {
+  # TODO
```
spacewander: Let's either fulfill it or drop it

bin/git-ignore
```diff
@@ -94,29 +106,157 @@ add_patterns() {
...
-    *)
-      add_local "$@"
-      ;;
-  esac
-fi
+  # parse flags and args
+  level="none"
+  action="none"
+  declare -a args
+  while [ "$1" != "" ]; do
```
spacewander: It breaks documented empty-string patterns. while [ "$1" != "" ] stops parsing as soon as a pattern is "", so existing examples like git ignore --local "" "# Temporary files" no longer add anything after the empty argument. Use while [ $# -gt 0 ] instead.

bin/git-ignore
```diff
@@ -94,29 +106,157 @@ add_patterns() {
...
+
+remove_patterns() {
+  declare -a args
+  local tmpfile
+  args=( "${@:3}" )
+  tmpfile="$(get_tmp_file)"
+  local file="${2/#~/$HOME}"
+  if [ -f "$file" ]; then
+    echo "Removing patterns from $1 gitignore ($2)..."
+    while read -r line; do
```
spacewander: It can corrupt preserved lines during remove. read without IFS= trims leading/trailing whitespace, echo "$line" mishandles values like -n, and the loop skips a final line without a trailing newline. Use while IFS= read -r line || [ -n "$line" ]; do and printf '%s\n' "$line".

bin/git-ignore
```diff
@@ -94,29 +106,157 @@ add_patterns() {
   done
 }
 
+# TODO: create a file if does not exist??
+edit_file() {
+  local file="${2/#~/$HOME}"
+  if [ -f "$file" ]; then
+    echo "Editing $1 gitignore ($2)..." && $(git var GIT_EDITOR) "$file"
```
spacewander: $(git var GIT_EDITOR) "$file" loses shell quoting for editor values like vim -c 'set ft=gitignore', and echo "Done." still runs even if the editor exits non-zero.

bin/git-ignore
```diff
@@ -29,6 +30,17 @@ cd_to_git_root() {
   fi
 }
 
+get_tmp_file() {
+  local tmpfile=
+  if ! git rev-parse --git-dir &>/dev/null || [ -z "$GIT_DIR" ]; then
+    tmpfile="$(mktemp -q "${TMPDIR:-/tmp}"/git-extras-ignore.XXXXXX 2>/dev/null)"
+  else
+    cd_to_git_root --warn &>/dev/null
+    tmpfile="$(mktemp -q "$GIT_DIR"/git-extras-ignore.XXXXXX 2>/dev/null)"
```
spacewander: Why don't we always put the temp file in $TMPDIR?

bin/git-ignore
```diff
@@ -94,29 +106,157 @@ add_patterns() {
...
+      for pattern in "${args[@]}"; do
+        test "$pattern" = "$line" && drop="yes"
+      done
+      if [ -z "$drop" ]; then
+        echo "$line" >> "$tmpfile"
+      else
+        echo "Removed $line pattern."
+      fi
+    done < "$file"
+    mv "$tmpfile" "$file" # tmp file created in get_tmp_file 'deleted'
```
spacewander: If .gitignore or the global ignore file is a symlink, mv "$tmpfile" "$file" replaces the symlink instead of updating its target; it can also lose file metadata. Can we update it in place?

spacewander (commented):
Let's update the docs,  and completion script in https://github.com/tj/git-extras/tree/main/etc (if possible)

bin/git-ignore
kdergachev: While thinking about these issues found that perl has \E \Q escapes. These could work too (something like perl -pe "s/\Q${pattern1}\E\n?//; s/\Q${pattern2}\E\n?//; ..."). Which of the approaches do you think is better?

bin/git-ignore
kdergachev: Changed the line to use eval - it seems to fix this issue.

bin/git-ignore
kdergachev: Wanted it to work the same way git does and it writes its tmp files in .git. Changed to always use global tmp directory.

bin/git-ignore
spacewander: I wonder if we have to add perl in https://github.com/tj/git-extras/blob/main/Installation.md#dependencies. Could it be solved with awk or sed?

kdergachev: I think I found a way to do that in awk and also keep echos saying what was dropped. Added it in 3e5076b.

bin/git-ignore
```diff
@@ -94,29 +114,148 @@ add_patterns() {
...
+}
+
+remove_patterns() {
+  declare -a args
+  local tmpfile
+  args=( "${@:3}" )
+  tmpfile="$(mktemp -q "${TMPDIR:-/tmp}"/git-extras-ignore.XXXXXX 2>/dev/null)"
+  local file="${2/#~/$HOME}"
+  if [ -f "$file" ]; then
+    echo "Removing patterns from $1 gitignore: $2"
```
spacewander, line 162: Let's move tmpfile creation under here. We don't need temp file if -f is false

spacewander (commented):
LGTM except one thing. Let's update usage and docs.

kdergachev:
Done

bin/git-ignore
```diff
@@ -94,29 +114,172 @@ add_patterns() {
   done
 }
 
+edit_file() {
+  local file="${2/#~/$HOME}"
+  if [ -f "$file" ]; then
+    echo "Editing $1 gitignore: ($2)" && eval "$(git var GIT_EDITOR) $file"
```
spacewander: Will spaces or shell metacharacters in the $file break the eval since we don't quote them?

kdergachev: Yes, I guess so. The safest way to prevent that seems to be with $(printf '%q' "$file") (${file@Q} should work too, but for old versions of bash is unavailable). Should we do the same to the GIT_EDITOR part or is it assumed to be safe?

spacewander: Let's do it.

kdergachev: Fixed it. Though left GIT_EDITOR part the same since I realized that it will then treat the whole string in that variable as the command, but GIT_EDITOR (I suppose) can contain flags or be even more complex than "\<editor\> \<flags\>..." (for example try eval " 'vim -e' ").

spacewander (commented):
LGTM. But could you exclude others' commits and fix the typo?

kdergachev:
Fixed.
Had to resort to resetting with a force-push and reapplying commits (made a mistake of pulling to local, adding commits, pushing to fork and only then syncing the fork).

spacewander (approved):

spacewander:
@kdergachev 
Merged. Thank you!

PR #1261, opened 2026-07-06
feat(scp): add verbose, interactive, and dry-run flags

anderewrey:
Adds `-v/--verbose`, `-i/--interactive`, and `-n/--dry-run` to `git scp`/`git rscp`.

- `-v` prints the diff (or an rsync preview) before syncing, via `--no-pager` so it can't block on the pager like the current unconditional `git diff` does.
- `-i` also prompts for confirmation before the real push/copy.
- `-n` previews with no changes made (implies `-v`).
- Default behavior (silent, real sync) is unchanged, so existing batch-loop usage isn't affected.
- Flags must precede `<remote>`.

### DIFF
--- a/bin/git-scp
+++ b/bin/git-scp
@@ -194,7 +194,7 @@ function reverse_scp()
 	local status=0
 	if [ "$verbose" -eq 1 ]
 	then
-		rsync -rlDvni --files-from="$_TMP" "$(git config "remote.$remote.url")/" ./
+		rsync -rlDvn --files-from="$_TMP" "$(git config "remote.$remote.url")/" ./
 		status=$?
 	fi
 
--- a/bin/git-scp
+++ b/bin/git-scp
@@ -77,12 +77,13 @@ function _sanitize()
 
 function scp_and_stage
 {
-	local verbose=0 interactive=0
+	local verbose=0 interactive=0 dry_run=0
 	while :
 	do
 		case "$1" in
 			-v|--verbose)     verbose=1; shift;;
 			-i|--interactive) verbose=1; interactive=1; shift;;
+			-n|--dry-run)     verbose=1; dry_run=1; shift;;
 			*)                break;;
 		esac
 	done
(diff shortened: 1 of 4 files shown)


bin/git-scp
```diff
@@ -94,46 +105,103 @@ function scp_and_stage
...
 	then
 		local _TMP=${0///}
 		# shellcheck disable=SC2086
-		echo "$list" > "$_TMP" &&
-		_sanitize $list &&
-		_info "Pushing to $remote ($(git config "remote.$remote.url"))" &&
-		rsync -rlDv --files-from="$_TMP" ./ "$(git config "remote.$remote.url")/" &&
-		git add --force $list &&
+		echo "$list" > "$_TMP"
+		if _sanitize $list
```
spacewander: _sanitize $list can modify local files, contradicting “change nothing”. Move mutating sanitizers out of the dry-run path.

bin/git-scp
```diff
@@ -94,46 +105,103 @@ function scp_and_stage
...
+		then
+			_info "Pushing to $remote ($(git config "remote.$remote.url"))"
+			if [ "$dry_run" -eq 1 ]
+			then
+				rsync -rlDvn --files-from="$_TMP" ./ "$(git config "remote.$remote.url")/"
```
spacewander, line 140: failed rsync or git add can be masked by later cleanup/control-flow returning 0. Scripts may see success even when upload/staging failed. Preserve the operation status, clean up, then return it.

bin/git-scp
```diff
@@ -94,46 +105,103 @@ function scp_and_stage
...
+
+	if [ "$verbose" -eq 1 ]
+	then
+		rsync -rlDvni --files-from="$_TMP" "$(git config "remote.$remote.url")/" ./
+	fi
+
+	if [ "$dry_run" -eq 1 ]
+	then
+		rm "$_TMP"
+		return 0
```
spacewander: git rscp -n ignores failed preview rsync and returns 0 after cleanup. Capture and return the dry-run rsync status.

bin/git-scp
```diff
@@ -94,46 +105,116 @@ function scp_and_stage
...
 
 	local _TMP=${0///}
-	echo "$@" > "$_TMP" &&
-		rsync -rlDv --files-from="$_TMP" "$(git config "remote.$remote.url")/" ./ &&
+	echo "$@" > "$_TMP"
+
+	local status=0
+	if [ "$verbose" -eq 1 ]
+	then
+		rsync -rlDvni --files-from="$_TMP" "$(git config "remote.$remote.url")/" ./
```
spacewander: OpenBSD openrsync does not support -i, while git-extras requires OpenBSD compatibility. Let's use `-rlDvn` like elsewhere.

spacewander (approved):

spacewander:
@anderewrey 
Merged. Thank you!

PR #1264, opened 2026-07-30
port python tests to bats

spacewander:


hyperupcall:
@spacewander Hey are the Python tests supposed to be converted to Bats because I just see commits that remove them entirely

spacewander:
Yes, they are all rewritten to bats.

PR #1266, opened 2026-08-10
fix(is-git-repo): recognize bare repositories

### LINKED ISSUE #1238: git-browse: fails in bare repositories due to --show-toplevel check
Hi maintainers — this report was automatically generated by OpenCode (model: gpt-5.3) acting on my behalf.

`git-browse` currently rejects bare repositories because the repo check uses `git rev-parse --show-toplevel`, which fails when no working tree exists.

## Reproduction

1. Enter a bare git repository directory (for example: `/path/to/repo.git`)
2. Run `git browse`

## Current behavior

`git-browse` exits with `Not a git repo!`.

## Expected behavior

Bare repositories should be accepted as valid git repositories.

## Proposed fix

In `is_git_repo()`, replace `git rev-parse --show-toplevel >/dev/null 2>&1` with `git rev-parse --git-dir >/dev/null 2>&1`.

`--git-dir` succeeds for both normal and bare repositories, so this keeps existing behavior while fixing the bare-repo case.

vjymisal0:
## What

`is_git_repo()` (in `helper/is-git-repo`, prepended to most `git-*` commands at install time) used:

```sh
git rev-parse --show-toplevel > /dev/null 2>&1
```

to decide whether the current directory is inside a git repository. `--show-toplevel` fails inside a **bare** repository because a bare repo has no working tree, so any command relying on this shared helper (e.g. `git browse`) reports `Not a git repo!` even when run from inside a perfectly valid bare repository.

## Fix

Switch to `git rev-parse --git-dir`, which succeeds for both normal and bare repositories, while still failing (and printing `Not a git repo!`) outside of any repository.

## How I found / verified it

This is exactly #1238, which already has a maintainer LGTM (spacewander, hyperupcall) on this precise fix, but no PR had been opened for it yet:

> In `is_git_repo()`, replace `git rev-parse --show-toplevel >/dev/null 2>&1` with `git rev-parse --git-dir >/dev/null 2>&1`. `--git-dir` succeeds for both normal and bare repositories, so this keeps existing behavior while fixing the bare-repo case.

I reproduced it locally: `git init --bare repo.git && cd repo.git && bash helper/is-git-repo` prints `Not a git repo!` and exits non-zero before the fix, and exits 0 after.

## Testing

Added `tests/is-git-repo.bats` with three cases:
- succeeds inside a normal (non-bare) repo
- succeeds inside a bare repo (the regression case — fails without the fix)
- fails outside of a repo, with the `Not a git repo!` message preserved

Ran the full `tests/*.bats` suite (bats 1.14.0) before and after the change; no new failures introduced (a handful of pre-existing failures in `git-continue`/`git-scp` tests are unrelated environment issues — missing `rsync`/editor setup — present on `main` as well).

Closes #1238.

### DIFF
--- a/helper/is-git-repo
+++ b/helper/is-git-repo
@@ -3,7 +3,9 @@
 #
 
 is_git_repo() {
-  git rev-parse --show-toplevel > /dev/null 2>&1
+  # --git-dir succeeds for both normal and bare repositories, whereas
+  # --show-toplevel fails inside a bare repository (it has no working tree).
+  git rev-parse --git-dir > /dev/null 2>&1
   result=$?
   if test $result != 0; then
     >&2 echo 'Not a git repo!'
--- a/scripts/checkstyle.py
+++ b/scripts/checkstyle.py
@@ -3,7 +3,7 @@ import argparse
 import os
 import re
 import sys
-from collections.abc import Callable  # Compatibility.
+from collections.abc import Callable  # compat
 from pathlib import Path
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


spacewander:
Would you update the branch to get rid of python format lint? We have applied https://github.com/tj/git-extras/pull/1264 upstream.

spacewander (approved):

vjymisal0:
Fixed the failing lint check in 70e97c2 — this PR's new Ruff CI job surfaced pre-existing style issues in scripts/checkstyle.py (unsorted imports, deprecated typing.List/Dict, nested ifs, bare exit()); cleaned those up so the job passes.

spacewander (approved):

spacewander:
@vjymisal0 
Would you resolve the conflict? Thanks!

spacewander:
@vjymisal0 
Merged. Thanks!

PR #1275, opened 2026-09-16
feat(changelog): support excluding pathspecs

### LINKED ISSUE #1199: feature-request: exclude files/directories from 'git changelog'
we are using `git subtree` to vendor a build-system. the subtree contains tons of commits, that we do not want to see in our project's changelog (too many to remove them manually).

it might be nice to be able to exclude some files/directories from the changelog.

vjymisal0:
### Summary

Add `-e/--exclude <pathspec>` to `git changelog` so vendored subtrees and other files/directories can be omitted from generated changelogs. The option may be supplied multiple times and uses Git pathspec exclude syntax. Existing invocations remain unchanged.

### Verification

- `bash -n bin/git-changelog`
- `./check_integrity.sh changelog`
- Manual regression: `--exclude vendor` omits vendor-only commits, while the default invocation still includes them.
- `git diff --check`
- `make test` could not run because `bats` is not installed in the environment.

Commit includes DCO signoff.

Fixes #1199


spacewander (commented):
LGTM. Please update tests and https://github.com/tj/git-extras/blob/main/etc/git-extras-completion.zsh

vjymisal0:
Thanks @spacewander! I added Bats coverage for excluding commits by pathspec and updated the zsh completion for `-e`/`--exclude`. The change is pushed in 7ac4233 with DCO sign-off. I also verified the changelog script with bash syntax checking and a manual exclusion case; the full Bats suite could not run here because bats is not installed.

vjymisal0:
Thanks @spacewander — added tests and updated shell completion/documentation for changelog pathspec exclusions. The branch is clean and pushed.

bin/git-changelog
```diff
@@ -154,17 +157,26 @@ _fetchCommitRange() {
   local list_all="${1:-false}"
   local start_tag="$2"
   local final_tag="$3"
+  local pathspec=()
+
+  if [[ "${#GIT_LOG_EXCLUDES[@]}" -gt 0 ]]; then
+    pathspec=(-- .)
+    local exclude
```
hyperupcall: Could you use parameter expansion for this part (like `arr=( "${arr[@]/%/$suffix}" )`)?

bin/git-changelog
```diff
@@ -18,6 +18,8 @@ git_editor() {
   eval "$GIT_EDITOR" '"$@"'
 }
 
+GIT_LOG_EXCLUDES=()
```
hyperupcall: Please move this up to where the other global variables are defined

hyperupcall (changes requested):
Thanks, looks good just a few requested changes.

bin/git-changelog
vjymisal0: Done! Moved GIT_LOG_EXCLUDES=() up alongside the other global variable declarations at the top.

bin/git-changelog
vjymisal0: Good suggestion! Updated to use parameter expansion pathspec=(-- . ${GIT_LOG_EXCLUDES[@]/#/:(exclude)}) instead of the loop.

spacewander (approved):

hyperupcall (approved):

